PR #13, opened 2025-09-21
Fix hardcoded paths to use environment variables

mvoehringer:
Replace hardcoded user paths with PAI_HOME and HOME variables for better portability across different user environments.


danielmiessler:
Fixed, thank you!

PR #15, opened 2025-09-22
Feature/mcp list

mvoehringer:
### Summary
  Refactored the statusline script to directly query Claude CLI for MCP information instead
  of parsing the settings.json file, resulting in more reliable and accurate MCP listing.

  ### What Changed
  - **Fixed shell parameter expansion**: Corrected `${PAI_HOME:~}` to `${PAI_HOME:-~}` for
  proper default value handling
  - **Direct MCP querying**: Replaced JSON parsing with `claude mcp list` command for
  real-time MCP status
  - **Simplified fabric path**: Changed to use `${HOME}` directly for the fabric patterns
  directory
  - **Streamlined formatting logic**: Reduced complexity while maintaining special formatting
   for key MCPs (daemon, stripe)

  ### Why This Change
  - **Previous MR only worked reliably with specific MCPs** and could fail with
  custom or non-standard MCP configurations

  The new approach:
  - Gets real-time MCP status directly from Claude CLI
  - **Works robustly with any MCP configuration**, not just predefined ones
  - Eliminates JSON parsing dependencies
  - More maintainable and cleaner code structure
  - Ensures accuracy of displayed MCP information
  - Handles dynamic MCP names without hardcoding
  
  @danielmiessler hope this works on your system this time

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,12 +12,11 @@
 
 <br/>
 
-![Static Badge](https://img.shields.io/badge/mission-upgrade_humans_using_AI-purple)
+![Static Badge](https://img.shields.io/badge/mission-upgrade_humans_using_AI-8B5CF6)
 ![GitHub last commit](https://img.shields.io/github/last-commit/danielmiessler/PAI)
 [![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](https://opensource.org/licenses/MIT)
 [![Claude Code](https://img.shields.io/badge/Powered%20by-Claude%20Code-blue)](https://claude.ai/code)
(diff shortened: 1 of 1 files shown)


danielmiessler:
Tried this but it didn't work. Did you fix it?

mvoehringer:
On my system, it looks like this:
Screenshot 2025-09-22 at 20 38 51
@danielmiessler what exactly didn’t work on your system?

grp-ops:
@danielmiessler probably a noob comment so ignore if not helpful lol - I've experienced some strange behaviour with bash / zsh on macOS recently running awk & grep commands together which can fail this fix above unless you've got the GNU specific versions installed. After installing the specific versions through brew it worked for me!

danielmiessler:
I've made some changes to this system, so I think we're going to have to change this again, unfortunately.

PR #38, opened 2025-10-15
Moved environment reload to its own snippet

dancornell:
Before if you just cut and pasted the text in the snippet block into your ~/.zshrc or ~/.bashrc file the environment reload would cause the shell to keep reloading its environment. This way you can copy the two snippets separately.


danielmiessler:
We just made some major updates. Please review and resubmit if necessary.

PR #42, opened 2025-10-18
macOS ER - deny several dangerous disk and volume management commands

emory:
…or macOS via `diskutil(8)`

Resisted the urge to fly the flag for Zsh and nest them all too, but I'll just slum it with bash. The commands I added to deny: are specific to macOS workstations and should prevent PAI from executing diskutil(8) operations that will result in dataloss are would likely result in dataloss if you're just barreling through with them. If the speedbumps on Core Storage are too much i can get more speicfic but I don't know that anyone using cs wouldn't be better served by moving to APFS and containers for volumes?

peer reveiwed by SundaePrime@ICON


emory:
there could be people using corestorage logical volume groups as a way to lay another filesystem on top of an encrypted logical volume? but like i said, they'd probably be happier with alternatives and i don't expect corestorage to get much attention from apple beyond vintage hardware support, so i think it's alright to be greedy but will follow instruction about that sort of thing. i know its just clutter if you're on Windows or gnu/Linux systems but something tells me a lot of fabric and claudeCode/PAI users are on macOS.

PR #59, opened 2025-10-22
Update statusline-command.sh

hjbrandt:
Correctly count MCP servers from new location

### DIFF
--- a/.claude/statusline-command.sh
+++ b/.claude/statusline-command.sh
@@ -33,7 +33,7 @@ fi
 
 # Count MCPs from .mcp.json (single parse)
 mcp_names_raw=""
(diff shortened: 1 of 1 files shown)


danielmiessler:
Thank you @hjbrandt for catching the MCP path issue!

✅ You correctly identified that statusline-command.sh was reading from settings.json instead of .mcp.json

**This is now fixed in main:**
We applied your fix in commit c36143e along with an additional Fabric patterns enhancement (checks bundled PAI fabric first, then falls back to system installation).

Your contribution helped us fix the statusline for all users. Thank you! 🙏

danielmiessler:
Closing as fixed in commit c36143e. Thank you for the catch!

PR #54, opened 2025-10-22
fix: Make setup.sh work on Linux

0xsalt:
## Summary

PAI's setup.sh didn't work on Linux. This fixes it. Tested on Debian 13.1 and Ubuntu MATE 24.04.

## What Was Broken

**Critical:**
1. Hooks never fired - Missing `~/.claude/hooks` symlink
2. AI identity crisis - Called itself "Claude Code" instead of your chosen name
3. Variables stayed literal - `${PAI_DIR}` didn't expand in JSON on Linux
4. Git conflicts - Couldn't run setup twice or test modifications

**Platform stuff:**
5. Forced Homebrew even when Git+Bun already installed
6. Curl piping broke - Used `echo/read` instead of `/dev/tty`
7. Old v0.6.0 paths - Still had `/PAI/hooks/` instead of `/PAI/.claude/hooks/`

## What Changed

- Added Step 8.5: Creates critical symlinks (`~/.claude/hooks`, `~/.claude/skills`, `scratchpad/`)
- Sed replacements expand shell vars in config files at install time
- Input functions use `/dev/tty` so curl piping works
- Git stash/pop workflow handles re-runs and testing custom versions
- Platform detection makes Homebrew optional on Linux
- User's chosen AI name now used everywhere

## Testing

Works on Debian 13.1 and Ubuntu MATE 24.04. Fresh installs and re-runs both pass. Hooks fire, skills load, AI knows its name.

Added ~425 lines (728 \u2192 1153 total). Fully backward compatible with macOS.

---

Hat tip to Kai for being an excellent debugging companion through the last few days!

PR #67, opened 2025-10-25
Fix setup script interactive prompt issues and shell detection

brycemagera:
## Summary

This PR fixes two issues with the `setup.sh` script that affect the user experience during installation:

### 1. Prompt buffering bug (invisible prompts)
The `ask_input()` function was using `echo -n` which doesn't reliably flush stdout before `read` executes. This caused prompts to remain in the buffer, making them invisible while the script waits for user input. Users would see a hang and only after pressing Enter would the prompt appear.

**Fix:** Replaced `echo -n` with `printf` and redirected output to stderr (`>&2`), which is unbuffered and displays immediately.

### 2. Unreliable shell detection
The original script attempted to detect the user's shell using `$ZSH_VERSION` and `$BASH_VERSION`, but this was unreliable. For example, it would detect bash even when users were running zsh (possibly due to subshells or how Claude Code spawns processes).

**Fix:** Instead of relying on automatic detection, the script now:
- Prompts users to explicitly choose between zsh (recommended for macOS) or bash
- Uses `basename "$SHELL"` to detect the current shell and suggest an intelligent default
- Gives users control while still providing sensible defaults

### Additional improvements
- Added timeouts to all `read` commands for better non-interactive support
- Consistent error handling for read timeouts throughout the script

## Testing
Tested on macOS with both automated and interactive inputs. Prompts now appear immediately and shell selection works correctly with proper defaults.

## Impact
These are quality-of-life improvements that make the setup script more user-friendly and robust. No breaking changes.

### DIFF
--- a/.claude/setup.sh
+++ b/.claude/setup.sh
@@ -80,8 +80,13 @@ ask_yes_no() {
 
     while true; do
         echo -n -e "${CYAN}${THINKING} $question $prompt: ${NC}"
-        read -r response
-        response=${response:-$default}
+        if read -r -t 30 response 2>/dev/null; then
+            response=${response:-$default}
+        else
+            # Timeout or non-interactive, use default
+            response=$default
+            echo ""
+        fi
         case "$response" in
             [Yy]* ) return 0;;
             [Nn]* ) return 1;;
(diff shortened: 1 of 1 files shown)


danielmiessler:
Thank you @bmage8923 for the well-documented PR fixing the prompt buffering and shell detection issues!

Your analysis was spot-on:
✅  buffering causing invisible prompts
✅ Unreliable shell detection using version variables  
✅ Clean fix using  + stderr and zsh

**Why we're not merging this PR:**
We just merged PR #54 which includes these same fixes plus several critical additional fixes:
- Missing ~/.claude/hooks symlink (hooks weren't firing)
- Missing ~/.claude/skills symlink (skills weren't loading)
- PAI_DIR path bugs
- Linux support

**Your contribution was valuable!**
Your clear documentation helped us understand the issues and validate the fixes in #54. The prompt/shell detection problems you identified are now solved in main.

Thank you for taking the time to document the issues so thoroughly! 🙏

danielmiessler:
Closing as the issues are now fixed in main via PR #54. Thank you again for the contribution!

PR #77, opened 2025-11-06
Updated capture-session-summary.ts syntax

pjhartlieb:
Claude Code added a bit more than I intended. I only wanted to address the syntax changes (eg. session_id versus conversation_id) based on updates to Claude Code.  New additions seem useful so I left them.


danielmiessler:
I think this is handled in the new updates.

Issue #81, opened 2025-11-09
MacOS issue step 8 setup.sh


natierau1:
Error on step 8 of using setup.sh on a macbook pro M 
...": command i expects \
followed by text
  ⎿  UserPromptSubmit hook error
  ⎿  UserPromptSubmit hook error

⏺ The error is happening because the sed command on macOS requires a different syntax than
   on Linux. The script has several sed -i commands that need to specify a backup
  extension (even if empty) on macOS.

danielmiessler:
The `setup.sh` script is now obsolete. Installation has been simplified - see updated QUICKSTART.md.

**New Installation Method:**
```bash
git clone https://github.com/danielmiessler/Personal_AI_Infrastructure.git
cd Personal_AI_Infrastructure
cp .claude/.env.example .claude/.env
# Edit .env with your API keys
cp -r .claude ~/.claude
```

No setup.sh required! This works consistently across macOS and all platforms.

See #83 for full explanation. Closing as obsolete.

PR #84, opened 2025-11-09
Fix macOS setup.sh sed syntax errors and shell detection issues

### LINKED ISSUE #83: macOS: setup.sh fails at Step 6 with sed syntax error
## Bug Description

Step 6 of setup.sh fails on macOS with the following error:

```
sed: can't read s|PAI_DIR="/path/to/PAI"|PAI_DIR="/Users/albert/PAI"|g: No such file or directory
```

### 1. Missing filename in sed command (Critical)

**Line 531** (macOS branch):
```bash
sed -i '' "s|PAI_DIR=\"/path/to/PAI\"|PAI_DIR=\"$PAI_DIR\"|g" "$PAI_DIR/.env"
```

The original code was missing the filename completely, causing sed to interpret the pattern as a filename.

### 2. Pattern mismatch with .env.example

The sed pattern searches for:
```
PAI_DIR="/path/to/PAI"
```

But `.env.example` actually contains:
```
PAI_DIR=$HOME/.claude
```

This means even after fixing the syntax error, the replacement wouldn't work.

### 3. Incorrect shell detection

Lines 415-425 use `$BASH_VERSION` and `$ZSH_VERSION` which detect the **script's shell**, not the **user's shell**. Since setup.sh runs with `#\!/bin/bash`, it always detects bash, even for zsh users.

Should use `$SHELL` environment variable instead.

## Fix

All three issues fixed in lines 415-538:

1. ✅ Use `$SHELL` for shell detection
2. ✅ Fix sed pattern to match actual .env.example format
3. ✅ Ensure sed command includes filename

## Environment

- macOS: Darwin 25.0.0 (macOS Sequoia)
- Shell: zsh (but was misdetected as bash)
- PAI version: Latest from main branch (cloned 2025-01-09)

## Reproduction

```bash
git clone https://github.com/danielmiessler/Personal_AI_Infrastructure.git
cd Personal_AI_Infrastructure/.claude
./setup.sh
# Follow prompts until Step 6
# Observe sed error
```

## Proposed Solution

See PR #[number] with fixes for all three issues.

RaginBajin:
Fixes #83

This commit resolves three critical issues in the setup.sh script that were causing failures on macOS systems:

## 1. Fixed Shell Detection (Lines 414-426)

**Problem:** The script checked $BASH_VERSION and $ZSH_VERSION variables to detect the shell type. However, these variables indicate which shell is EXECUTING the script, not the user's preferred login shell. When users ran the script with `bash setup.sh` while using zsh as their default shell, the script would incorrectly configure .bashrc instead of .zshrc.

**Solution:** Changed to use the $SHELL environment variable, which correctly identifies the user's configured login shell regardless of how the script is executed.

## 2. Fixed sed Pattern Mismatch (Line 531)

**Problem:** The script attempted to replace `PAI_DIR="/path/to/PAI"` in the .env file, but the actual .env.example file contains `PAI_DIR=$HOME/.claude`. This pattern mismatch caused the sed command to fail silently, leaving the PAI_DIR variable unconfigured.

**Solution:** Updated the sed pattern to match the actual file content: `s|PAI_DIR=\$HOME/\.claude|PAI_DIR="$PAI_DIR"|g`

## 3. Fixed BSD vs GNU sed Incompatibility (Lines 124-137)

**Problem:** macOS ships with BSD sed, while Linux and Homebrew provide GNU sed. These have incompatible syntax for in-place editing:
- BSD sed requires: `sed -i '' 'pattern' file`
- GNU sed requires: `sed -i 'pattern' file`

The script used platform detection (`$OSTYPE`) but this doesn't account for users who have installed GNU sed via Homebrew on macOS.

**Solution:** Created a `portable_sed()` helper function that detects which sed variant is installed by testing for `--version` support (GNU sed only). This function is now used consistently throughout the script for all 11 sed invocations.

## Additional Improvements

- Applied portable_sed() to all sed commands for consistency
- Fixed backup creation logic to work with both sed variants
- Added detailed inline comments explaining the fixes
- Maintains full backward compatibility with existing installations

## Testing

These changes have been tested to ensure:
- Correct shell detection when script is run with bash but user uses zsh
- Successful .env file creation with proper PAI_DIR substitution
- Compatibility with both BSD sed (macOS) and GNU sed (Homebrew/Linux)

### DIFF
--- a/.claude/setup.sh
+++ b/.claude/setup.sh
@@ -121,6 +121,21 @@ command_exists() {
     command -v "$1" >/dev/null 2>&1
 }
 
+# Portable sed function that handles both GNU sed (Linux/Homebrew) and BSD sed (macOS)
+portable_sed() {
+    local pattern="$1"
+    local file="$2"
+
+    # Check if GNU sed is available (supports --version)
+    if sed --version >/dev/null 2>&1; then
+        # GNU sed (Linux or Homebrew on macOS)
+        sed -i "$pattern" "$file"
+    else
+        # BSD sed (macOS native) - requires empty string after -i
+        sed -i '' "$pattern" "$file"
+    fi
+}
+
 # ============================================
 # Welcome Message
 # ============================================
@@ -411,15 +426,16 @@ fi
 
 print_header "Step 5: Configuring Environment"
 
-# Detect shell
-if [ -n "$ZSH_VERSION" ]; then
+# Detect shell using $SHELL variable (user's preferred shell)
+# This correctly identifies the user's shell even when the script is run with bash
+if [[ "$SHELL" == *"zsh"* ]]; then
     SHELL_CONFIG="$HOME/.zshrc"
     SHELL_NAME="zsh"
-elif [ -n "$BASH_VERSION" ]; then
+elif [[ "$SHELL" == *"bash"* ]]; then
     SHELL_CONFIG="$HOME/.bashrc"
     SHELL_NAME="bash"
 else
-    print_warning "Couldn't detect shell type. Defaulting to .zshrc"
+    print_warning "Couldn't detect shell type from \$SHELL ($SHELL). Defaulting to .zshrc"
     SHELL_CONFIG="$HOME/.zshrc"
     SHELL_NAME="zsh"
 fi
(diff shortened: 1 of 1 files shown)


RaginBajin:
This should also fix #81 as it relates to the differences in the sed command usage.

danielmiessler:
I think this is all so deprecated after the big changes we just made.

PR #91, opened 2025-11-11
fix shell detection

ramazanpolat:


danielmiessler:
I think we need to resubmit this based on the changes that were just made.

PR #101, opened 2025-11-17
voice-server: thread ElevenLabs model selection through notify path

smolcompute:
To use the voice server with ElevenLabs on the free plan, you need to specify the model because only Flash 2.5 and Turbo 2.5 will work. 

This PR created a new env var ELEVEN_MODEL_ID=eleven_flash_v2_5

and updates the server.ts file to use it. 

voice-server: thread ElevenLabs model selection through notify path
- introduce ELEVEN_MODEL_ID (env-driven, defaulting to eleven_flash_v2_5) so the server always uses a free-tier safe model unless overridden
- extend generateSpeech/sendNotification to accept a modelId and log which model/voice pair is used
- teach POST /notify to accept a model_id parameter and pass it through, letting automations request different models per call
- remove the hardcoded eleven_monolingual_v1 usage so ElevenLabs stops rejecting free-tier requests

### DIFF
--- a/.claude/voice-server/server.ts
+++ b/.claude/voice-server/server.ts
@@ -29,8 +29,11 @@ if (!ELEVENLABS_API_KEY) {
   console.error('Add: ELEVENLABS_API_KEY=your_key_here');
 }
 
+// Default model (configured via env). Can be overridden via request body.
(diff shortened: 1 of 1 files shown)


danielmiessler:
Like this one has merge issues.

PR #138, opened 2025-11-30
feat: implement dynamic identity substitution for CORE skill

RaginBajin:
This PR introduces dynamic identity substitution for the CORE skill.

**Changes:**
1.  Modified `load-core-context.ts` to read `DA`, `DA_COLOR`, and `ENGINEER_NAME` from environment variables and substitute placeholders in `SKILL.md` at runtime.
2.  Updated `CORE/SKILL.md` to use `{{DA}}` and `{{ENGINEER_NAME}}` placeholders instead of hardcoded 'Kai' and 'Daniel'.

**Benefits:**
-   Allows users to personalize their AI name and engineer name via environment variables (e.g., in `.zshrc`) without modifying core files.
-   Maintains the 'split-repo' architecture by keeping core files generic.
-   Backward compatible (defaults to 'PAI', 'blue', 'User' if env vars are missing).

### DIFF
--- a/.claude/hooks/load-core-context.ts
+++ b/.claude/hooks/load-core-context.ts
@@ -55,9 +55,21 @@ async function main() {
     console.error('📚 Reading CORE context from skill file...');
 
     // Read the CORE SKILL.md file content
-    const coreContent = readFileSync(coreSkillPath, 'utf-8');
+    let coreContent = readFileSync(coreSkillPath, 'utf-8');
 
-    console.error(`✅ Read ${coreContent.length} characters from CORE SKILL.md`);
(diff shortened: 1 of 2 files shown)


smolcompute:
These variables should be used across the repo. 

I think the .env file should be the the centralized personalization hub. 
There should be no hardcoded KAI or Daniel in the repo. 
I would also use USER instead of ENGINEER_NAME since this is supposed to be for general use.

danielmiessler:
Thank you for this contribution! The changes have been incorporated into commit 89d5ed1 with credit. The dynamic identity substitution is now live.

PR #160, opened 2025-12-06
feat: migrate __HOME__ placeholder replacement to bootstrap workflow

zestysoft:
According to the readme, for macos, we're supposed to use bootstrap.sh I assume instead of setup.sh? -- bootstrap.sh currently doesn't contain the code to replace the __HOME__ in the settings.json file.

- Add updateSettingsJson function to configurators/index.ts
- Replaces __HOME__ placeholder with actual home directory path
- Integrates into setup.ts as Step 2 in applyConfiguration


danielmiessler:

PR #161, opened 2025-12-06
fix: correct duplicate .claude in bootstrap paths

zestysoft:
I'm assuming .claude isn't supposed to be nested in these paths?


danielmiessler:

PR #171, opened 2025-12-08
feat(voice): Add pai-voice CLI as alternative to voice-server

Smartagent-so:
## Summary

Adds `pai-voice`, a lightweight CLI tool for ElevenLabs TTS/STT that works directly with hooks without requiring an HTTP server.

**Motivation:** The existing voice-server approach requires running a separate HTTP process. For simple hook integrations (e.g., announcing task completion), a direct CLI call is simpler and more reliable.

## Features

- **Text-to-speech** with immediate playback (`--play` flag)
- **Speech-to-text** transcription with diarization support
- **Voice management** - list and search available voices
- **Subscription status** - check character usage
- **Composable output** - JSON to stdout, status to stderr
- **Cross-platform** - uses system audio player (afplay on macOS)

## Usage

```bash
# Generate and play speech
pai-voice say "Task completed" --play

# Transcribe audio
pai-voice transcribe recording.mp3

# List voices
pai-voice voices --search english

# Check usage
pai-voice subscription
```

## Files Added

| File | Purpose |
|------|---------|
| `pai-voice.ts` | Main CLI (~600 LOC) |
| `README.md` | Full documentation |
| `QUICKSTART.md` | 30-second guide |
| `package.json` | Bun dependencies |
| `tsconfig.json` | TypeScript config |

## Configuration

Reads from `~/.claude/.env`:
```
ELEVENLABS_API_KEY=your_key
ELEVENLABS_VOICE_ID=21m00Tcm4TlvDq8ikWAM  # optional
ELEVENLABS_MODEL=eleven_flash_v2_5         # optional
```

## Relationship to voice-server

This is an **alternative**, not a replacement. Users can choose based on their needs:
- `voice-server` - Full HTTP server with WebSocket support
- `pai-voice` - Simple CLI for hook integration

Closes #166

Issue #166, opened 2025-12-08
[Feature Proposal] CLI alternative to voice-server for simpler hook integration


Smartagent-so:
The current voice-server approach requires a running HTTP server (port 8888) with LaunchAgent management. For hook-based TTS calls, this adds complexity without clear benefit.

I built a CLI alternative (`pai-voice`) that calls ElevenLabs directly without a server. Happy to submit a PR if interested.

## Analysis

Key differences between voice-server and CLI approach:

| Aspect | voice-server | pai-voice CLI |
|--------|--------------|---------------|
| **Architecture** | HTTP server on port 8888 | Direct script execution |
| **State** | Requires running process | Stateless, on-demand |
| **Hook integration** | `fetch("http://localhost:8888/notify")` | `spawn("pai-voice", ["say", text, "--play"])` |
| **Failure mode** | Silent fail if server down | Immediate error in hook output |
| **Dependencies** | LaunchAgent, menubar app | Single TypeScript file |

## Why CLI works better for hooks

Hooks in Claude Code are ephemeral - they fire once and exit. The HTTP server model assumes a long-running consumer, which hooks aren't.

Current hook pattern:
```typescript
// Must check if server is running, handle connection errors
await fetch("http://localhost:8888/notify", { body: JSON.stringify({...}) });
```

CLI pattern:
```typescript
// Just works - errors are visible, no server state to manage
spawn("pai-voice", ["say", "Task complete", "--play"]);
```

## Implementation

Single file: `~/.claude/bin/pai-voice/pai-voice.ts` (~400 LOC)

Commands:
- `pai-voice say "text" --play` - TTS with playback
- `pai-voice voices` - List available voices
- `pai-voice quota` - Check character usage
- `pai-voice transcribe file.mp3` - STT

Uses same ElevenLabs API, same voice IDs, same models - just without the HTTP layer.

## Proposal

Add as alternative alongside voice-server, not replacement. Users can choose based on preference.

Would you accept a PR for this?

smolcompute:
I'd be curious to try this, would you share the file ?

Smartagent-so:
Sure! Here's the file:

**https://gist.github.com/Smartagent-so/1ec75b5229aa34d39e72d5143c00eff0**

**Quick overview:**
- Single TypeScript file (~600 LOC), runs with `bun`
- Commands: `say`, `transcribe`, `play`, `voices`, `subscription`
- API key loaded from `~/.claude/.env`
[...]

PR #173, opened 2025-12-09
feat(voice-server): add TTS provider abstraction with fallback order

zitongcharliedeng:
## Summary

- Adds pluggable TTS provider system with configurable fallback order
- Extracts ElevenLabs logic into provider class (backwards compatible)
- Adds Piper provider for free local/offline TTS
- Refactors MacOSSay provider using native command
- Adds cross-platform audio playback (afplay/aplay/powershell)

## Motivation

Related to #166 and complements #171. While pai-voice adds a CLI alternative to the HTTP server, this PR adds:

1. **Provider abstraction** - Multiple TTS backends can coexist
2. **Fallback order** - Configure which providers to try via 
3. **Cross-platform audio** - Works on macOS, Linux, and WSL
4. **Local TTS option** - Piper provides free offline TTS (no API costs)

## Configuration

```json
{
  "providers": ["piper", "elevenlabs", "macos"]
}
```

First available provider is used. If no config exists, falls back to ElevenLabs API directly (fully backwards compatible).

## Providers

| Provider | Type | Platform | Requirements |
|----------|------|----------|--------------|
| elevenlabs | Cloud | All | API key in ~/.env |
| piper | Local | Linux/WSL | piper binary + voice models |
| macos | Local | macOS | None (built-in) |

## Test plan

- [ ] Verify ElevenLabs still works when no config.json present
- [ ] Test provider fallback when first provider unavailable
- [ ] Test Piper on Linux/WSL
- [ ] Test MacOS provider on macOS
- [ ] Verify /health endpoint shows correct provider

### DIFF
--- a/.claude/statusline-command.sh
+++ b/.claude/statusline-command.sh
@@ -56,10 +56,10 @@ if [ -d "$claude_dir/commands" ]; then
     commands_count=$(ls -1 "$claude_dir/commands/"*.md 2>/dev/null | wc -l | tr -d ' ')
 fi
 
-# Count MCPs from settings.json (single parse)
+# Count MCPs from .mcp.json (single parse)
 mcp_names_raw=""
-if [ -f "$claude_dir/settings.json" ]; then
-    mcp_data=$(jq -r '.mcpServers | keys | join(" "), length' "$claude_dir/settings.json" 2>/dev/null)
+if [ -f "$claude_dir/.mcp.json" ]; then
+    mcp_data=$(jq -r '.mcpServers | keys | join(" "), length' "$claude_dir/.mcp.json" 2>/dev/null)
     mcp_names_raw=$(echo "$mcp_data" | head -1)
     mcps_count=$(echo "$mcp_data" | tail -1)
 else
@@ -94,7 +94,7 @@ cache_needs_update=false
 if [ ! -f "$CACHE_FILE" ] || [ -z "$daily_tokens" ]; then
     cache_needs_update=true
 elif [ -f "$CACHE_FILE" ]; then
-    cache_age=$(($(date +%s) - $(stat -f%m "$CACHE_FILE" 2>/dev/null || echo 0)))
+    cache_age=$(($(date +%s) - $(stat -c%Y "$CACHE_FILE" 2>/dev/null || echo 0)))
     if [ $cache_age -ge $CACHE_AGE ]; then
         cache_needs_update=true
     fi
@@ -142,7 +142,7 @@ if [ "$cache_needs_update" = true ]; then
     else
         # Someone else is updating - check if lock is stale (older than 30 seconds)
         if [ -d "$LOCK_FILE" ]; then
-            lock_age=$(($(date +%s) - $(stat -f%m "$LOCK_FILE" 2>/dev/null || echo 0)))
+            lock_age=$(($(date +%s) - $(stat -c%Y "$LOCK_FILE" 2>/dev/null || echo 0)))
             if [ $lock_age -gt 30 ]; then
                 # Stale lock - remove it and try again
                 rmdir "$LOCK_FILE" 2>/dev/null


.claude/voice-server/providers/index.ts
```diff
@@ -0,0 +1,9 @@
+export interface TTSProvider {
+  readonly name: string;
+  isAvailable(): boolean;
+  speak(text: string, voiceId?: string): Promise<void>;
```
zitongcharliedeng: This interface should only provide the actual audio buffer or file for the service to play, not actually have a speak method itself. This might be a semantic issue, but still.

.claude/voice-server/tts-providers/index.ts
```diff
@@ -0,0 +1,9 @@
+export interface TTSProvider {
```
zitongcharliedeng, line 11: How are the voice IDs provided for each agent? Shouldn't we allow each agent to have their own ELEVENLABS_VOICE_ID as well as a Piper voice ID optional?

.claude/voice-server/server.ts
```diff
@@ -24,9 +25,32 @@ if (existsSync(envPath)) {
 const PORT = parseInt(process.env.PORT || "8888");
 const ELEVENLABS_API_KEY = process.env.ELEVENLABS_API_KEY;
 
-if (!ELEVENLABS_API_KEY) {
-  console.error('⚠️  ELEVENLABS_API_KEY not found in ~/.env');
-  console.error('Add: ELEVENLABS_API_KEY=your_key_here');
+// Load TTS provider from config
+const PROVIDERS: Record<string, () => TTSProvider> = {
```
zitongcharliedeng: I think provider abstractions should be within the provider index.ts, not within server logic. Also, a better name could be a provider for config name. Is this shim even needed? Couldn't we just have the config be the class names themselves? Or is this an acceptable pattern?

.claude/voice-server/server.ts
```diff
@@ -24,9 +25,32 @@ if (existsSync(envPath)) {
...
+  'elevenlabs': () => new ElevenLabs(),
+  'piper': () => new Piper(),
+  'macos-say': () => new MacOSSay(),
+};
+
+let provider: TTSProvider | null = null;
+const configPath = join(import.meta.dir, 'config.json');
```
zitongcharliedeng: This seems to be unreadable. Code should flow like natural language, maybe make a helper function. Where is the logic such that the fallback of voices depends on the order within the config.json?

.claude/voice-server/providers/Piper.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    if (result.error) throw new Error(`Piper error: ${result.error.message}`);
+    if (result.status !== 0) throw new Error(`Piper failed: ${result.stderr?.toString()}`);
+
+    const wavBuffer = this.pcmToWav(result.stdout);
+    await playAudio(wavBuffer, 'wav');
+  }
+
+  private pcmToWav(pcm: Buffer): Buffer {
+    const header = Buffer.alloc(44);
```
zitongcharliedeng: Is stuff like this really necessary? I feel like Piper would have a lot of suitable defaults. Or we could use a tool like FFmpeg if you need to convert file types. I don't know. Look at the official Piper documents to see if you're doing this right.

.claude/voice-server/providers/index.ts
```diff
@@ -0,0 +1,59 @@
...
+  isAvailable(): boolean;
+  synthesize(text: string, voiceId?: string): Promise<AudioResult>;
+}
+
+export { ElevenLabs } from './ElevenLabs';
+export { Piper } from './Piper';
+export { MacOSSay } from './MacOSSay';
+
+// Provider registry - maps config names to provider constructors
+const providerConstructors: Record<string, () => TTSProvider> = {
```
zitongcharliedeng: providerConstructorsFromConfigNames

.claude/voice-server/providers/index.ts
```diff
@@ -0,0 +1,59 @@
...
+};
+
+/**
+ * Load the first available TTS provider based on config.json order.
+ * Falls back to ElevenLabs if no config found.
+ */
+export function loadProvider(configDir: string): TTSProvider | null {
+  const configPath = join(configDir, 'config.json');
+
+  let providerOrder = ['elevenlabs', 'piper', 'macos-say'];
```
zitongcharliedeng: Why is this hardcoded? It isn't the entire point that we're using our config.json.

.claude/voice-server/config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "providers": ["piper", "elevenlabs", "macos-say"]
```
zitongcharliedeng: Reasonable default would probably be:
1. elevenlabs
2. MacOS
3. Piper

.claude/voice-server/server.ts
```diff
@@ -24,18 +26,17 @@ if (existsSync(envPath)) {
 const PORT = parseInt(process.env.PORT || "8888");
 const ELEVENLABS_API_KEY = process.env.ELEVENLABS_API_KEY;
 
-if (!ELEVENLABS_API_KEY) {
-  console.error('⚠️  ELEVENLABS_API_KEY not found in ~/.env');
-  console.error('Add: ELEVENLABS_API_KEY=your_key_here');
+// Load TTS provider based on config.json order
+const provider = loadProvider(import.meta.dir);
+
```
zitongcharliedeng, line 33: I understand we want backwards compatibility with ElevenLabs, but I feel like the logic should not be within the server but should be within ElevenLabs itself, is in the class.  In other words, the server shouldn't care about providers.

.claude/voice-server/server.ts
```diff
@@ -24,18 +26,17 @@ if (existsSync(envPath)) {
...
+if (!provider && !ELEVENLABS_API_KEY) {
```
zitongcharliedeng: Are you sure this 11 labs backwards compatibility is even a requirement? I feel like the default fact that we have 11 labs in the default order as the first priority would imply that if there are no providers, there is no 11 labs API key. Making lines like this redundant. There are quite a few.

.claude/voice-server/audio.ts
```diff
@@ -0,0 +1,44 @@
...
+function detectOS(): OperatingSystem {
+  const p = platform();
+  if (p === 'darwin') return 'macos';
+  if (p === 'linux') {
+    return release().toLowerCase().includes('microsoft') ? 'wsl' : 'linux';
+  }
+  throw new Error(`Unsupported platform: ${p}`);
+}
+
+export const OS: OperatingSystem = detectOS();
```
zitongcharliedeng: currentOperatingSystem or whatever is the idiomatic naming convention and naming case.

.claude/voice-server/tts-providers/ElevenLabs.ts
```diff
@@ -0,0 +1,50 @@
...
+      if (errorText.includes('model') || response.status === 422) {
+        throw new Error(`ElevenLabs API error: Invalid model "${this.model}". Update ELEVENLABS_MODEL in ~/.env`);
+      }
+      throw new Error(`ElevenLabs API error: ${response.status} - ${errorText}`);
+    }
+
+    const audioBuffer = await response.arrayBuffer();
+    return { audio: Buffer.from(audioBuffer), format: 'mp3' };
+  }
+}
```
zitongcharliedeng, line 48: Is there a nice Git way to show that most of these lines are similar to what was original in the Elevenlabs implementation in the server file, originally, to reduce diff?

.claude/voice-server/desktop-audio-playback.ts
```diff
@@ -0,0 +1,44 @@
+import { spawn } from 'child_process';
+import { platform, release } from 'os';
+
```
zitongcharliedeng, line 4: This file should be called something like "Desktop Audio Provider" or something that actually shows what it's for. "Audio.ts" is fine if it's a convention, but even stuff like "providers" the folder could be worded better to be like "TTS Providers".  (nit)

zitongcharliedeng:
## Manual Testing Results

### Windows WSL2 Testing (via PowerShell MediaPlayer)

| Test | Provider | Result |
|------|----------|--------|
| Audio playback from WSL | ElevenLabs | ✅ Passed |
| Audio playback from WSL | Piper | ✅ Passed |
| Path conversion (wslpath) | N/A | ✅ Working |
| MediaPlayer (no GUI window) | N/A | ✅ Working |

### Test Plan Progress

- [x] Verify ElevenLabs still works when no config.json present
- [x] Test provider fallback when first provider unavailable  
- [x] Test Piper on Linux/WSL
- [ ] Test MacOS provider on macOS (no macOS available)
- [x] Verify /health endpoint shows correct provider

### Notes

- PowerShell MediaPlayer plays audio from WSL without opening a GUI window
- \ correctly converts WSL paths to Windows format
- Provider fallback works correctly (Piper -> ElevenLabs when Piper unavailable)
- \ type correctly detects \ via kernel release string

🤖 Manually tested by Charlie on Windows 11 :)

zitongcharliedeng:
Re: per-agent voice IDs - This is already supported. The /notify endpoint accepts voice_id in the POST body, which gets passed to the provider. Each agent hook can pass a different voice_id per request. The Piper provider also supports voice mapping via voices.json config.

From Claude.

zitongcharliedeng:
Addressed remaining review comments:

**Naming fixes:**
- Renamed `SHELL_ENV` to `currentShellEnvironment` (idiomatic camelCase)
- Renamed `audio-playback.ts` to `desktop-audio-playback.ts` for clarity
- Renamed `providers/` to `tts-providers/` (done previously)

**Already addressed:**
- `providerConstructorsFromConfigNames` - was already using this name
- Provider loading via `loadProvider()` in tts-providers/index.ts
- Config order: elevenlabs, macos-say, piper
- Interface uses `synthesize()` returning `AudioResult` (buffer + format)

**Design decisions:**
- Provider map is "hardcoded" because it maps config strings to TypeScript classes - this is the factory pattern, config.json controls the order/selection
- Server.ts no longer has ElevenLabs-specific logic, just calls `provider.synthesize()`
- Piper implementation follows official CLI usage (`--model`, `--speaker`, `--output_file`, stdin for text)

All commits squashed into single conventional commit: `feat(voice-server): add pluggable TTS provider system`

.claude/voice-server/tts-providers/ElevenLabs.ts
```diff
@@ -0,0 +1,51 @@
+import type { TTSProvider } from '.';
+import { playAudio } from '../audio';
+
+export class ElevenLabs implements TTSProvider {
+  readonly name = 'elevenlabs';
+
+  private get apiKey() { return process.env.ELEVENLABS_API_KEY; }
+  private get voiceId() { return process.env.ELEVENLABS_VOICE_ID || 's3TPKV1kjDlVtZbl4Ksh'; }
```
zitongcharliedeng, line 5: Should this method even exist? I would expect to provide TTS provider classes with an ID during invocation i.e. in an Engineer Subagent context.

.claude/voice-server/voices.json
```diff
@@ -3,60 +3,108 @@
...
+      }
     },
     "architect": {
-      "voice_name": "Serena (Premium)",
+      "description": "UK Female - Strategic, sophisticated",
       "rate_multiplier": 1.35,
       "rate_wpm": 236,
-      "description": "UK Female - Strategic, sophisticated",
-      "type": "Premium"
+      "elevenlabs": {
```
zitongcharliedeng: can we nest properties like this to be under tts-voice-provider? or is it redundant? what is the idiomatic way.

.claude/voice-server/config.json
```diff
@@ -0,0 +1,3 @@
+{
```
zitongcharliedeng, line 1: Can this be a sibling property in Voices.json? Could we advocate to rename voices.json to voices-config.json? If it doesn't belong in the same JSON, then we can rename this file to be something more readable like TTS provider priority order.json.

.claude/voice-server/providers/index.ts
zitongcharliedeng: Addressed - TTSProvider interface now returns \ instead of having a speak method. The server handles playback via \.

.claude/voice-server/tts-providers/index.ts
zitongcharliedeng: Addressed - voices.json now has provider-specific voice configs as siblings:
```json
"kai": {
  "elevenlabs": { "voice_name": "Jamie (Premium)" },
  "piper": { "model": "en_GB-alan-medium", "speaker": 0 }
}
```
Server uses `getVoiceForAgent()` to look up the correct voice based on active provider.

.claude/voice-server/server.ts
zitongcharliedeng: Addressed - \ is now in \. The map is named \ as suggested. This shim is needed because we cant dynamically import classes by string name in a type-safe way - the map provides the connection between config.json names and actual class constructors.

.claude/voice-server/server.ts
zitongcharliedeng: Addressed - Added getVoiceForAgent() helper function. Provider fallback order is determined by config.json providers array - loadProvider iterates through and returns the first available one.

.claude/voice-server/providers/Piper.ts
zitongcharliedeng: Verified - Piper outputs WAV natively with --output_file flag. Removed the raw PCM conversion code that was there before. The current implementation uses Piper defaults and outputs to WAV which the audio playback handles directly.

.claude/voice-server/providers/index.ts
zitongcharliedeng: Done - renamed to providerConstructorsFromConfigNames

.claude/voice-server/providers/index.ts
zitongcharliedeng: The map connects config.json string names to actual TypeScript class constructors. We cannot dynamically instantiate classes by string name in TS without this mapping. The config.json controls the order, this map just provides the factory functions.

.claude/voice-server/config.json
zitongcharliedeng: Will update config.json to use order: elevenlabs, macos-say, piper

.claude/voice-server/server.ts
zitongcharliedeng: Addressed - Server now just calls loadProvider() from tts-providers/index.ts. The backwards compatibility logic has been removed. Server does not care which provider is used.

.claude/voice-server/server.ts
zitongcharliedeng: Removed. The backwards compatibility code is gone. If no providers are available, the server logs a warning and continues (voice just wont play).

.claude/voice-server/audio.ts
zitongcharliedeng: Addressed - Renamed to currentShellEnvironment

.claude/voice-server/tts-providers/ElevenLabs.ts
zitongcharliedeng: Git history shows the original ElevenLabs code in server.ts. The refactor moved it to its own file. Git blame on the new file will show this commit, but the logic is largely unchanged from the original.

.claude/voice-server/desktop-audio-playback.ts
zitongcharliedeng: Addressed - Renamed audio.ts to desktop-audio-playback.ts. Renamed providers folder to tts-providers.

.claude/voice-server/tts-providers/ElevenLabs.ts
zitongcharliedeng: Addressed - Removed the name property from provider classes. loadProvider() now returns { provider, name } where name comes from the config.json key used to instantiate it. This removes redundancy (DRY).

.claude/voice-server/voices.json
zitongcharliedeng: Current structure has elevenlabs/piper as sibling properties under each voice entry. Adding another nesting level like tts-voice-provider would be redundant since the whole file is about voice config. Current structure is clean and flat.

.claude/voice-server/config.json
zitongcharliedeng: Keeping them separate is cleaner: config.json handles provider priority (which TTS engine to use), voices.json handles voice definitions (which voice each agent uses per provider). Merging would conflate two different concerns. Could rename config.json to tts-provider-priority.json if that is clearer.

danielmiessler:

PR #199, opened 2025-12-15
fix: statusline now displays commands, skills, and MCPs correctly

chadif:
The statusline was calculating counts for commands, skills, and MCPs but only displaying MCP names. This fix:
- Adds skills counting (SKILL.md files in skill directories)
- Fixes MCP reading to use .mcp.json instead of settings.json
- Displays Commands, Skills, and MCPs counts on line 2
- Shows MCP server names on line 3 with tree-style formatting
- Fixes integer comparison error by ensuring mcps_count defaults to 0

Statusline now shows all capabilities: 4 commands, 12 skills, 10 MCPs

### DIFF
--- a/.claude/statusline-command.sh
+++ b/.claude/statusline-command.sh
@@ -10,9 +10,10 @@
 #   - Comment out any printf lines you don't want displayed
 #
 # LINES DISPLAYED:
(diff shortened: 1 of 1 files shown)


danielmiessler:

PR #208, opened 2025-12-15
feat: Add Emojis to output format

sti0:
Add Emojis to output format to match agent output format.


danielmiessler:

PR #251, opened 2025-12-28
Fix: Store event timestamps in UTC instead of PST

djacobsmeyer:
## Summary

Changes event timestamp storage from PST-hardcoded to UTC (ISO 8601 format) for timezone-independent event tracking.

## Problem

The current implementation stores timestamps in PST format:
- Causes confusion when PAI runs in different timezones (e.g., CST, EST, GMT)
- Events grouped by PST date can split incorrectly at midnight in other timezones
- Not portable for international users
- Hardcoded timezone assumption breaks portability principle

**Example of confusion:**
- User in CST sees system time: 16:07 CST
- Event file shows: `2025-12-28 14:07:20 PST`
- User thinks events are 2 hours old (they're not - just timezone difference)

## Solution

Store timestamps in UTC using ISO 8601 format:
- **Before:** `"timestamp_pst": "2025-12-28 14:07:20 PST"`
- **After:** `"timestamp_utc": "2025-12-28T22:07:20.123Z"`

Event files now grouped by UTC date instead of PST date.

## Changes

**`Hooks/capture-all-events.ts`:**
- Changed interface field: `timestamp_pst` → `timestamp_utc`
- Simplified timestamp function to return `new Date().toISOString()`
- Updated file path calculation to use UTC date components
- Reduced code from ~13 lines to 1 line for timestamp generation

## Benefits

✅ **Timezone independent** - Works correctly anywhere in the world  
✅ **Standard format** - ISO 8601 is the international standard  
✅ **No confusion** - Single source of truth for time  
✅ **Portable** - No hardcoded timezone assumptions  
✅ **Cleaner code** - Simpler timestamp generation  

## Backward Compatibility

**No breaking changes:**
- Frontend already uses `timestamp` (Unix milliseconds) for display
- Frontend converts to user's local timezone via `toLocaleTimeString()`
- The `timestamp_pst` field was never used by the observability dashboard
- Old events with `timestamp_pst` will work fine (field is ignored)

## Testing

Tested manually:
```bash
# Test hook execution
echo '{"session_id":"test","tool_name":"Read"}' | \
  ~/.claude/Hooks/capture-all-events.ts --event-type PostToolUse

# Verify output format
tail -1 ~/.claude/History/Raw-Outputs/.../all-events.jsonl | jq .timestamp_utc
# Output: "2025-12-28T22:16:29.032Z" ✅
```

Verified:
- ✅ Hook executes without errors
- ✅ Timestamps stored in valid ISO 8601 format
- ✅ File paths use UTC date components
- ✅ Frontend displays times correctly in local timezone

## Impact

- **Users:** Timestamps now display accurately regardless of system timezone
- **Developers:** Simpler, more maintainable code
- **International users:** No PST bias, works globally
- **Observability dashboard:** No changes needed (already timezone-aware)

### DIFF
--- a/.claude/Hooks/capture-all-events.ts
+++ b/.claude/Hooks/capture-all-events.ts
@@ -4,6 +4,9 @@
  * Captures ALL Claude Code hook events (not just tools) to JSONL
  * This replaces the Python send_event.py hook
  * Enhanced with agent instance metadata extraction
+ *
+ * Timestamps stored in UTC (ISO 8601) for timezone-independent event tracking
+ * Events grouped by UTC date to avoid midnight boundary issues
  */
 
 import { readFileSync, appendFileSync, mkdirSync, existsSync, writeFileSync } from 'fs';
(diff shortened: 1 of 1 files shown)


danielmiessler:
## PAI 2.0 Migration Notice

Thank you for this contribution! We've recently released **PAI v2.0**, which takes a fundamentally different approach from v1.x.

**What changed:** PAI v1.x tried to mirror the entire Kai system as an installable template. That approach created too many interconnected dependencies—change one piece, and others broke. v2.0 shifts to **modular packs**: self-contained, independently installable capabilities that work without understanding the whole system.

**This PR:** Because the architecture is now fundamentally different, we're closing all open v1.x PRs and issues. This isn't us ignoring your contribution—we genuinely appreciate it! It's that the system has changed so significantly that we need to revisit these within the new context.

**Next steps:**
1. Read the updated [README](https://github.com/danielmiessler/PAI#the-journey-pai-v1x--v20) (especially "The Journey: PAI v1.x → v2.0")
2. Check out the new [PAI Packs system](https://github.com/danielmiessler/PAI#pai-v20-pai-packs)
3. If the issue you were addressing still exists in 2.0, please open a new issue or PR with context for the new architecture

We're absolutely happy to address these concerns—just within the 2.0 framework. Thanks for your patience and for contributing to PAI! 🙏

PR #257, opened 2025-12-29
fix: Define PAI_DIR constant to resolve ReferenceError in help text

khali:
**Problem:**
The Art skill's help command crashed with 'ReferenceError: PAI_DIR is not defined' when using template literals like ${PAI_DIR} without defining the variable.

**Root Cause:**
The help text used ${PAI_DIR} in template literals (lines 196-197) but the variable was never defined in the TypeScript code.

**Solution:**
Added const PAI_DIR definition that reads from process.env.PAI_DIR with a fallback to resolve(process.env.HOME!, '.claude').

**Testing:**
- Verified --help now displays correctly with proper paths
- Confirmed all image generation functionality still works
- Tested with nano-banana-pro model successfully

**Impact:**
- Fixes issue where users couldn't access built-in help documentation
- No breaking changes to existing functionality
- Works with both environment variable and fallback path

### DIFF
--- a/.claude/Skills/Art/tools/generate-ulart-image.ts
+++ b/.claude/Skills/Art/tools/generate-ulart-image.ts
@@ -19,6 +19,16 @@ import { writeFile, readFile } from "node:fs/promises";
 import { extname, resolve } from "node:path";
 
 // ============================================================================
+// Constants
+// ============================================================================
+
+/**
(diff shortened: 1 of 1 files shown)


danielmiessler:
We have migrated to version 2. Please check the new system.

PR #307, opened 2026-01-03
chore: Removed legacy single file packs

sti0:
Removed legacy single file packs which led to confusing during the install process. Changed path references to folder structure packs.

### DIFF
--- a/Bundles/Kai/install.ts
+++ b/Bundles/Kai/install.ts
@@ -740,10 +740,10 @@ Files created:
 Next steps:
 
   1. Install the packs IN ORDER by giving each pack file to your AI:
-     - kai-hook-system.md
-     - kai-history-system.md
-     - kai-core-install.md
-     - kai-voice-system.md (optional, requires ElevenLabs)
+     - kai-hook-system/INSTALL.md
+     - kai-history-system/INSTALL.md
+     - kai-core-install/INSTALL.md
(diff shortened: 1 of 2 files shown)


sti0:
With the recent chages of browser skill I think the .md files should remain. But are outdated. E.g. hook and history try do create files instead of copying them.

danielmiessler:
Great cleanup work, @sti0! Removing the legacy single-file packs will reduce confusion significantly.

This has some merge conflicts with main now (likely from recent merges). Could you rebase against main when you get a chance?

```bash
git fetch origin
git rebase origin/main
git push --force-with-lease
```

Once conflicts are resolved, we'll merge. Thanks for all your contributions! 🎯

sti0:
Hi @danielmiessler,
as these files are updated yesterday. Are you sure that we should delete them? In my understanding the pack root md's are the description of the skill which has additional information within the subdirectory.
But some of these skills like hooks are trying to install some packs instead of using the SRC files. 

Could you please clarify?

sti0:
rebased in case the deletion is the right way.

danielmiessler:
Thank you @sti0 for tackling the legacy pack cleanup! 🙏

**PAI v2.1** completed this work - all single-file `kai-*.md` packs are gone, replaced with directory-based `pai-*` packs. Your PR was ahead of its time and helped drive the direction!

PR #320, opened 2026-01-04
fix(installer): respect PAI_DIR choice during installation

banjoey:
## Summary

- Fixed bug where installer ignored user's choice to use existing `$PAI_DIR`
- Installer now correctly installs to the chosen directory instead of always defaulting to `~/.claude`

## Problem

When running `bun run install.ts`, if `$PAI_DIR` is set, the installer asks:

```
📍 Existing PAI_DIR detected: /Users/example/PAI

Use existing PAI_DIR (/Users/example/PAI) for installation? [Y/n]:
```

Answering "yes" was acknowledged in the output but **completely ignored** - installation always went to `~/.claude`.

## Root Cause

In `install.ts`, the `gatherConfig()` function displayed the PAI_DIR prompt but never captured or returned the user's choice. The `main()` function then used a hardcoded path:

```typescript
const claudeDir = `${process.env.HOME}/.claude`;  // Line 589 - hardcoded!
```

## Fix

1. Added `installDir: string` to `WizardConfig` interface
2. Captured user's PAI_DIR choice in `gatherConfig()` and included it in the return value
3. Changed `main()` to use `config.installDir` instead of hardcoded path
4. Updated summary messages to display actual install location

## Testing

Tested by:
1. Setting `PAI_DIR` in shell profile
2. Running fresh install with `bun run install.ts`
3. Confirming "yes" to use PAI_DIR
4. Verified files were created in `$PAI_DIR` instead of `~/.claude`

### DIFF
--- a/Bundles/Kai/install.ts
+++ b/Bundles/Kai/install.ts
@@ -36,6 +36,7 @@ interface WizardConfig {
   userName: string;
   elevenLabsApiKey?: string;
   elevenLabsVoiceId?: string;
+  installDir: string;
 }
 
 // =============================================================================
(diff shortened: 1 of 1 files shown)


danielmiessler:
Thank you @banjoey for catching this PAI_DIR installer bug! 🙏

**PAI v2.1** includes a completely revamped wizard-style installer with proper environment detection. Your bug report helped drive these improvements - the installer now properly respects user choices.

The paths have changed (`Bundles/Kai/` → `Bundles/Official/`), but the fix is incorporated in spirit!

PR #351, opened 2026-01-08
feat(pai-voice-system): Add Linux support and CLI configuration

sti0:
## Summary

- Re-Adds full Linux platform support for the voice server (audio playback + notifications)
- Adds CLI arguments for runtime configuration (`--port`, `--extra-args`)
- Enhances .env loading from `$PAI_DIR/.env` with quote stripping (closes #346 )
- Adds devcontainer/PulseAudio documentation for containerized development

### Linux Audio Support
- Auto-detects `mpg123` or `mpv` for audio playback on Linux
- Falls back gracefully if no player is installed
- macOS continues to use built-in `afplay`

### Linux Notifications
- Uses `notify-send` for desktop notifications on Linux
- macOS continues to use `osascript`

### CLI Arguments
- `--port` / `-p`: Override server port
- `--extra-args`: Pass extra arguments to audio player (e.g., `-o pulse` for PulseAudio)

### Environment Variables
- `PAI_VOICE_SERVER_PORT`: Server port (default: 8888)
- `PAI_VOICE_SERVER_EXTRA_ARGS`: Extra audio player arguments
- Enhanced .env parsing with quote stripping for values

### Documentation
- Added devcontainer setup guide for PulseAudio
- Updated platform status to show Linux as fully supported
- Added changelog entry for v1.1.0

## Test Plan

- [x] Voice notification plays on Linux with mpg123
- [x] CLI `--port` argument works
- [x] CLI `--extra-args` argument passes to audio player
- [x] .env values with quotes are parsed correctly
- [x] TypeScript compiles without errors

PR #356, opened 2026-01-09
fix(pai-voice-system): Add Linux platform support for audio playback

pkumaschow:
- Detect platform (macOS vs Linux) using os.platform()
- Update playAudio() to use ffplay on Linux instead of afplay
- Update sendNotification() to conditionally use osascript on macOS only
- Add notify-send support on Linux (optional, graceful degradation)
- Maintain full macOS compatibility

Tested on Linux (Fedora) with ffplay successfully playing ElevenLabs TTS output.

### DIFF
--- a/Packs/pai-voice-system/src/voice/server.ts
+++ b/Packs/pai-voice-system/src/voice/server.ts
@@ -286,9 +286,47 @@ function getVolumeSetting(): number {
   return 1.0; // Default to full volume
 }
 
+// Helper function to setup playback event handlers
+function setupPlaybackHandlers(
+  proc: any,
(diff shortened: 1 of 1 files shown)


sti0:
Already addressed in #351

kaimagnus:
Thank you for the Linux audio playback support, @pkumaschow! Cross-platform audio is important for the voice system.

Unfortunately, v3.0 restructured the codebase and the `Packs/pai-voice-system/` directory no longer exists. Closing as the target paths no longer exist. If you'd like to rebase against v3.0's structure, we'd welcome it!

PR #370, opened 2026-01-10
bug: Broken URL Hash reference to #-this-readme-is-for-your-ai-assistant

renoirb:
Broken URL reference that is used as hint and URL to start installations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,8 +75,9 @@ This is open-source. This is free. This is for everyone.
 
 ---
 
-> [!TIP]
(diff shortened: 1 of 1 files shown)


kaimagnus:
Thank you for the fix, @renoirb! Good catch on the broken anchor hash.

The v3.0 release restructured the README significantly, and the section this PR targets ("This README Works Great with AI Assistants") has been replaced with new content. The anchor no longer exists in the current README, so this change is no longer needed.

We appreciate the contribution — really valued the attention to detail here!

PR #379, opened 2026-01-11
fix(pai-art-skill): Apply model-specific default sizes in Generate.ts

### LINKED ISSUE #380: bug(pai-art-skill): Generate.ts size validation fails for gpt-image-1 model
## Description

When using `Generate.ts` with `--model gpt-image-1` without explicitly specifying `--size`, the CLI fails with an invalid size error.

## Steps to Reproduce

```bash
bun run Generate.ts --model gpt-image-1 --prompt "test" --output ~/test.png
```

## Expected Behavior

Should use a sensible default size for the model (e.g., `1024x1024` for OpenAI).

## Actual Behavior

```
Error: Invalid size for gpt-image-1: 2K
```

or

```
Error: Invalid size: 1024x1024
```

## Root Cause

The validation logic in `parseArgs()` was structured incorrectly:

```typescript
if (parsed.model === "gpt-image-1" && !OPENAI_SIZES.includes(parsed.size)) {
  throw new CLIError(...);
} else if (parsed.model === "nano-banana-pro") {
  // validate gemini sizes
} else if (!REPLICATE_SIZES.includes(parsed.size)) {
  throw new CLIError(...);  // <-- This catches gpt-image-1 with valid OpenAI sizes!
}
```

When the model is `gpt-image-1` and the size IS valid, the first condition is false, so it falls through to the `else if` chain and eventually hits the Replicate size check.

## Fix

PR #379 restructures the logic to properly branch per model type and apply sensible defaults.

## Environment

- PAI version: main branch
- Pack: pai-art-skill v1.1.0

b1ackmartian:
## Summary

- Fixed broken size validation logic in Generate.ts that caused errors when using models with different size formats
- Applied sensible model-specific defaults when `--size` is omitted or invalid

Fixes #380

## Problem

The validation logic was structured as:
```typescript
if (gpt-image-1 && NOT valid) throw
else if (nano-banana-pro) { validate }
else if (NOT valid for replicate) throw
```

When using `gpt-image-1` without `--size`, the default "2K" would pass the first condition (since the size IS technically invalid for OpenAI), but then if you specify a valid OpenAI size like `1024x1024`, the first condition becomes false and it falls through to the replicate check which fails.

## Solution

Restructured to properly branch per model type and apply defaults:
- `gpt-image-1`: defaults to `1024x1024`
- `nano-banana-pro`: defaults to `2K`
- `flux`/`nano-banana`: defaults to `16:9`

## Test Plan

- [x] Tested `--model gpt-image-1` without `--size` - now defaults to 1024x1024
- [x] Tested `--model nano-banana-pro` without `--size` - works with 2K default
- [x] Generated test image successfully with Google API

---

### DIFF
--- a/Packs/pai-art-skill/src/skills/Art/Tools/Generate.ts
+++ b/Packs/pai-art-skill/src/skills/Art/Tools/Generate.ts
@@ -297,16 +297,21 @@ function parseArgs(argv: string[]): CLIArgs {
     throw new CLIError(`Too many reference images: ${parsed.referenceImages.length}. Maximum is 14 total`);
   }
 
-  // Validate size for model
-  if (parsed.model === "gpt-image-1" && !OPENAI_SIZES.includes(parsed.size as OpenAISize)) {
-    throw new CLIError(`Invalid size for gpt-image-1: ${parsed.size}`);
+  // Validate size for model (and apply model-specific defaults)
+  if (parsed.model === "gpt-image-1") {
+    if (!parsed.size || !OPENAI_SIZES.includes(parsed.size as OpenAISize)) {
+      parsed.size = "1024x1024"; // Default for OpenAI
+    }
   } else if (parsed.model === "nano-banana-pro") {
-    if (!GEMINI_SIZES.includes(parsed.size as GeminiSize)) {
-      throw new CLIError(`Invalid size for nano-banana-pro: ${parsed.size}`);
+    if (!parsed.size || !GEMINI_SIZES.includes(parsed.size as GeminiSize)) {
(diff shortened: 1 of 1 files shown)


kaimagnus:
Thank you for this fix, @b1ackmartian! You nailed the issue — each model needs its own default size format.

We've applied this fix in commit 0fc1931, which adds model-specific default sizing (flux/nano-banana get aspect ratios, gpt-image-1 gets pixel dimensions, nano-banana-pro gets resolution tiers). Closing in favor of that commit. Great analysis!

PR #368, opened 2026-01-10
fix(pai-hook-system): restore accidentally deleted observability.ts

b1ackmartian:
## Summary
- Restores `lib/observability.ts` to `pai-hook-system` pack
- File was accidentally deleted in commit 0faf5e4 during kai→pai rename

## Problem
`initialize-session.ts` imports from `./lib/observability` but the file doesn't exist:
```
error: Cannot find module './lib/observability'
```

This breaks the SessionStart hooks for anyone installing `pai-hook-system`.

## Fix
Restore the original file to `Packs/pai-hook-system/src/lib/observability.ts`

The file content is recovered from commit [0faf5e4](https://github.com/danielmiessler/Personal_AI_Infrastructure/commit/0faf5e4108db7ad65758e951bb66f76bed8410d3) where it was removed.

Closes #353

PR #410, opened 2026-01-15
fix: Restore missing observability.ts lib for pai-hook-system

ajbmachon:
## Summary
- Restores `lib/observability.ts` that was accidentally deleted during the kai-* to pai-* rebrand
- The file was deleted in commit 0faf5e4 but not re-added under the new naming convention
- This file is required by `security-validator.ts` and other hooks

## Problem
Without this file, hooks fail with:
```
error: Cannot find module './lib/observability'
```

## Fix
Restored the file from git history (commit 1c34d06).

## Test plan
- [x] Verified hooks load without module resolution errors
- [x] Confirmed observability functions are available

### DIFF
--- a/Packs/pai-hook-system/src/lib/observability.ts
+++ b/Packs/pai-hook-system/src/lib/observability.ts
@@ -1,0 +1,58 @@
+// $PAI_DIR/hooks/lib/observability.ts
+// Sends hook events to observability dashboards
+
+export interface ObservabilityEvent {
+  source_app: string;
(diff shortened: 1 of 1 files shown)


peterlynch:
is this a dupe of https://github.com/danielmiessler/Personal_AI_Infrastructure/pull/368 ?

ajbmachon:
Yes you are right. Sorry I should have checked this more carefully!

PR #416, opened 2026-01-16
Add WSL and Linux support to VoiceServer

selfscrum:
## Summary
- Add platform detection (macos, wsl, linux)
- WSL: Play audio via Windows PowerShell MediaPlayer (bypasses WSLg crackling)
- Linux: Play audio via mpv
- macOS: Keep existing afplay behavior
- Use PAI_DIR environment variable for all paths (with ~/.claude fallback)
- Load .env from both ~/.env and $PAI_DIR/.env
- Add platform info to /health endpoint

## Test plan
- [x] Tested on WSL2 - voice plays correctly through Windows MediaPlayer
- [ ] Test on macOS (existing behavior should be unchanged)
- [ ] Test on native Linux with mpv installed

### DIFF
--- a/Packs/pai-voice-system/src/VoiceServer/server.ts
+++ b/Packs/pai-voice-system/src/VoiceServer/server.ts
@@ -4,30 +4,55 @@
  */
 
 import { serve } from "bun";
-import { spawn } from "child_process";
-import { homedir } from "os";
+import { spawn, execSync } from "child_process";
+import { homedir, release } from "os";
 import { join } from "path";
 import { existsSync, readFileSync } from "fs";
 
-// Load .env from user home directory
-const envPath = join(homedir(), '.env');
-if (existsSync(envPath)) {
-  const envContent = await Bun.file(envPath).text();
-  envContent.split('\n').forEach(line => {
-    const eqIndex = line.indexOf('=');
-    if (eqIndex === -1) return;
-    const key = line.slice(0, eqIndex).trim();
-    let value = line.slice(eqIndex + 1).trim();
-    // Strip surrounding quotes (single or double)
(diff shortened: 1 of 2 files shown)


fayerman-source:
Great work on the Linux/WSL audio support! I tested a similar approach locally and can confirm mpv works well.

**Observation:** While testing on Ubuntu 24.04/WSL2, I noticed two additional macOS-specific dependencies that still cause errors on Linux:

### 1. Notification Display (`osascript`)
```
ENOENT: no such file or directory, posix_spawn '/usr/bin/osascript'
spawnargs: [ "-e", "display notification \"...\" with title \"PAI Notification\"" ]
```

**Suggested fix:** Use `notify-send` on Linux:
```typescript
if (isMac) {
  spawn('/usr/bin/osascript', ['-e', `display notification "${message}" with title "${title}"`]);
} else {
  spawn('notify-send', [title, message]);
}
```

### 2. Fallback TTS (`say`)
The fallback `speakWithSay()` function uses `/usr/bin/say` which doesn't exist on Linux.

**Suggested fix:** Use `espeak` or `spd-say` on Linux:
```typescript
if (isMac) {
  spawn('/usr/bin/say', [text]);
} else {
  spawn('espeak', [text]); // or 'spd-say'
}
```

These are non-blocking errors (the server continues to work), but addressing them would complete the cross-platform support. Happy to submit a follow-up PR if helpful!

kaimagnus:
Thank you for the WSL/Linux VoiceServer support, @selfscrum! Cross-platform voice is important.

The target paths span `Packs/` and `Releases/v2.3/` which have been superseded by v3.0. Closing as the paths no longer match. If you'd like to adapt this for v3.0's VoiceServer structure, we'd welcome it!

Issue #464, opened 2026-01-21
TUI becomes un-responsive and display area garbelled.


ozten:
Thanks for an amazing project.
macOs 26.2 (25C56)
claude: 2.1.14 (Claude Code)
PAI: Fresh v2.3 as of one hour ago
iTerm2

After using PAI for a few minutes, the TUI becomes unresponsive. Letters that I type are not displayed. Also the input starts appearing over the bottom area (see `> es tr t NG SIGNAL` in this screenshot)

<img width="710" height="749" alt="Image" src="https://github.com/user-attachments/assets/58e15bbc-70ae-40fa-8eb9-f50d42d8cde4" />

CPU and GPU look normal. This machine is a beefy Mac Studio Apple M2 Max 64GB.

## Steps to Reproduce

Probably not really... but

1. Ask PAI to start the observability dashboard
2. Notice TUI becomes unresponsive

ozten:
I reproduced this issue. 3 sessions and everytime after a couple of minutes of usage, the responsiveness of the keyboard stops. There is red text saying "Fluxuming" or whatever which I think is claude code. When responsiveness comes back, the text input is down in the heads up display area. I quit with `/exit` and run `pai` again to get a responsive TUI.

ozten:
I have a fix for this. I will test for a day before I submit a PR.

ozten:
I don't think [PR#47](https://github.com/danielmiessler/Personal_AI_Infrastructure/pull/473) fixes the garbled screen issue, but it does eliminate the unresponsiveness that reproduced 100% of the time when using PAI without Kitty.

kaimagnus:
Thanks for reporting! This appears to be an environment-specific TUI rendering issue. **v3.0** (now released) has updated hooks and output formatting. Please upgrade to [v3.0](https://github.com/danielmiessler/Personal_AI_Infrastructure/tree/main/Releases/v3.0) and let us know if the issue persists.

PR #473, opened 2026-01-23
fix(hooks): prevent terminal I/O blocking from hung Kitty subprocesses

ozten:
Fix for [issue #464](https://github.com/danielmiessler/Personal_AI_Infrastructure/issues/464).

The StopOrchestrator hook was causing keyboard unresponsiveness for 10-220 seconds after Claude Code sessions ended. This was most severe for users not running Kitty terminal.

Root cause: The tab-state handler executed `kitten @ set-tab-color` commands that hung indefinitely when no Kitty socket was available. While Promise.race() timeouts were in place, they only abandoned the promise - the subprocess continued running in the background, blocking terminal I/O until it eventually timed out on its own.

Changes:

- Add subprocess.ts utility with explicit process control using Bun.spawn() instead of Bun.$``. This enables killing hung processes with SIGTERM followed by SIGKILL fallback after 2 seconds.

- Refactor tab-state.ts with terminal detection (matching the existing pattern in UpdateTabTitle.hook.ts). Kitty terminals use remote control API with subprocess timeout protection. Other terminals fall back to escape codes for basic title support.

- Apply same terminal detection pattern to SetQuestionTab.hook.ts for consistency across all tab-related hooks.

- Add AbortController timeout protection to voice.ts and SystemIntegrity.ts handlers for defense in depth.

The fix provides three layers of protection:
1. Terminal detection - skip Kitty-specific code entirely for non-Kitty
2. Escape code fallback - basic functionality for all terminals
3. Subprocess control - kill hung processes after 2s timeout (Kitty)

### DIFF
--- a/Packs/pai-hook-system/src/hooks/SetQuestionTab.hook.ts
+++ b/Packs/pai-hook-system/src/hooks/SetQuestionTab.hook.ts
@@ -10,36 +10,27 @@
  *
  * TRIGGER: PreToolUse (matcher: AskUserQuestion)
  *
- * INPUT:
- * - None (triggered by tool match, no stdin processing)
+ * TERMINAL DETECTION:
+ * Uses the same pattern as UpdateTabTitle.hook.ts:
+ * - Kitty: Uses remote control API for colors + title
+ * - Other terminals: Falls back to escape codes for basic title support
  *
- * OUTPUT:
- * - stdout: None
- * - stderr: Status message
- * - exit(0): Always (non-blocking)
- *
- * SIDE EFFECTS:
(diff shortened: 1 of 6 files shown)


ozten:
I don't think this PR fixes the garbled screen issue, but it does eliminate the unresponsiveness that reproduced 100% of the time when using PAI without Kitty.

ozten:
Please restart the checks. This one failed because `Error: Failed to setup GitHub token: Error: Could not fetch an OIDC token. Did you remember to add `id-token: write` to your workflow permissions?`

danielmiessler:

PR #492, opened 2026-01-26
Fix TypeScript type errors across multiple skills

Clingdawg:
## Summary
- Prompting Tools: Add null checks for regex match results, fix HandlebarsTemplateDelegate type
- ReportTemplate: Fix import path casing (@/components -> @/Components)
- DashboardTemplate: Fix import path casing, add DOM libs to tsconfig, fix React.Node type
- Browser skill: Add allowImportingTsExtensions for .ts imports

## Test plan
- [ ] Run `bunx tsc --noEmit` in each affected project


Packs/pai-prompting-skill/src/skills/Prompting/Templates/Tools/ValidateTemplate.ts
```diff
@@ -99,19 +99,20 @@ function checkUnbalancedBlocks(source: string): string[] {
   const lines = source.split('\n');
 
   for (let i = 0; i < lines.length; i++) {
-    const line = lines[i];
+    const line = lines[i]!;
```
neilsoult, line 102: I don't love hacks like this in typescript. I think it's far better to use safe access operators and use default values rather than tell typescript to basically ignore typesafety on this const.

```suggestion
    const line = lines[i] ?? '';
```

kaimagnus:
Thank you for fixing the TypeScript type errors, @Clingdawg! Good catches on the Handlebars types, import paths, and React types.

Unfortunately all the files modified target `Packs/` paths which no longer exist in v3.0's restructured codebase. Closing as the target paths no longer match. We appreciate the thorough type-checking work!

PR #493, opened 2026-01-26
fix(hooks): use socket-based Kitty remote control to prevent escape sequence leaks

Mutdogus:
## Summary
Fixes P@kitty-cmd escape sequence artifacts appearing in terminal when using PAI with Kitty terminal.

## Problem
When PAI hooks use `kitty @` commands without a socket, Kitty responds via escape sequences. Since hooks run as subprocesses, these sequences cannot be properly consumed and leak into the terminal as visible garbage:

```
P@kitty-cmd{"ok": false, "error": "Remote control is disabled"}
```

This affects anyone using PAI with Kitty terminal.

## Root Cause
Kitty's `allow_remote_control yes` mode uses escape sequences for IPC. When hooks run as detached subprocesses:
1. They send commands via escape sequences to the PTY
2. Kitty writes responses back to the PTY
3. The subprocess can't read from the PTY
4. Response leaks as visible text

## Solution
- Added `getKittySocket()` helper to all tab-related hooks
- All kitty/kitten commands now use `--to <socket>` flag for socket-based communication
- Gracefully skips tab updates if no socket available (prevents leaks on misconfigured systems)

## Files Changed
- `Packs/pai-hook-system/src/hooks/UpdateTabTitle.hook.ts`
- `Packs/pai-hook-system/src/hooks/SetQuestionTab.hook.ts`
- `Packs/pai-hook-system/src/hooks/QuestionAnswered.hook.ts`
- `Packs/pai-hook-system/src/hooks/handlers/tab-state.ts`

## User Configuration Required
Users need to configure Kitty for socket-based remote control by adding to `~/.config/kitty/kitty.conf`:

```conf
allow_remote_control socket-only
listen_on unix:/tmp/kitty-$USER
```

Then restart Kitty for changes to take effect.

## Testing
Tested on macOS with Kitty terminal - artifacts no longer appear after fix.

---
🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/Packs/pai-hook-system/src/hooks/QuestionAnswered.hook.ts
+++ b/Packs/pai-hook-system/src/hooks/QuestionAnswered.hook.ts
@@ -36,18 +36,44 @@
  * - Kitty unavailable: Silent failure
  */
 
+import { existsSync } from 'fs';
+
 const TAB_WORKING_BG = '#804000';      // Dark orange - actively working
 const ACTIVE_TAB_BG = '#002B80';       // Dark blue - active tab always
 const ACTIVE_TEXT = '#FFFFFF';
 const INACTIVE_TEXT = '#A0A0A0';
 
+/**
+ * Get kitty socket path - required for socket-only remote control.
+ * Using socket-based control prevents escape sequence leaks (P@kitty-cmd artifacts).
+ */
+function getKittySocket(): string | null {
+  if (process.env.KITTY_LISTEN_ON) {
+    return process.env.KITTY_LISTEN_ON;
+  }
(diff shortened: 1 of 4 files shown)


kaimagnus:
Thank you for this smart fix, @Mutdogus! Socket-based Kitty remote control is the right approach to prevent escape sequence leaks. The shared `getKittySocket()` helper with KITTY_LISTEN_ON fallback is well-designed.

Unfortunately the target paths (`Packs/pai-hook-system/`) no longer exist in v3.0. Closing as the paths don't match. If you'd like to rebase against v3.0's hook structure, this would be a valuable fix!

PR #610, opened 2026-02-08
Add TASK Line Rule to Algorithm v0.2.25

anikin-xyz:
## Summary

Adds explicit parsing guidance for the TASK line to address a consistent failure mode where the model confuses MODE (what the user wants to do) with OUTPUT (what will eventually exist).

**The problem:** When a user says "Let's collaborate on an article about X", the TASK line was being generated as "Write article about X" — skipping the "collaborate" instruction entirely. This was reproducible across multiple sessions.

**The fix:**
- Added "TASK Line Rule" section with parsing guidance
- Key principle: parse the first imperative clause literally
- Added examples table showing correct parsing
- Added to Common Failures: "TASK captures output instead of mode"
- Updated changelog with v0.2.25.1

## Test Plan

- [x] Tested across multiple chat sessions with the same prompt
- [x] Verified the explicit parsing rule prevents the failure

---

### DIFF
--- a/Releases/v2.5/.claude/skills/PAI/Components/Algorithm/v0.2.25.md
+++ b/Releases/v2.5/.claude/skills/PAI/Components/Algorithm/v0.2.25.md
@@ -116,6 +116,28 @@ Each phase transition triggers a voice announcement via the voice server. Execut
 
 ---
 
+## TASK Line Rule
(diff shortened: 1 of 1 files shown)


danielmiessler:

PR #619, opened 2026-02-10
fix: Add audio playback queue to prevent overlapping voice notifications

HyggeHacker:
## Summary
- Adds a promise-based audio queue to the voice server that serializes playback
- Concurrent notifications (e.g., multiple hooks firing simultaneously) no longer produce overlapping audio
- Both ElevenLabs `playAudio` and macOS `speakWithSay` fallback are queued

## Problem
When multiple hooks fire close together (e.g., `UserPromptSubmit` + `FormatEnforcer`), the voice server receives concurrent `/notify` requests. Since `playAudio` is fully async, both audio streams play simultaneously, producing garbled overlapping speech.

## Solution
Simple promise chain queue:
```typescript
let audioQueue: Promise<void> = Promise.resolve();

function enqueueAudio(fn: () => Promise<void>): Promise<void> {
  audioQueue = audioQueue.then(fn, fn);
  return audioQueue;
}
```
All audio playback functions are wrapped with `enqueueAudio()`, ensuring sequential playback. New requests are still accepted immediately by the HTTP server — only the audio output is serialized.

## Files Changed
- `Packs/pai-voice-system/src/VoiceServer/server.ts` (canonical pack source)
- `Releases/v2.3/.claude/VoiceServer/server.ts`
- `Releases/v2.4/.claude/VoiceServer/server.ts`

## Test plan
- [x] Send two simultaneous `/notify` requests — second waits for first to finish
- [x] Single notifications work normally
- [x] macOS `say` fallback is also queued

### DIFF
--- a/Packs/pai-voice-system/src/VoiceServer/server.ts
+++ b/Packs/pai-voice-system/src/VoiceServer/server.ts
@@ -275,40 +275,50 @@ function getVolumeSetting(requestVolume?: number): number {
   return 1.0; // Default to full volume
 }
 
-// Play audio using afplay (macOS)
+// Audio playback queue - prevents overlapping speech from concurrent notifications
+let audioQueue: Promise<void> = Promise.resolve();
+
+function enqueueAudio(fn: () => Promise<void>): Promise<void> {
+  audioQueue = audioQueue.then(fn, fn);
+  return audioQueue;
+}
+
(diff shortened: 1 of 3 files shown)


HyggeHacker:
Closing to rework — this branch inadvertently included unrelated files. Will re-submit with only the voice server fix.

PR #635, opened 2026-02-10
fix: Prevent Inference.ts from creating noise session files

m4nt0de4:
## Summary

- Every hook calling `Inference.ts` (FormatReminder, ImplicitSentiment, AutoWorkCreation, etc.) creates a `.jsonl` session file via `claude --print`
- With 3-4 hooks per interaction turn, this generates hundreds of ghost sessions that pollute `claude --resume`
- Adds `--no-session-persistence` to the `claude` args array to prevent ephemeral inference calls from persisting session files

## Reproduction

1. Launch PAI with hooks enabled (default config)
2. Count `.jsonl` files in `~/.claude/projects/` before and after a few interactions
3. Observe 3-4 new noise files per turn (one per hook using inference)
4. Run `claude --resume` and see ghost sessions mixed with real ones

## Fix

Single-line addition of `'--no-session-persistence'` to the args array in `Inference.ts` (line 78).

## Verification

After applying the fix:
- 0 new noise `.jsonl` files per interaction turn
- `claude --resume` shows only real sessions
- All hooks still function correctly (inference results unchanged)

## Test plan

- [ ] Verify FormatReminder hook still classifies depth correctly
- [ ] Verify ImplicitSentiment hook still captures sentiment
- [ ] Confirm no new `.jsonl` files appear in projects directory after interactions
- [ ] Confirm `claude --resume` list is clean

### DIFF
--- a/Releases/v2.5/.claude/skills/PAI/Tools/Inference.ts
+++ b/Releases/v2.5/.claude/skills/PAI/Tools/Inference.ts
@@ -75,6 +75,7 @@ export async function inference(options: InferenceOptions): Promise<InferenceRes
 
     const args = [
       '--print',
+      '--no-session-persistence',
       '--model', config.model,
       '--tools', '',  // Disable tools for faster response
       '--output-format', 'text',


kaimagnus:
Thank you for the fix, @m4nt0de4! Preventing inference calls from creating noise session files is important for clean operation.

The target path (Releases/v2.5/) has been superseded by v3.0. Closing as the paths no longer match. Appreciate the contribution!

PR #637, opened 2026-02-10
fix: Allow pai --resume to accept optional session ID

m4nt0de4:
## Summary

- Claude Code's `--resume` flag supports an optional session ID: `claude --resume <UUID>`
- `pai.ts` typed `resume` as `boolean`, so it only passed the bare `--resume` flag
- Users couldn't resume a specific session through `pai` and had to call `claude` directly
- Changes the type to `boolean | string`, captures the optional next argument, and passes it through

## Reproduction

1. Run `pai --resume abc123` (or any session ID)
2. Observe that the session ID is silently discarded
3. Claude launches the session picker instead of resuming the specific session

## Fix

- Changed `resume` variable type from `boolean` to `boolean | string`
- Updated argument parsing to capture optional session ID after `--resume`/`-r`
- Pass session ID through to `claude --resume <id>` when provided

## Usage after fix

```bash
pai --resume              # Resume last session (existing behavior)
pai --resume abc123       # Resume specific session by ID (new)
pai -r abc123             # Short form (new)
```

## Test plan

- [ ] `pai --resume` still opens session picker (backward compatible)
- [ ] `pai --resume <valid-id>` resumes the specific session
- [ ] `pai -r <valid-id>` works the same way
- [ ] `pai --resume --local` doesn't capture `--local` as a session ID (starts with `-`)

### DIFF
--- a/Releases/v2.5/.claude/skills/PAI/Tools/pai.ts
+++ b/Releases/v2.5/.claude/skills/PAI/Tools/pai.ts
@@ -390,7 +390,7 @@ function cmdWallpaper(args: string[]) {
 // Commands
 // ============================================================================
 
-async function cmdLaunch(options: { mcp?: string; resume?: boolean; skipPerms?: boolean; local?: boolean }) {
+async function cmdLaunch(options: { mcp?: string; resume?: boolean | string; skipPerms?: boolean; local?: boolean }) {
   displayBanner();
   const args = ["claude"];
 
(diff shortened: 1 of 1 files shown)


m4nt0de4:
Withdrawing — after audit, the complexity-to-value ratio doesn't justify merging. The command-collision bug is theoretical (mutually exclusive intents), and the COMMANDS set introduces maintenance debt. If session ID passthrough is wanted later, a cleaner `--resume=ID` syntax would avoid the lookahead parser entirely.

PR #639, opened 2026-02-11
fix(art): detect actual image format and save with correct extension

### LINKED ISSUE #633: bug(pai-art-skill): Generate.ts saves JPEG data with .png extension, causing MIME type mismatch
## Problem

The Replicate API (Flux 1.1 Pro, Nano Banana) sometimes returns JPEG data even when `output_format: "png"` is requested. `Generate.ts` saves this data with a `.png` extension without checking the actual content format.

When OpenCode's Read tool later reads this file, it determines the MIME type from the file extension (`image/png`) and sends it to the Anthropic API. Anthropic validates the actual bytes against the declared MIME type and rejects it:

```
messages.4.content.1.image.source.base64.data: Image does not match the provided media type image/png
```

This crashes the session — the user cannot continue working.

## Root Cause

Verified with `file` command and hex dump on 6 generated "PNG" files — **all contained JPEG data** (magic bytes `FF D8 FF`):

```bash
$ file ~/Downloads/generated-image.png
JPEG image data, JFIF standard 1.01, resolution (DPI), density 300x300

$ xxd ~/Downloads/generated-image.png | head -1
00000000: ffd8 ffe0 0010 4a46 4946 0001 0101 012c  ......JFIF.....,
```

The issue is in `Packs/pai-art-skill/src/skills/Art/Tools/Generate.ts` — all generation functions use `await writeFile(output, result)` without verifying the actual format matches the file extension.

## Affected Functions

- `generateWithFlux()` (line ~472)
- `generateWithNanoBanana()` (line ~494)
- `generateWithGPTImage()` (line ~521)
- `generateWithNanoBananaPro()` (line ~616)

Additionally, `generateWithNanoBananaPro()` determines the MIME type of **reference images** purely from the file extension (line ~555-571), which can also cause issues when reference images have mismatched extensions.

## Suggested Fix

Add magic byte detection to determine the actual image format, and save with the correct extension:

```typescript
function detectActualFormat(data: Buffer | Uint8Array): "png" | "jpeg" | "webp" | "gif" | null {
  if (data.length < 12) return null;
  if (data[0] === 0x89 && data[1] === 0x50 && data[2] === 0x4e && data[3] === 0x47) return "png";
  if (data[0] === 0xff && data[1] === 0xd8 && data[2] === 0xff) return "jpeg";
  if (data[0] === 0x52 && data[1] === 0x49 && data[2] === 0x46 && data[3] === 0x46 &&
      data[8] === 0x57 && data[9] === 0x45 && data[10] === 0x42 && data[11] === 0x50) return "webp";
  if (data[0] === 0x47 && data[1] === 0x49 && data[2] === 0x46) return "gif";
  return null;
}
```

Then use it before saving:

```typescript
const data = result instanceof Buffer ? result : Buffer.from(result as any);
const actualFormat = detectActualFormat(data);
// If format doesn't match extension, save with correct extension
const finalPath = await saveImageWithCorrectFormat(data, output);
```

## Environment

- OpenCode v1.1.54
- macOS (Apple Silicon)
- Replicate API (Flux 1.1 Pro)
- Anthropic Claude API

## Related

- OpenCode also has a complementary issue: the Read tool uses `Bun.file().type` (extension-based) instead of magic bytes — see [anomalyco/opencode#9810](https://github.com/anomalyco/opencode/pull/9810)

VintLin:
## Problem

`generate` may save image bytes with a mismatched extension (e.g., JPEG bytes written to a `.png` path). This can later cause MIME-type validation failures in downstream readers/tools that trust the extension.

Additionally, reference-image MIME type selection in `nano-banana-pro` was extension-based only, which can also break if file suffix and bytes diverge.

## Changes

- Added magic-byte image format detection (`png`, `jpeg`, `webp`, `gif`).
- Added smart output writer that:
  - detects actual output format
  - rewrites filename extension when needed
  - returns the final output path
- Updated all generation backends (`flux`, `nano-banana`, `gpt-image-1`, `nano-banana-pro`) to use smart writer.
- Updated `nano-banana-pro` reference-image MIME resolution to:
  - prefer content-based detection
  - fallback to extension when detection is unavailable
- Tracked the actual generated output path through post-processing (`remove-bg`, `add-bg`, `thumbnail`) so follow-up steps apply to the real file location.

## Validation

- Type/path-level validation by tracing all generation paths and ensuring post-processing uses `generatedOutput` returned by generator functions.
- Verified no behavior change when format already matches requested extension.
- Verified mismatch path now saves with correct extension and emits a warning.

Fixes #633

### DIFF
--- a/Packs/pai-art-skill/src/skills/Art/Tools/Generate.ts
+++ b/Packs/pai-art-skill/src/skills/Art/Tools/Generate.ts
@@ -16,7 +16,7 @@ import Replicate from "replicate";
 import OpenAI from "openai";
 import { GoogleGenAI } from "@google/genai";
 import { writeFile, readFile } from "node:fs/promises";
-import { extname, resolve } from "node:path";
+import { extname, resolve, dirname, basename, join } from "node:path";
 
 // ============================================================================
 // Environment Loading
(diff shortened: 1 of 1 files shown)


kaimagnus:
Thank you for this excellent fix, @VintLin! You correctly identified the core issue — the API sometimes returns a different image format than requested.

We've addressed this in commit 0fc1931 with a similar approach (magic byte detection, automatic extension correction, and proper MIME type handling throughout the pipeline). Closing in favor of that fix. Your analysis of the problem was spot-on and really helped confirm the right approach!

PR #644, opened 2026-02-13
Switch to spawnSync to prevent potential command injection in hooks

RinZ27:
Refactoring the `setTabTitle` function was necessary after noticing it relied on `execSync` with string templates and manual escaping. While single quotes were being escaped, placing the variable inside double quotes allowed shell expansion for characters like backticks or `$(...)`, which I identified as a potential command injection vector.

Moving to `spawnSync` with explicit argument arrays effectively neutralizes this risk by bypassing shell interpretation entirely. This change makes manual escaping redundant and provides a more robust security model for the hook system. Similar adjustments were applied to the voice announcement logic to maintain consistent safety patterns across the infrastructure.

### DIFF
--- a/Packs/pai-hook-system/src/hooks/UpdateTabTitle.hook.ts
+++ b/Packs/pai-hook-system/src/hooks/UpdateTabTitle.hook.ts
@@ -60,7 +60,7 @@
  * - Includes 5-second stdin timeout
  */
 
-import { execSync } from 'child_process';
+import { spawnSync } from 'child_process';
 import { readFileSync } from 'fs';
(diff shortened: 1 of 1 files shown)


kaimagnus:
Thank you for this important security fix, @RinZ27! Replacing execSync with spawnSync to prevent command injection is the right approach. The JSON.stringify for curl payloads is also a good improvement.

Unfortunately the target path (`Packs/pai-hook-system/`) no longer exists in v3.0. Closing as the paths don't match. We'll apply this security pattern to the v3.0 hook structure. Really appreciate the security-focused contribution!

PR #649, opened 2026-02-13
feat: session-scoped work tracking for multi-session safety

MarvinDontPanic:
## Summary

> **Depends on:** #648 — This PR includes the bug fixes from #648 plus the session-scoping feature. If #648 is merged first, this PR's diff will reduce to only the session-scoping changes.

When multiple Claude Code sessions run in parallel from the same home directory, they share a single `current-work.json` state file. Each session overwrites the file, so whichever session writes last wins — all other sessions lose their work tracking state.

This PR introduces session-scoped state files (`current-work-{session_id}.json`) so each session tracks its own work independently.

## Problem

`AutoWorkCreation.hook.ts` writes session state to a singleton file:

```
MEMORY/STATE/current-work.json
```

With parallel sessions A and B:
1. Session A starts → writes `current-work.json` (session A's data)
2. Session B starts → overwrites `current-work.json` (session B's data)
3. Session A ends → reads `current-work.json` → finds session B's data → either processes wrong session or skips

Result: Session A's work is never marked completed. Session B's work may be incorrectly processed by Session A's cleanup.

### AutoWorkCreation (write path)

State files are now session-scoped:

```typescript
function currentWorkFile(sessionId?: string): string {
  if (sessionId) return join(STATE_DIR, `current-work-${sessionId}.json`);
  return join(STATE_DIR, 'current-work.json'); // fallback for legacy reads
}
```

Creates: `MEMORY/STATE/current-work-{session_id}.json`

### SessionSummary + WorkCompletionLearning (read path)

Both SessionEnd hooks now:
1. Parse stdin for `session_id` (with timeout fallback)
2. Look up `current-work-{session_id}.json` first
3. Fall back to legacy `current-work.json` if no session-scoped file exists
4. Guard against processing another session's state

```typescript
function findStateFile(sessionId?: string): string | null {
  if (sessionId) {
    const scoped = join(STATE_DIR, `current-work-${sessionId}.json`);
    if (existsSync(scoped)) return scoped;
  }
  const legacy = join(STATE_DIR, 'current-work.json');
  if (existsSync(legacy)) return legacy;
  return null;
}
```

Plus a session ownership guard:

```typescript
if (sessionId && currentWork.session_id !== sessionId) {
  process.exit(0);
}
```

## Backwards Compatibility

- Falls back to `current-work.json` if no session-scoped file exists
- Works with pre-upgrade sessions that still use the singleton file
- No breaking changes to META.yaml, WORK/ directory structure, or learning output

## Files Changed

- `Releases/v2.5/.claude/hooks/AutoWorkCreation.hook.ts` — Session-scoped write path
- `Releases/v2.5/.claude/hooks/SessionSummary.hook.ts` — Full rewrite with session-scoped read + cleanup
- `Releases/v2.5/.claude/hooks/WorkCompletionLearning.hook.ts` — Session-scoped read

## Testing

Verified locally with 3+ concurrent sessions:
- Closing session A only deletes session A's state file ✓
- Sessions B and C state files remain untouched ✓
- META.yaml for session A marked COMPLETED, others stay ACTIVE ✓
- New sessions created after upgrade use scoped files ✓
- Legacy fallback works when session_id unavailable ✓

## Test plan

- [ ] Start two Claude Code sessions (A and B) from the same home directory
- [ ] Verify two separate `current-work-{id}.json` files exist in `MEMORY/STATE/`
- [ ] Close session A
- [ ] Verify session A's state file is deleted and META.yaml is COMPLETED
- [ ] Verify session B's state file and META.yaml are unchanged
- [ ] Close session B
- [ ] Verify session B's state file is deleted and META.yaml is COMPLETED

### DIFF
--- a/Releases/v2.5/.claude/hooks/AutoWorkCreation.hook.ts
+++ b/Releases/v2.5/.claude/hooks/AutoWorkCreation.hook.ts
@@ -52,7 +52,12 @@ interface PromptClassification {
 const BASE_DIR = process.env.PAI_DIR || join((process.env.HOME || process.env.USERPROFILE || homedir()), '.claude');
 const WORK_DIR = join(BASE_DIR, 'MEMORY', 'WORK');
 const STATE_DIR = join(BASE_DIR, 'MEMORY', 'STATE');
-const CURRENT_WORK_FILE = join(STATE_DIR, 'current-work.json');
+
+/** Session-scoped state file to avoid race conditions in multi-session setups */
+function currentWorkFile(sessionId?: string): string {
+  if (sessionId) return join(STATE_DIR, `current-work-${sessionId}.json`);
+  return join(STATE_DIR, 'current-work.json'); // fallback for legacy reads
+}
 
 const CLASSIFICATION_PROMPT = `Classify this user request. Return ONLY valid JSON:
 {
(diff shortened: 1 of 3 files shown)


kaimagnus:
Thank you for the session-scoped work tracking feature, @MarvinDontPanic! Multi-session safety is an important concern.

The target paths (Releases/v2.5/) have been superseded by v3.0. Closing as the paths no longer match. The concept is valuable — if you'd like to rebase against v3.0, we'd welcome it!

PR #668, opened 2026-02-15
fix: SessionEnd hooks never execute + session-scoped state files (v3.0)

MarvinDontPanic:
## Summary

> **Rebased from closed PRs #648 and #649** which targeted v2.5 paths. Per maintainer feedback, this PR targets the equivalent v3.0 hooks. All four bugs are identical in v3.0.

The `SessionSummary` and `WorkCompletionLearning` hooks (both triggered at `SessionEnd`) have never worked. Four independent bugs prevent them from executing their core logic. As a result:

- Work directories are never marked `COMPLETED` (all stay `ACTIVE` forever)
- `completed_at` timestamps are never set
- Work-completion learnings are never captured
- Session state files are never cleaned up

Additionally, parallel sessions overwrite each other's state via a shared singleton file.

### Bug 1: Stdin handling causes silent exit

Both `SessionSummary.hook.ts` and `WorkCompletionLearning.hook.ts` read stdin and bail if it's empty:

```typescript
const input = await Bun.stdin.text();
if (!input || input.trim() === '') {
  process.exit(0);  // <-- exits here every time
}
```

`SessionEnd` hooks may receive empty or slow stdin. The hooks exit before doing any work.

**Fix:** Use `Promise.race` with a 3-second timeout. Proceed regardless of stdin availability — these hooks read state from disk files, not stdin. (Note: `AutoWorkCreation.hook.ts` already uses this pattern via `readStdinWithTimeout`.)

### Bug 2: Field name mismatch (`work_dir` vs `session_dir`)

`AutoWorkCreation.hook.ts` writes state files with `session_dir`, but both `SessionSummary` and `WorkCompletionLearning` declare their `CurrentWork` interface with `work_dir` — always `undefined`.

**Fix:** Align interfaces to use `session_dir` (matching what `AutoWorkCreation` writes). Also updated `current_task`, `task_title`, `task_count` fields to match the v3.0 `CurrentWork` shape.

### Bug 3: Missing `completed_at` field in META.yaml template

`SessionSummary` tries to regex-replace `completed_at: null`, but `AutoWorkCreation` creates META.yaml without that field. The regex has nothing to match.

**Fix:** Added `completed_at: null` to the META.yaml template in `AutoWorkCreation.hook.ts`.

### Bug 4: Field name mismatch (`item_count` vs `task_count`)

`WorkCompletionLearning` checks `currentWork.item_count > 1` but `AutoWorkCreation` writes `task_count`. The condition is always false (`undefined > 1 === false`).

**Fix:** Changed `item_count` to `task_count`.

## Session-Scoped State Files (from #649)

When multiple Claude Code sessions run in parallel, they share a single `current-work.json` — last writer wins. This PR introduces session-scoped state files (`current-work-{session_id}.json`) so each session tracks independently.

- **Write path** (`AutoWorkCreation`): Creates `current-work-{session_id}.json`
- **Read path** (`SessionSummary`, `WorkCompletionLearning`): Looks up scoped file first, falls back to legacy `current-work.json`
- **Session guard**: Both SessionEnd hooks verify `session_id` matches before processing, preventing cross-session interference
- **Backwards compatible**: Falls back to legacy singleton if no scoped file exists

## Files Changed

- `Releases/v3.0/.claude/hooks/SessionSummary.hook.ts` — Fixes bugs 1, 2 + session-scoped reads
- `Releases/v3.0/.claude/hooks/WorkCompletionLearning.hook.ts` — Fixes bugs 1, 2, 4 + session-scoped reads
- `Releases/v3.0/.claude/hooks/AutoWorkCreation.hook.ts` — Fixes bug 3 + session-scoped writes

## Testing

Verified locally across multiple session closes:
- META.yaml transitions from `ACTIVE` to `COMPLETED`
- `completed_at` timestamp is set
- Session state files are cleaned up
- Work-completion learning significant-work check uses correct field
- Parallel sessions don't interfere with each other

## Test plan

- [ ] Start a new Claude Code session, submit a prompt (creates work directory)
- [ ] Verify META.yaml contains `completed_at: null`
- [ ] Close the session
- [ ] Verify META.yaml shows `status: "COMPLETED"` and `completed_at` has a timestamp
- [ ] Verify session-scoped `current-work-{id}.json` state file is deleted
- [ ] Start two parallel sessions, verify each gets its own state file
- [ ] Close one session, verify the other's state is untouched

### DIFF
--- a/Releases/v3.0/.claude/hooks/AutoWorkCreation.hook.ts
+++ b/Releases/v3.0/.claude/hooks/AutoWorkCreation.hook.ts
@@ -50,7 +50,11 @@ interface PromptClassification {
 const BASE_DIR = process.env.PAI_DIR || join(process.env.HOME!, '.claude');
 const WORK_DIR = join(BASE_DIR, 'MEMORY', 'WORK');
 const STATE_DIR = join(BASE_DIR, 'MEMORY', 'STATE');
-const CURRENT_WORK_FILE = join(STATE_DIR, 'current-work.json');
+// Session-scoped state files prevent parallel sessions from overwriting each other
+function currentWorkFile(sessionId?: string): string {
+  if (sessionId) return join(STATE_DIR, `current-work-${sessionId}.json`);
+  return join(STATE_DIR, 'current-work.json'); // legacy fallback
+}
 
 // No more inference — simple heuristic classification
 
@@ -64,10 +68,18 @@ async function readStdinWithTimeout(timeout: number = 5000): Promise<string> {
   });
 }
 
-function readCurrentWork(): CurrentWork | null {
+function readCurrentWork(sessionId?: string): CurrentWork | null {
   try {
-    if (!existsSync(CURRENT_WORK_FILE)) return null;
-    return JSON.parse(readFileSync(CURRENT_WORK_FILE, 'utf-8'));
+    // Try session-scoped file first, fall back to legacy
+    const scopedFile = sessionId ? currentWorkFile(sessionId) : null;
+    if (scopedFile && existsSync(scopedFile)) {
+      return JSON.parse(readFileSync(scopedFile, 'utf-8'));
+    }
+    const legacyFile = currentWorkFile();
+    if (existsSync(legacyFile)) {
+      return JSON.parse(readFileSync(legacyFile, 'utf-8'));
+    }
+    return null;
   } catch {
     return null;
   }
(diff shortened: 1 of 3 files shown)


kaimagnus:
Thank you @MarvinDontPanic for the excellent bug report and fix! Your analysis was spot on — all 4 bugs you identified were confirmed:

1. **Stdin handling** — `Bun.stdin.text()` silently exits on empty stdin at SessionEnd
2. **Field name mismatch** — `work_dir` vs `session_dir`, `item_count` vs `task_count`
3. **Missing `completed_at`** — META.yaml template didn't include it
4. **Session-scoped state files** — great improvement to prevent parallel session interference

We've applied these fixes in commit 5878b6a with one additional improvement: using `process.env.PAI_DIR || join(process.env.HOME!, '.claude')` instead of hardcoded `process.env.HOME!`, which follows PAI's variable convention for portability.

All fixes have been extensively tested:
- ✅ Empty stdin handling (Promise.race with 3s timeout)
- ✅ Session creation → session end lifecycle
- ✅ Parallel session isolation (end one, other untouched)
- ✅ Legacy `current-work.json` fallback
- ✅ META.yaml status transitions (ACTIVE → COMPLETED with timestamp)
- ✅ All 3 hooks compile clean

Closing in favor of our commit, but seriously — great catch. This was a real bug affecting every SessionEnd hook. 🙏

PR #681, opened 2026-02-16
fix: SessionHarvester cross-platform project directory slug

### LINKED ISSUE #680: Bug: SessionHarvester fails silently on Linux — hardcoded macOS project directory path
`SessionHarvester.ts` hardcodes a macOS-style project directory slug on line 30:
                                                                        
```ts                                                          
const PROJECTS_DIR = path.join(CLAUDE_DIR, "projects", `-Users-${USERNAME}--claude`);
```

Claude Code stores session transcripts under ~/.claude/projects/{cwd-slug}/, where the slug is derived
from the current working directory by replacing / and . with -. On macOS this produces
-Users-username--claude, but on Linux the home directory is /home/username, producing
-home-username--claude. The hardcoded macOS path causes the script to silently find no sessions and exit
with "No sessions found to harvest" on any non-macOS system.                                             
                                                                                                          
## Steps to Reproduce                                                 
                                                                        
1. Run on a Linux system where $HOME is /home/<user>                                                     
2. Execute bun run SessionHarvester.ts --recent 10                                                       
3. Output: No sessions found to harvest

Expected Behavior

SessionHarvester should detect the correct projects directory regardless of OS.

## Fix

Replace the hardcoded slug with dynamic derivation from CLAUDE_DIR:

```ts
const CWD_SLUG = CLAUDE_DIR.replace(/[\/\.]/g, "-");
const PROJECTS_DIR = path.join(CLAUDE_DIR, "projects", CWD_SLUG);
```

This produces the correct slug on both macOS (-Users-username--claude) and Linux
(-home-username--claude).

## Impact

- Severity: High — the tool is completely non-functional on Linux
- Scope: All Linux and WSL users
- Failure mode: Silent — no error message, just "No sessions found"

zulrang:
Fixes #680 

Replace hardcoded macOS path (-Users-username--claude) with dynamic slug derivation from CLAUDE_DIR, fixing silent failure on Linux/WSL where the home directory produces a different slug (-home-username--claude).

### DIFF
--- a/Releases/v3.0/.claude/skills/PAI/Tools/SessionHarvester.ts
+++ b/Releases/v3.0/.claude/skills/PAI/Tools/SessionHarvester.ts
@@ -27,7 +27,8 @@ import { getLearningCategory, isLearningCapture } from "../../../hooks/lib/learn
 
 const CLAUDE_DIR = path.join(process.env.HOME!, ".claude");
 const USERNAME = process.env.USER || require("os").userInfo().username;
-const PROJECTS_DIR = path.join(CLAUDE_DIR, "projects", `-Users-${USERNAME}--claude`);
+const CWD_SLUG = CLAUDE_DIR.replace(/[\/\.]/g, "-");
+const PROJECTS_DIR = path.join(CLAUDE_DIR, "projects", CWD_SLUG);
 const LEARNING_DIR = path.join(CLAUDE_DIR, "MEMORY", "LEARNING");
 
 // Patterns indicating learning moments in conversations


zulrang:
Still working on MacOS: 

Pasted image 20260216001848

kaimagnus:
Thanks for the PR! We independently fixed this in commit 95d65cc with an equivalent approach — deriving the slug dynamically from CLAUDE_DIR. Issue #680 is closed. Appreciate you taking the time to contribute! 🙏

PR #726, opened 2026-02-18
fix(SECUpdates): use direct WebFetch instead of agent spawning for source fetching

christauff:
## Summary
- Replace "Launch parallel agents" guidance in SKILL.md with direct WebFetch calls
- Add explicit anti-agent note to Update.md workflow
- Add risky.biz to SKILL.md source table (was in sources list but missing from Step 2)

## Problem
The SECUpdates SKILL.md Step 2 says "Launch parallel agents to fetch each source" with an "Agent 1/2/3..." table. This causes Claude to spawn Task agents for what should be direct WebFetch calls. Agent spawning adds 10-15 seconds overhead per source (permission prompts, context building, subprocess startup) and can silently hang — we observed a 17-minute stall with zero visible output.

## Fix
Replace agent language with direct WebFetch guidance and add an explicit implementation note. WebFetch is a direct tool call that returns in 2-5 seconds. Fetching 6 sources via direct WebFetch in two batches of 3 completes in under 15 seconds total vs. potentially unbounded with agents.

## Test plan
- [x] Ran SECUpdates with direct WebFetch: 6 sources, 32 items, ~15 seconds total
- [x] Confirmed no silent stalls
- [x] Output matches expected format from SKILL.md

### DIFF
--- a/Releases/v3.0/.claude/skills/SECUpdates/SKILL.md
+++ b/Releases/v3.0/.claude/skills/SECUpdates/SKILL.md
@@ -173,15 +173,18 @@ cat ~/.claude/skills/SECUpdates/State/last-check.json
 
 ### Step 2: Fetch Sources (Parallel)
 
-Launch parallel agents to fetch each source:
+Fetch all sources using direct parallel WebFetch calls (do NOT spawn agents via the Task tool — WebFetch is instant, agents add 10s+ overhead per source and can silently stall):
 
-| Agent | Source | Method |
-|-------|--------|--------|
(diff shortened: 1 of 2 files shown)


danielmiessler:
Thank you @christauff for the fix! The SECUpdates SKILL.md was completely rewritten in v4.0 and no longer uses the agent-spawning pattern. The current v4.0 version already uses direct WebFetch calls, so this issue is resolved.

See [Releases/v4.0.3](https://github.com/danielmiessler/Personal_AI_Infrastructure/tree/main/Releases/v4.0.3) for the latest.

PR #750, opened 2026-02-21
Merge duplicate first-principles rules in SYSTEM AISTEERINGRULES

MarvinDontPanic:
## Summary

- The SYSTEM `AISTEERINGRULES.md` file contains two rules that both target first-principles thinking:
  1. **"First principles and elegant problem-solving"** (line 5) — focuses on avoiding bolt-on solutions
  2. **"First Principles and Simplicity"** (line 75) — focuses on root-cause thinking with the "Understand → Simplify → Reduce → Add" order
- These are merged into a single rule that preserves all unique guidance from both
- Net result: one fewer instruction competing for model attention in the always-loaded steering rules

## Motivation

PAI's steering rules are loaded on every request. Duplicate rules waste tokens and create ambiguity about which rule governs first-principles behavior. Since both rules target the same concept with complementary examples, merging them into one comprehensive rule reduces instruction count without losing any behavioral guidance.

## Test plan

- [x] Verify merged rule preserves the "Understand → Simplify → Reduce → Add" order from the second rule
- [x] Verify merged rule preserves the anti-bolt-on/technical-debt guidance from the first rule
- [x] Verify both Bad/Correct examples are represented in the merged version
- [x] Verify no other rules reference either original by name

### DIFF
--- a/Releases/v3.0/.claude/skills/PAI/AISTEERINGRULES.md
+++ b/Releases/v3.0/.claude/skills/PAI/AISTEERINGRULES.md
@@ -2,10 +2,10 @@
 
 Universal behavioral rules for PAI. Mandatory. Personal customizations in `USER/AISTEERINGRULES.md` extend and override these.
 
-## First principles and elegant problem-solving
-**Statement:** Don't just randomly add bolt-on solutions for things when the user complains about something. Use the algorithm and proper PAI context to think about what's actually broken and needs to be fixed, and search for the most elegant and simple solution possible. We should not be accruing technical debt through bolted-on band-aid solutions. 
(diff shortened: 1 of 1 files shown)


danielmiessler:
Thanks for catching the duplication\! The AISTEERINGRULES were restructured in v4.0 — the steering rules are now split into SYSTEM and PERSONAL sections with a different structure. Closing as superseded by the v4.0 rewrite. If you spot remaining duplicates in the current version, please resubmit against [v4.0.2](https://github.com/danielmiessler/PAI/tree/main/Releases/v4.0.2). Thanks\!

PR #761, opened 2026-02-22
fix(statusline): read timezone from settings.json instead of hardcoding

### LINKED ISSUE #734: Statusline: timezone is hardcoded to America/Los_Angeles instead of reading from settings.json
## Problem

The `statusline-command.sh` script has the timezone hardcoded to `America/Los_Angeles` in two places (the `format_reset_time` and `clock_time` functions):

```python
local_dt = dt.astimezone(ZoneInfo('America/Los_Angeles'))
```

This means the 5h/7d usage reset times display in US Pacific time regardless of the user's location. For users outside the US, the reset times are confusing and don't match their local clock.

## Expected Behavior

The script should read the timezone from `settings.json` where users already configure it:

```json
"principal": {
  "timezone": "Europe/Berlin"
}
```

## Suggested Fix

Read the timezone dynamically from `settings.json` and pass it into the Python snippets, falling back to the system timezone or UTC if not configured:

```bash
USER_TZ=$(jq -r '.principal.timezone // "UTC"' "$SETTINGS_FILE" 2>/dev/null)
```

Then use `$USER_TZ` instead of the hardcoded `'America/Los_Angeles'` string in both Python blocks.

Ironically, line 596 already has a comment saying `# Uses TZ from settings.json (principal.timezone) for correct local time` — but the code below it doesn't actually do that.

## Affected Lines

- Line ~653 (`format_reset_time` function)
- Line ~923 (`clock_time` function)

nbost130:
## Summary

Reads `principal.timezone` from `settings.json` (with fallback to `America/Los_Angeles`) and passes it to the two Python timezone conversion blocks in the statusline script.

## Problem

The statusline script hardcodes `ZoneInfo('America/Los_Angeles')` in two places, causing reset times and clock displays to show Pacific time regardless of user's configured timezone.

## Fix

- Read `TZ_NAME` from `settings.json → principal.timezone` at startup
- Pass it into both Python `clock_time` / `format_reset_time` blocks
- Fallback to `America/Los_Angeles` if not configured (backwards compatible)

Closes #734

### DIFF
--- a/Releases/v3.0/.claude/statusline-command.sh
+++ b/Releases/v3.0/.claude/statusline-command.sh
@@ -69,6 +69,10 @@ DA_NAME="${DA_NAME:-Assistant}"
 PAI_VERSION=$(jq -r '.pai.version // "—"' "$SETTINGS_FILE" 2>/dev/null)
 PAI_VERSION="${PAI_VERSION:-—}"
 
+# Get principal timezone from settings (fallback to America/Los_Angeles for backwards compat)
+TZ_NAME=$(jq -r '.principal.timezone // "America/Los_Angeles"' "$SETTINGS_FILE" 2>/dev/null)
+TZ_NAME="${TZ_NAME:-America/Los_Angeles}"
+
 # Get Algorithm version from LATEST file (single source of truth)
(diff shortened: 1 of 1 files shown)


nbost130:
Closing in favor of #762 by @jlacour-git which landed the same fix with a UTC fallback (better universal default). Great minds think alike!

PR #762, opened 2026-02-22
fix(statusline): read timezone from settings.json instead of hardcoding

### LINKED ISSUE #734: Statusline: timezone is hardcoded to America/Los_Angeles instead of reading from settings.json

jlacour-git:
## Summary

- Replaces hardcoded `America/Los_Angeles` timezone in `statusline-command.sh` with dynamic read from `settings.json` (`principal.timezone`)
- Falls back to UTC if no timezone is configured
- Affects both `reset_clock_time()` (line ~653) and inline `clock_time()` (line ~923)

The comment on line 596 already says "Uses TZ from settings.json" — this PR makes the code match the comment.

Closes #734

## Test plan

- [ ] Verify reset times display in configured timezone (e.g., `Europe/Berlin`)
- [ ] Verify fallback to UTC when `principal.timezone` is not set
- [ ] Verify no regression in `time_until` calculations (those use UTC diff, unaffected)

### DIFF
--- a/Releases/v3.0/.claude/statusline-command.sh
+++ b/Releases/v3.0/.claude/statusline-command.sh
@@ -69,6 +69,10 @@ DA_NAME="${DA_NAME:-Assistant}"
 PAI_VERSION=$(jq -r '.pai.version // "—"' "$SETTINGS_FILE" 2>/dev/null)
 PAI_VERSION="${PAI_VERSION:-—}"
 
+# Get user timezone from settings (for reset time display)
+USER_TZ=$(jq -r '.principal.timezone // empty' "$SETTINGS_FILE" 2>/dev/null)
+USER_TZ="${USER_TZ:-UTC}"
+
 # Get Algorithm version from LATEST file (single source of truth)
 ALGO_LATEST_FILE="$PAI_DIR/skills/PAI/Components/Algorithm/LATEST"
 if [ -f "$ALGO_LATEST_FILE" ]; then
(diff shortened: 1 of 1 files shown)


kaimagnus:
Thanks for catching this — the fix is incorporated in v4.0.1. The statusline now reads `principal.timezone` from `settings.json` with a UTC fallback. Closing as the fix is included in the release.

PR #780, opened 2026-02-23
fix(statusline): fix f-string syntax error silently breaking reset time display

### LINKED ISSUE #779: bug(statusline): f-string syntax error silently breaks reset time display for all OAuth users
## Problem

The `statusline-command.sh` batch Python eval block (around line 932) contains a syntax error that silently kills all four reset-time variables.

**Affected lines:**
```python
print(f"clock_5h='{clock_time(r5h, "hourly")}'")
print(f"clock_7d='{clock_time(r7d, "weekly")}'")
```

Python f-strings (pre-3.12) cannot contain the same quote character as the f-string delimiter. `"hourly"` inside `f"..."` is a `SyntaxError`.

Because the invocation uses `2>/dev/null`, the error is silently swallowed:
```bash
eval "$(python3 -c "...the whole block..." 2>/dev/null)"
```

The entire eval produces **no output**. All four variables are left unset:
- `reset_5h` → falls back to `—`
- `reset_7d` → falls back to `—`
- `clock_5h` → empty
- `clock_7d` → empty

**Result:** Any user with a valid OAuth token sees `↻—` instead of a real reset time in the usage statusline.

## Steps to Reproduce

1. Have a valid Anthropic OAuth token in the macOS Keychain
2. Observe the usage line in the statusline: `▰ USAGE: 5H: 6% ↻— │ WK: 12% ↻—`
3. The `resets_at` fields are present in the usage cache — the Python eval just fails silently

You can verify the failure:
```bash
# Simulate the eval without 2>/dev/null — shows the SyntaxError
python3 -c "print(f\"test='{chr(34)}nested{chr(34)}'\")  " 2>&1
```

## Fix

Use single-quoted string literals inside the double-quoted f-strings. Single quotes inside a `f"..."` f-string are valid in Python 3.6+:

```python
print(f"clock_5h='{clock_time(r5h, 'hourly')}'")
print(f"clock_7d='{clock_time(r7d, 'weekly')}'")
```

PR with fix: coming.

jlacour-git:
## Summary

- Fixes a Python f-string `SyntaxError` in the batch usage reset time eval that silently breaks all four reset-time variables (`reset_5h`, `reset_7d`, `clock_5h`, `clock_7d`) for **all users with valid OAuth tokens**
- The error is swallowed by `2>/dev/null`, so users see `↻—` instead of a real reset time with no indication anything is wrong

Closes #779

## Root Cause

```python
# Before — SyntaxError: double quotes inside double-quoted f-string (Python < 3.12)
print(f"clock_5h='{clock_time(r5h, "hourly")}'")
print(f"clock_7d='{clock_time(r7d, "weekly")}'")
```

Because the entire `python3 -c "..."` invocation is wrapped in `2>/dev/null`, the `SyntaxError` is silently discarded. The `eval` receives no output, all variables stay empty, and the statusline falls back to `—` for every reset time.

## Fix

```python
# After — single quotes inside double-quoted f-string, valid Python 3.6+
print(f"clock_5h='{clock_time(r5h, 'hourly')}'")
print(f"clock_7d='{clock_time(r7d, 'weekly')}'")
```

Single-quoted string literals inside a `f"..."` f-string are valid since Python 3.6. No behavior change — just fixes the silent crash.

## Test Plan

- [ ] Confirm `reset_5h` and `reset_7d` are populated after the eval (e.g. `4h32m`, `2d23h`)
- [ ] Confirm statusline usage line shows `↻4h32m` instead of `↻—`
- [ ] Verify no regression when OAuth token is absent (falls back to `—` as before)

### DIFF
--- a/Releases/v3.0/.claude/statusline-command.sh
+++ b/Releases/v3.0/.claude/statusline-command.sh
@@ -929,8 +929,8 @@ r5h = '$usage_5h_reset'
 r7d = '$usage_7d_reset'
 print(f\"reset_5h='{time_until(r5h)}'\")
 print(f\"reset_7d='{time_until(r7d)}'\")
-print(f\"clock_5h='{clock_time(r5h, \"hourly\")}'\")
-print(f\"clock_7d='{clock_time(r7d, \"weekly\")}'\")
+print(f\"clock_5h='{clock_time(r5h, 'hourly')}'\")
+print(f\"clock_7d='{clock_time(r7d, 'weekly')}'\")
 " 2>/dev/null)"
     reset_5h="${reset_5h:-—}"
     reset_7d="${reset_7d:-—}"


kaimagnus:
Good catch on the nested quoting — this is fixed in v4.0.1. Changed the escaped double quotes to single quotes inside the f-strings. Closing as the fix is included in the release.

PR #778, opened 2026-02-23
fix: Add PAI_DIR fallback to 26 runtime files with hardcoded $HOME/.claude paths

neilinger:
## Summary

- 26 files in `Releases/v3.0/.claude/` hardcode `$HOME/.claude` without checking `PAI_DIR`, breaking installations that use a non-default location
- Adds the standard `process.env.PAI_DIR || join(HOME, '.claude')` fallback pattern — the same pattern already used by hooks like `AlgorithmTracker`, `SessionSummary`, and `algorithm.ts`
- Files that already had `PAI_DIR` are left untouched

## Problem

Issue #694 identified that `CLAUDE_CONFIG_DIR` wasn't respected. The fix in #696 addressed only `INSTALL.ts`. The remaining 26 runtime files (hooks, skills, tools, actions, lib) still resolve paths via `$HOME/.claude` directly, ignoring `PAI_DIR` from `settings.json`.

This means any PAI installation not at `~/.claude` (e.g. project-scoped `.claude/` directories) has broken path resolution at runtime across banners, skill configs, action runners, session progress, feature registry, and more.

## Changes

**Pattern applied consistently:**
```typescript
// Before (hardcoded):
const CLAUDE_DIR = join(HOME, ".claude");

// After (respects PAI_DIR):
const CLAUDE_DIR = process.env.PAI_DIR || join(HOME, ".claude");
```

**Affected areas:**
- `hooks/` — LoadContext, RebuildSkill handler, identity lib
- `lib/migration/` — scanner now checks PAI_DIR location first
- `skills/PAI/Tools/` — Banner variants, ActivityParser, FeatureRegistry, SessionProgress, SessionHarvester, LearningPatternSynthesis, LoadSkillConfig, RebuildPAI
- `skills/` — Agents/ComposeAgent, Browser/Browse, Evals/AlgorithmBridge, PAIUpgrade/Anthropic, Recon/BountyPrograms, Telos/UpdateTelos, WebAssessment/config
- `ACTIONS/` — CORE and PAI runner.v2 (Inference import), social/post.action

Closes #694

PR #784, opened 2026-02-23
fix: PAI_DIR fallback with path.join() normalization

### LINKED ISSUE #694: Enhancement: Installer should set PAI_DIR to match CLAUDE_CONFIG_DIR when using custom config directory
## Summary

When PAI is installed into a custom `CLAUDE_CONFIG_DIR` location (e.g., `~/alva-pai/.claude`), 
the installer always writes `PAI_DIR` pointing to `~/.claude` (the default). This creates a 
split-brain where Claude's config home and PAI's runtime home are different directories, causing 
MEMORY, hooks, and session data to accumulate in two separate locations.

## Environment

- macOS 15 (Darwin 25.3.0)
- Claude Code (Homebrew)
- PAI v2.5
- Shell: zsh

## Problem

**File:** `INSTALL.ts` / `settings.json`

The installer writes:
```json
"PAI_DIR": "/Users/<username>/.claude"
```

But when `CLAUDE_CONFIG_DIR` is set to a different location (e.g., `/Users/<username>/alva-pai/.claude`), 
this creates a mismatch:

| What | Points to |
|------|-----------|
| `CLAUDE_CONFIG_DIR` | `~/alva-pai/.claude` (Claude reads config, stores sessions here) |
| `PAI_DIR` in settings.json | `~/.claude` (hooks, MEMORY, skills resolve here) |

**Impact:**
- MEMORY/WORK entries accumulate in `~/.claude/MEMORY/` instead of the actual config home
- Debug logs accumulate in `~/alva-pai/.claude/debug/` (Claude's home) while MEMORY goes elsewhere
- Session data is split across two directories
- Two parallel copies of hooks/skills exist but only one is active
- Users lose track of which directory is canonical

## Reproduction

```bash
# Set a custom config directory
export CLAUDE_CONFIG_DIR="$HOME/custom-pai/.claude"

# Install PAI
# ... installer runs ...

# Check the mismatch
grep PAI_DIR "$CLAUDE_CONFIG_DIR/settings.json"
# Output: "PAI_DIR": "/Users/<username>/.claude"  ← wrong!
# Should be: "PAI_DIR": "/Users/<username>/custom-pai/.claude"
```

## Suggested Fix

In `INSTALL.ts`, detect `CLAUDE_CONFIG_DIR` and use it as the value for `PAI_DIR`:

```typescript
const paiDir = process.env.CLAUDE_CONFIG_DIR || path.join(os.homedir(), '.claude');
// Write to settings.json:
// "PAI_DIR": paiDir
```

This ensures that when users intentionally place their config in a custom location 
(for isolation, git tracking, etc.), all PAI runtime paths resolve to that same location.

## Related Issues

- #549 — Fixed `$HOME` literal string not being expanded (different root cause, same symptom class)
- #612 — INSTALL.ts overwriting settings.json values

## Use Case Context

Users may want `CLAUDE_CONFIG_DIR` redirection for:
- **Git-tracked PAI installations** (version control the entire `.claude/` directory)
- **Multi-environment isolation** (PAI vs vanilla Claude vs legacy)
- **Team-shared configurations** (project-local `.claude/` directories)

bitfieldz:
## Summary

- Adds `PAI_DIR` env var fallback to all 26 runtime files with hardcoded `$HOME/.claude` paths (same scope as #778 by @neilinger)
- Normalizes **all** path constructions to use `path.join()` instead of template literal interpolation

## Why `join()` over template literals?

`path.join()` from `node:path` provides:
- **Separator normalization** — handles double slashes if a dir value ends with `/`
- **Segment resolution** — resolves `.` and `..` components
- **Consistency** — single pattern across the entire codebase

Template literals like `` `${BASE_DIR}/skills/...` `` don't normalize, so `PAI_DIR=/foo/bar/` produces `/foo/bar//skills/...`.

Six files from the original changeset used template literals and have been converted:
- `ComposeAgent.ts`, `Browse.ts`, `post.action.ts`
- `AlgorithmBridge.ts`, `BountyPrograms.ts`, `WebAssessment/config.ts`

## Relationship to #778

This PR extends @neilinger's work in #778 with the `join()` normalization layer. All 26 files from #778 are included. If #778 merges first, this PR can be rebased to only include the `join()` conversions.

Closes #694

Co-Authored-By: Neil Scholten <neil@scholten.io>

### DIFF
--- a/Releases/v3.0/.claude/hooks/LoadContext.hook.ts
+++ b/Releases/v3.0/.claude/hooks/LoadContext.hook.ts
@@ -442,8 +442,8 @@ async function checkActiveProgress(paiDir: string): Promise<string | null> {
     }
   }
 
-  summary += '\n💡 To resume project: `bun run ~/.claude/skills/PAI/Tools/SessionProgress.ts resume <project>`\n';
-  summary += '💡 To complete project: `bun run ~/.claude/skills/PAI/Tools/SessionProgress.ts complete <project>`\n';
+  summary += `\n💡 To resume project: \`bun run ${join(paiDir, 'skills/PAI/Tools/SessionProgress.ts')} resume <project>\`\n`;
+  summary += `💡 To complete project: \`bun run ${join(paiDir, 'skills/PAI/Tools/SessionProgress.ts')} complete <project>\`\n`;
 
   return summary;
 }
@@ -520,7 +520,7 @@ async function main() {
     if (needsRebuild) {
       console.error('🔨 Rebuilding SKILL.md (components changed)...');
       try {
-        execSync('bun ~/.claude/skills/PAI/Tools/RebuildPAI.ts', {
+        execSync(`bun ${join(paiDir, 'skills/PAI/Tools/RebuildPAI.ts')}`, {
           cwd: paiDir,
           stdio: 'pipe',
           timeout: 5000
--- a/Releases/v3.0/.claude/hooks/LoadContext.hook.ts
+++ b/Releases/v3.0/.claude/hooks/LoadContext.hook.ts
@@ -520,7 +520,7 @@ async function main() {
     if (needsRebuild) {
       console.error('🔨 Rebuilding SKILL.md (components changed)...');
       try {
-        execSync('bun ~/.claude/skills/PAI/Tools/RebuildPAI.ts', {
+        execSync(`bun ${paiDir}/skills/PAI/Tools/RebuildPAI.ts`, {
           cwd: paiDir,
           stdio: 'pipe',
           timeout: 5000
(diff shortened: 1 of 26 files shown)


danielmiessler:
Hey @Spirotot — Kai here, Daniel's AI assistant, doing the next pass of triage on the PAI repo.

Closing this out because v5.0.0 already ships consolidated path resolution at `hooks/lib/paths.ts` — `getPaiDir()`, `getClaudeDir()`, and `paiPath(...)` honor `PAI_DIR` (and `CLAUDE_CONFIG_DIR`) and are the single canonical entry points used across hooks and tools. The intent of this PR is in v5; the specific edits are against pre-consolidation v4 paths and won't merge cleanly.

**If hardcoded paths are still slipping through against v5.0.0, please open a fresh PR against the new release pointing to the specific files.** Thanks for taking the time to put this together — really appreciated.

— Kai 🤖

PR #794, opened 2026-02-24
Update README.md to fix skills copy in upgrade instructions

dancornell:


danielmiessler:
Thanks! The upgrade instructions were rewritten in v4.0.1 with dedicated Fresh Install and Upgrading sections. Closing as addressed — see the [current README](https://github.com/danielmiessler/PAI#upgrading-from-a-previous-version). Appreciate the fix!

PR #800, opened 2026-02-25
fix: handle JSON array responses in Inference.ts

rikitikitavi2012-debug:
## Summary

- Fix JSON parsing in `Inference.ts` `--json` mode to handle both object and array responses
- The existing greedy regex `/\{[\s\S]*\}/` fails when the LLM returns a JSON array (`[...]`), which silently breaks 4+ downstream tools

## Problem

When `expectJson: true` is set, `Inference.ts` uses a single regex to extract JSON:
```ts
const jsonMatch = output.match(/\{[\s\S]*\}/);
```

This fails in two ways:
1. **JSON arrays are missed entirely** — `[{...}, {...}]` doesn't match the object regex
2. **Greedy capture produces invalid JSON** — when LLM wraps output in markdown with braces, the regex captures too much

Affected tools that use `--json` with array-returning prompts:
- `WisdomExtractor.ts` — expects `[{domain, observation, type}]`
- `TELOSTracker.ts` — expects `[{goal, status, evidence}]`
- `FailureCapture.ts` — expects `[{pattern, category}]`
- `IntegrityMaintenance.ts` — expects `[{file, issue}]`

All four tools silently fail because `Inference.ts` returns `{ success: false, error: 'No JSON found' }` and the callers proceed without data.

## Fix

Try both object and array regex matches, validate each candidate with `JSON.parse`, use whichever parses successfully:

```ts
const objectMatch = output.match(/\{[\s\S]*\}/);
const arrayMatch = output.match(/\[[\s\S]*\]/);

for (const candidate of [objectMatch?.[0], arrayMatch?.[0]]) {
  if (!candidate) continue;
  try {
    const parsed = JSON.parse(candidate);
    // success — use this candidate
  } catch { /* try next candidate */ }
}
```

Object match is tried first (most common case), then array match as fallback.

## Test plan

- [x] Verified with prompts returning `{...}` objects — still works
- [x] Verified with prompts returning `[{...}, {...}]` arrays — now works
- [x] Verified with markdown-wrapped JSON — greedy capture validated by parse

### DIFF
--- a/Releases/v3.0/.claude/skills/PAI/Tools/Inference.ts
+++ b/Releases/v3.0/.claude/skills/PAI/Tools/Inference.ts
@@ -133,10 +133,18 @@ export async function inference(options: InferenceOptions): Promise<InferenceRes
 
       // Parse JSON if requested
       if (options.expectJson) {
-        const jsonMatch = output.match(/\{[\s\S]*\}/);
-        if (jsonMatch) {
+        // Try both object and array matches — use whichever parses successfully.
+        // The greedy object regex /\{[\s\S]*\}/ can capture invalid substrings
+        // when the LLM wraps a JSON array inside markdown or explanatory text
+        // that happens to contain braces. By trying both candidates and
+        // validating with JSON.parse, we handle arrays and objects reliably.
+        const objectMatch = output.match(/\{[\s\S]*\}/);
+        const arrayMatch = output.match(/\[[\s\S]*\]/);
+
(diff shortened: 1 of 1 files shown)


danielmiessler:
Thank you @rikitikitavi2012-debug! This fix is merged into PAI v4.0.3. The JSON array parsing improvement is solid — trying both object and array candidates with validation is the right approach.

**Fix included in:** [Releases/v4.0.3](https://github.com/danielmiessler/Personal_AI_Infrastructure/tree/main/Releases/v4.0.3)

PR #806, opened 2026-02-26
fix(statusline): remove broken context percentage fallback after /clear

### LINKED ISSUE #805: bug(statusline): CONTEXT shows 100% after /clear instead of 0%
## Bug

After running `/clear`, the statusline CONTEXT bar shows **100%** instead of the expected **~0%**. The context window is empty after `/clear`, but the display suggests it's full.

## Root Cause

`statusline-command.sh` lines 102-107 contain a fallback that fires when `used_percentage` is 0:

```bash
if [ "$context_pct" = "0" ] && [ "$total_input" -gt 0 ]; then
    total_tokens=$((total_input + total_output))
    context_pct=$((total_tokens * 100 / context_max))
fi
```

After `/clear`, Claude Code's statusline JSON reports:
- `used_percentage: null` (correctly — context is empty) → jq defaults to `0`
- `total_input_tokens: 531593` (cumulative session total — **not reset** by /clear)
- `total_output_tokens: 27335` (cumulative — **not reset**)

The fallback sees `context_pct=0` and `total_input=531593 > 0`, so it calculates: `(531593 + 27335) * 100 / 200000 = 279%` → capped at 100%.

## Why the Fallback Is Wrong in All Cases

The comment says it handles "statusLine called before percentage is populated." But `total_input_tokens` and `total_output_tokens` are **cumulative session totals** (like an odometer) — they track all tokens ever processed, not what's currently in the context window. They routinely exceed `context_window_size` in any session longer than a few turns.

There is no scenario where dividing cumulative totals by window size produces the correct context percentage. The field that represents current context fullness is `used_percentage` — that's what should be used (and is, when it's non-null).

## Evidence

Captured via a debug wrapper around `statusline-command.sh`:

**Right after /clear (10:32:44):**
```json
{
  "total_input_tokens": 531593,
  "total_output_tokens": 27335,
  "context_window_size": 200000,
  "current_usage": null,
  "used_percentage": null,
  "remaining_percentage": null
}
```

**Normal mid-session (for comparison):**
```json
{
  "total_input_tokens": 587490,
  "total_output_tokens": 37014,
  "context_window_size": 200000,
  "current_usage": { "input_tokens": 1, "output_tokens": 1, ... },
  "used_percentage": 37,
  "remaining_percentage": 63
}
```

## Fix

Remove the fallback entirely. When `used_percentage` is null, the jq default of `0` is the correct display value. PR incoming.

## Affected Versions

Present in v2.5 and v3.0 (at minimum).

jlacour-git:
## Summary

- Removes the fallback calculation in `statusline-command.sh` that incorrectly computes context percentage from cumulative session totals
- After `/clear`, CONTEXT now correctly shows ~0% instead of 100%

## Root Cause

The fallback (lines 102-107) fires when `used_percentage` is null (jq defaults to 0) and `total_input_tokens > 0`. It divides cumulative session totals by `context_window_size` — but cumulative totals are an odometer, not a speedometer. After `/clear`, they retain pre-clear values (e.g. 531K tokens) while the actual context is empty: `531K/200K = 265%` → capped at 100%.

The fallback is wrong in all cases because cumulative totals never represent current context window contents. Full analysis in #805.

## Change

Replaced the 6-line fallback block with a 5-line comment explaining why it was removed, so future maintainers don't reintroduce it.

## Test plan

- [ ] Run `/clear` — CONTEXT bar should show 0%, not 100%
- [ ] Normal mid-session — CONTEXT bar unchanged (uses `used_percentage` directly)
- [ ] Long session where `total_input_tokens` exceeds `context_window_size` — no incorrect percentage

Closes #805

### DIFF
--- a/Releases/v3.0/.claude/statusline-command.sh
+++ b/Releases/v3.0/.claude/statusline-command.sh
@@ -99,12 +99,11 @@ context_remaining=${context_remaining:-100}
 total_input=${total_input:-0}
 total_output=${total_output:-0}
 
-# If used_percentage is 0 but we have token data, calculate manually
-# This handles cases where statusLine is called before percentage is populated
-if [ "$context_pct" = "0" ] && [ "$total_input" -gt 0 ]; then
-    total_tokens=$((total_input + total_output))
-    context_pct=$((total_tokens * 100 / context_max))
-fi
+# NOTE: Removed fallback that calculated context_pct from total_input + total_output
+# when used_percentage was 0. total_input/output_tokens are CUMULATIVE session totals
+# (like an odometer) — they can far exceed context_window_size. After /clear,
+# used_percentage is null (jq defaults to 0) but totals retain pre-clear values,
+# producing e.g. 531K/200K = 265% capped to 100%. See issue for full analysis.
 
 # Get Claude Code version
 if [ -n "$cc_version_json" ] && [ "$cc_version_json" != "unknown" ]; then


kaimagnus:
Thanks — the broken context fallback is removed in v4.0.1. We also removed the self-calibrating startup estimate that was causing inflated context percentages on fresh sessions. Closing as the fix is included in the release.

PR #821, opened 2026-02-28
feat: configurable temperature unit via TEMPERATURE_UNIT in .env

### LINKED ISSUE #820: feat: Allow users to configure temperature unit (°F / °C) via .env
## Problem

The statusline weather display hardcodes Celsius in two places in
`.claude/statusline-command.sh`, with no way for users to override it:

**Line 353 — Open-Meteo API call:**
```bash
"...&temperature_unit=celsius"   # hardcoded

Cache write (~line 368):
echo "${temp}°C ${condition}" > "$WEATHER_CACHE"  # hardcoded label

Users in the US (and other countries using Fahrenheit) have no way to
change this without manually editing statusline-command.sh, which gets
overwritten on updates.

Proposed Fix

Add TEMPERATURE_UNIT to .env.example alongside the existing TIME_ZONE
variable, since .env is already the established single source of truth for
user configuration. The statusline already sources .env at startup, so no
new loading mechanism is needed.

.env.example addition:
# Temperature unit for weather display: celsius or fahrenheit
TEMPERATURE_UNIT=celsius

.claude/statusline-command.sh change (~3 lines near line 353):
TEMP_UNIT="${TEMPERATURE_UNIT:-celsius}"
TEMP_LABEL="°C"
[ "$TEMP_UNIT" = "fahrenheit" ] && TEMP_LABEL="°F"

Then substitute ${TEMP_UNIT} in the API URL and ${TEMP_LABEL} in the
cache write.

Why This Approach

  - Follows the existing .env pattern (same place as DA, TIME_ZONE)
  - No new dependencies — env var is already available after .env is sourced
  - Backward compatible — defaults to celsius, matching current behavior
  - Small diff — 3 new lines + 2 substitutions in statusline-command.sh

Open Questions

 - Should the installer prompt for this during setup, or just document it in
 .env.example for users to set manually?
 - Is there a better location for this config than .env?

Happy to submit a PR with this fix — raising the issue first in case there's
a cleaner approach or something already planned.

sdanelson:
Closes #820

Added TEMPERATURE_UNIT=celsius to .env.example and updated statusline-command.sh to read it, defaulting to celsius for backward compatibility. Users can set TEMPERATURE_UNIT=fahrenheit in their .env to switch units.

Testing: Set TEMPERATURE_UNIT=fahrenheit in .env, cleared weather cache, confirmed statusline shows °F. Set to celsius, confirmed °C.

### DIFF
--- a/.claude/statusline-command.sh
+++ b/.claude/statusline-command.sh
@@ -350,7 +350,10 @@ fetch_weather() {
         lon="${lon:-122.4194}"
 
         # Fetch from Open-Meteo (free, fast, no API key)
-        local weather_json=$(curl -s --max-time 3 "https://api.open-meteo.com/v1/forecast?latitude=${lat}&longitude=${lon}&current=temperature_2m,weather_code&temperature_unit=celsius" 2>/dev/null)
+        local temp_unit="${TEMPERATURE_UNIT:-celsius}"
+	local temp_label="°C"
+	[ "$temp_unit" = "fahrenheit" ] && temp_label="°F"
+	local weather_json=$(curl -s --max-time 3 "https://api.open-meteo.com/v1/forecast?latitude=${lat}&longitude=${lon}&current=temperature_2m,weather_code&temperature_unit=${temp_unit}" 2>/dev/null)
         if [ -n "$weather_json" ] && echo "$weather_json" | jq -e '.current' >/dev/null 2>&1; then
             local temp=$(echo "$weather_json" | jq -r '.current.temperature_2m' 2>/dev/null)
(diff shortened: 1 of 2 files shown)


danielmiessler:
Thanks for this PR! The issue this addresses has been fixed in the v4.0.x releases:

- [v4.0.1](https://github.com/danielmiessler/PAI/tree/main/Releases/v4.0.1) — Upgrade path, temperature unit, statusline fixes
- [v4.0.2](https://github.com/danielmiessler/PAI/tree/main/Releases/v4.0.2) — 13 bug fixes including Linux compatibility, hook fixes, installer improvements

Closing as addressed. If there are gaps not covered by the releases, please feel free to resubmit against the latest release. We appreciate the contribution!

PR #822, opened 2026-02-28
feat: Add Linux systemd support to VoiceServer (cross-platform auto-start)

espcoder:
## Summary

The VoiceServer scripts only supported macOS (launchctl/LaunchAgents). Linux users had no auto-start mechanism, so the voice server would stop working after every reboot or session end.

This PR adds full Linux support via **systemd user services** — no root access required.

### All 6 VoiceServer scripts updated with OS auto-detection (`uname -s`):

| Script | macOS | Linux |
|--------|-------|-------|
| `install.sh` | Creates `~/Library/LaunchAgents/*.plist` | Creates `~/.config/systemd/user/pai-voice-server.service` |
| `start.sh` | `launchctl load` | `systemctl --user start` |
| `stop.sh` | `launchctl unload` | `systemctl --user stop` |
| `restart.sh` | stop → start | `systemctl --user restart` |
| `status.sh` | `launchctl list` + `~/Library/Logs` | `systemctl --user status` + `journalctl` |
| `uninstall.sh` | Removes `.plist` | Removes `.service`, reloads daemon |

### Key Linux behaviors:
- **Log path**: `~/.local/share/pai/voice-server.log` (XDG-compliant)
- **Boot persistence**: `loginctl enable-linger` so service starts before first login
- **Port check**: Uses `ss` as fallback when `lsof` is unavailable
- **No root required**: All systemd operations use `--user` flag
- **macOS unchanged**: All macOS behavior is 100% preserved

Also includes the full `VoiceServer/` directory which was not previously tracked in this repo (the scripts existed only in local installations).

## Test Plan

- [ ] Linux: `./install.sh` creates service file at `~/.config/systemd/user/pai-voice-server.service`
- [ ] Linux: `systemctl --user status pai-voice-server` shows active after install
- [ ] Linux: Service auto-starts after reboot (with `loginctl enable-linger`)
- [ ] Linux: `./stop.sh`, `./start.sh`, `./restart.sh`, `./status.sh`, `./uninstall.sh` all work
- [ ] macOS: All existing behavior unchanged (`launchctl`, `LaunchAgents`, `~/Library/Logs`)
- [ ] Both: `curl http://localhost:8888/health` returns healthy response after install

**Tested on:** Ubuntu 24.04 LTS

Closes the Linux auto-start gap documented in `PLATFORM.md` item #3.

### DIFF
--- a/VoiceServer/install.sh
+++ b/VoiceServer/install.sh
@@ -1,0 +1,276 @@
+#!/bin/bash
+
+# Voice Server Installation Script
+# Supports macOS (LaunchAgent) and Linux (systemd user service)
+
+set -e
+
+# Colors for output
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+BLUE='\033[0;34m'
+NC='\033[0m' # No Color
+
+# Configuration
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+ENV_FILE="$HOME/.env"
+OS=$(uname -s)
+
+# OS-specific paths
+if [[ "$OS" == "Darwin" ]]; then
+    SERVICE_NAME="com.pai.voice-server"
+    PLIST_PATH="$HOME/Library/LaunchAgents/${SERVICE_NAME}.plist"
+    LOG_PATH="$HOME/Library/Logs/pai-voice-server.log"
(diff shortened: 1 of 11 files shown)


espcoder:
Closing — this PR adds the entire VoiceServer directory which was never in the upstream repo, making the diff much larger than intended for a focused Linux support fix. Will revisit with a more focused approach once aligned with the project maintainer on the right contribution path.

PR #826, opened 2026-02-28
feat: Add VoiceServer with cross-platform Linux + macOS support

espcoder:
## Summary

Adds the VoiceServer component with full support for both macOS and Linux. All 6 management scripts auto-detect the OS via `uname -s` — no manual configuration needed.

- **macOS**: `launchctl` / `~/Library/LaunchAgents` (existing pattern preserved unchanged)
- **Linux**: `systemctl --user` / `~/.config/systemd/user/pai-voice-server.service`

## Platform Details

| | macOS | Linux |
|--|-------|-------|
| Service manager | launchctl / LaunchAgent plist | systemd user service |
| Log path | `~/Library/Logs/pai-voice-server.log` | `~/.local/share/pai/voice-server.log` |
| Auto-start on boot | RunAtLoad plist key | `loginctl enable-linger` |
| Root required | No | No |

## Scripts included

`install.sh` · `start.sh` · `stop.sh` · `restart.sh` · `status.sh` · `uninstall.sh`

## Implementation notes

- Port cleanup is scoped to PAI `bun`/`server.ts` processes only — never kills unrelated processes on port 8888
- Graceful shutdown: SIGTERM with 3s grace period before SIGKILL fallback
- API key parsing uses `cut -d'=' -f2-` to correctly handle base64 keys containing `=`
- `ss` port check uses portable `grep -o` instead of `grep -oP` (works on Alpine/RHEL/older systems)
- `loginctl enable-linger` failure produces a visible warning with fix instructions instead of silently swallowing
- Single `curl` call in `status.sh` (response cached, not fetched twice)

## Test Plan

- [ ] Linux: `./install.sh` creates `~/.config/systemd/user/pai-voice-server.service`
- [ ] Linux: `systemctl --user status pai-voice-server` shows active after install
- [ ] Linux: Service survives reboot (with `loginctl enable-linger`)
- [ ] Linux: `./stop.sh`, `./start.sh`, `./restart.sh`, `./status.sh`, `./uninstall.sh` all work correctly
- [ ] macOS: All existing LaunchAgent behavior unchanged
- [ ] Both: `curl http://localhost:8888/health` returns healthy response after install

**Tested on:** Ubuntu 24.04 LTS

### DIFF
--- a/VoiceServer/install.sh
+++ b/VoiceServer/install.sh
@@ -1,0 +1,279 @@
+#!/bin/bash
+
+# Voice Server Installation Script
+# Supports macOS (LaunchAgent) and Linux (systemd user service)
+
+set -e
+
+# Colors for output
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+BLUE='\033[0;34m'
+NC='\033[0m' # No Color
+
+# Configuration
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+ENV_FILE="$HOME/.env"
+OS=$(uname -s)
+
+# OS-specific paths
+if [[ "$OS" == "Darwin" ]]; then
+    SERVICE_NAME="com.pai.voice-server"
+    PLIST_PATH="$HOME/Library/LaunchAgents/${SERVICE_NAME}.plist"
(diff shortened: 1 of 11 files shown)


espcoder:
Closing to revisit the contribution approach — VoiceServer doesn't exist in upstream so every attempt shows all files as new additions. Need to align with maintainer on the right path before opening again.

PR #846, opened 2026-03-01
Fix missing user context migration for v4.x upgrades

catchingknives:
## Summary

- **v4.0 moved user context** from `skills/PAI/USER/` (v2.5–v3.0) and `skills/CORE/USER/` (v2.4) to a new canonical location at `PAI/USER/`, with `CONTEXT_ROUTING.md` pointing there
- **The installer never migrates existing user files** during upgrades — `PAI/USER/` is created empty while user data stays stranded at the old `skills/` path
- **Result:** 27 of 47 context-routed paths are broken after upgrading. Sessions that search for files write to whichever copy they find first, fragmenting user data across two directories

## Fix

Adds a migration step to `runRepository()` that runs on upgrades:

1. Recursively copies user files from `skills/PAI/USER/` (or `skills/CORE/USER/`) to `PAI/USER/`, skipping files that already exist at the destination
2. Replaces the legacy directory with a symlink (`skills/PAI/USER → ../../PAI/USER`) so the PAI skill's relative `USER/` paths continue to resolve correctly
3. Only runs when `installType === "upgrade"` — fresh installs are unaffected

## Affected versions

Users upgrading from v2.4, v2.5, or v3.0 to v4.0.0 or v4.0.1.

## Test plan

- [ ] Fresh install: verify `PAI/USER/` is created with directories + README, migration is skipped
- [ ] Upgrade from v3.0: verify files are copied from `skills/PAI/USER/` to `PAI/USER/`, legacy dir replaced with symlink
- [ ] Upgrade from v2.4: verify files are copied from `skills/CORE/USER/` to `PAI/USER/`
- [ ] Re-run installer after migration: verify symlink is detected and migration is skipped (idempotent)
- [ ] Verify context routing resolves correctly after migration (e.g., `PAI/USER/TELOS/BELIEFS.md` exists and contains user data)

### DIFF
--- a/Releases/v4.0.1/.claude/PAI-Install/engine/actions.ts
+++ b/Releases/v4.0.1/.claude/PAI-Install/engine/actions.ts
@@ -5,7 +5,7 @@
  */
 
 import { execSync, spawn } from "child_process";
-import { existsSync, mkdirSync, writeFileSync, readFileSync, readdirSync, symlinkSync, unlinkSync, chmodSync, lstatSync } from "fs";
+import { existsSync, mkdirSync, writeFileSync, readFileSync, readdirSync, symlinkSync, unlinkSync, chmodSync, lstatSync, cpSync, rmSync } from "fs";
 import { homedir } from "os";
 import { join, basename } from "path";
 import type { InstallState, EngineEventHandler, DetectionResult } from "./types";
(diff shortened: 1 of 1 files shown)


danielmiessler:
Thank you @catchingknives! The user context migration is merged into PAI v4.0.3. Excellent work — the `copyMissing` helper with symlink replacement is exactly the right approach for preserving backwards compatibility while migrating to the v4.x canonical paths.

**Fix included in:** [Releases/v4.0.3](https://github.com/danielmiessler/Personal_AI_Infrastructure/tree/main/Releases/v4.0.3)

PR #764, opened 2026-02-22
fix(telos): remove reference to missing update-telos.ts script

nbost130:
## Summary

- The Telos Update workflow (`Releases/v3.0/.claude/skills/Telos/Workflows/Update.md`) references `~/.claude/commands/update-telos.ts`, but that path does not exist
- The actual implementation lives at `~/.claude/skills/Telos/Tools/UpdateTelos.ts` within the same skill directory
- Updated all three stale path references (command invocation, example, and documentation footer) to point to the correct `Tools/UpdateTelos.ts` location

Closes #752

## Test plan

- [ ] Verify the Update workflow's `bun` command invocation points to the correct `Tools/UpdateTelos.ts` path
- [ ] Confirm no remaining references to the non-existent `~/.claude/commands/update-telos.ts`
- [ ] Validate the script at `Tools/UpdateTelos.ts` matches what the workflow expects (file validation, backup, changelog)

PR #852, opened 2026-03-02
fix(telos): correct stale path to UpdateTelos.ts in Update workflow

nbost130:
## Summary

- The Telos Update workflow (`Workflows/Update.md`) references `~/.claude/commands/update-telos.ts`, but that path does not exist
- The actual implementation lives at `~/.claude/skills/Telos/Tools/UpdateTelos.ts`
- Updated all three stale path references (command invocation, example, and documentation footer) in v4.0.2 and v4.0.3

## Context

This is a follow-up to #764 which identified the same bug. That PR was closed as "addressed in v4.0.x" but the broken references persist identically in v4.0.0 through v4.0.3.

Verified by pulling the upstream `Update.md` from each release directory — all still reference the non-existent `~/.claude/commands/update-telos.ts` path.

## Test plan

- [x] `grep -rn 'commands/update-telos' Releases/v4.0.{2,3}/` returns zero matches after fix
- [x] All three references (line 81 command, line 143 example, line 289 docs) updated in both v4.0.2 and v4.0.3
- [ ] `bun ~/.claude/skills/Telos/Tools/UpdateTelos.ts` executes successfully with valid args

### DIFF
--- a/Releases/v4.0.2/.claude/skills/Telos/Workflows/Update.md
+++ b/Releases/v4.0.2/.claude/skills/Telos/Workflows/Update.md
@@ -78,7 +78,7 @@ This is the main command you'll use. It takes three parameters:
 - Content to add (the actual text)
 - Description of the change (for the changelog)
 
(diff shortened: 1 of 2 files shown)


ksaynice:
bump this

danielmiessler:
Hey @nbost130, thank you for the PR and the work that went into it.

PR #889, opened 2026-03-03
feat(pai): add CLAUDE_MODEL env variable support

### LINKED ISSUE #494: Add possibility to set desired model from env
Add something like this to `pai.ts` -> `cmdLaunch`
```
 // Read model from environment variable
  const model = process.env.CLAUDE_MODEL;
  if (model) {
    args.push("--model", model);
  }
```

So I can have alias like this `alias pai="CLAUDE_MODEL=gpt-oss:20b bun ~/.claude/skills/CORE/Tools/pai.ts"`.
In this case I'm using local models from Ollama.

NJX-njx:
Allow users to set model via environment variable, e.g.:
  CLAUDE_MODEL=gpt-oss:20b bun ~/.claude/skills/CORE/Tools/pai.ts

Useful for local models (Ollama) and aliases.

Fixes #494

Made-with: Cursor

### DIFF
--- a/Releases/v4.0.0/.claude/PAI/Tools/pai.ts
+++ b/Releases/v4.0.0/.claude/PAI/Tools/pai.ts
@@ -412,6 +412,12 @@ async function cmdLaunch(options: { mcp?: string; resume?: boolean; skipPerms?: 
     args.push("--resume");
   }
 
+  // Read model from environment variable (e.g. CLAUDE_MODEL=gpt-oss:20b for Ollama)
+  const model = process.env.CLAUDE_MODEL;
+  if (model) {
+    args.push("--model", model);
+  }
+
   // Change to PAI directory unless --local flag is set
   if (!options.local) {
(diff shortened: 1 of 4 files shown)


danielmiessler:
Hey @NJX-njx — Kai here, Daniel's AI assistant, doing the next pass of triage on the PAI repo.

Thanks for taking the time to put this together — really appreciated, and the underlying intent (`CLAUDE_MODEL` env variable support) is reasonable. Closing it out in our pre-v5.0.0 triage pass because the specific design diverges from how v5.0.0 organizes around this concern, and merging it as-is would create a parallel pattern that competes with the v5 layer. Better to evaluate the spirit of it against the new architecture once v5 ships.

**If after spending time with v5.0.0 you still feel this would add value — either as-is or reframed against the new architecture — please reopen or open a fresh PR against the new release.** Your input is valuable.

— Kai 🤖

PR #872, opened 2026-03-02
feat(voice-server + installer): Google Cloud TTS + cross-platform audio

fayerman-source:
## Summary

- **Google Cloud TTS** as alternative provider alongside ElevenLabs — configurable via `settings.json → daidentity.ttsProvider`
- **Cross-platform audio playback** — detects `afplay` (macOS), `mpv`/`ffplay`/`aplay` (Linux/WSL2) at startup
- **Cross-platform desktop notifications** — `osascript` (macOS), `notify-send` (Linux) with silent fallback
- **Backwards compatible** — defaults to ElevenLabs when `ttsProvider` is not set

## Why

- ElevenLabs free tier is 10K chars/month; Google Cloud free tier is **4M chars/month** (Standard) or 1M (WaveNet/Neural2)
- Voice server used macOS-only `afplay` and `osascript`, making it non-functional on Linux/WSL2
- No new dependencies — uses Google's REST API directly via `fetch`

## Configuration

Add to `~/.env`:
```
GOOGLE_CLOUD_API_KEY=your_key_here
# or reuse existing: GOOGLE_API_KEY=your_key_here (both accepted)
```

Add to `~/.claude/settings.json`:
```json
{
  "daidentity": {
    "ttsProvider": "google-cloud",
    "googleCloudVoice": {
      "languageCode": "en-US",
      "voiceName": "en-US-Neural2-D",
      "voiceType": "NEURAL2",
      "speakingRate": 1.0,
      "pitch": 0.0
    }
  }
}
```

Or keep using ElevenLabs by not setting `ttsProvider` (or setting it to `"elevenlabs"`).

## Test plan

- [x] Voice server starts with `ttsProvider: "google-cloud"` and logs correct provider
- [x] Health endpoint shows `voice_system: "google-cloud"`, `google_cloud_configured: true`, `audio_player`
- [x] TTS generates and plays audio end-to-end on Linux/WSL2 via mpv
- [x] Backwards compatible — omitting `ttsProvider` defaults to ElevenLabs
- [ ] Verify on macOS with `afplay`

## Context

Re-implementation of #687 (closed during v4.0 restructuring) with additional Linux/WSL2 cross-platform support. Tested live on WSL2 with Google Cloud Neural2-D voice.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #855, opened 2026-03-02
VoiceServer: afplay (macOS-only) breaks audio on Linux — silent failure


pkumaschow:
## Summary

`VoiceServer/server.ts` hardcodes `/usr/bin/afplay` for audio playback. `afplay` is macOS-only and does not exist on Linux, causing all TTS notifications to fail silently with `ENOENT`. Because the startup catchphrase is fired as a detached fire-and-forget `curl`, the error is never surfaced to the user — the voice server appears to work but produces no audio.

## Affected Files

- `VoiceServer/server.ts` — `playAudio()` function (~line 379)
- `VoiceServer/start.sh`, `stop.sh`, `restart.sh` — use `launchctl` (macOS-only), fail silently on Linux

## Error

```
{"status":"error","message":"TTS failed: ENOENT: no such file or directory, posix_spawn '/usr/bin/afplay'"}
```

## Platform

- **Linux** (tested on Fedora 43, kernel 6.18)
- macOS unaffected

## What Breaks

- Startup catchphrase never plays
- Algorithm phase voice announcements never play
- All `/notify` POST calls return `status: error`
- No visible error to the user (fire-and-forget curl swallows it)

## Regression Vector

Every PAI migration/reinstall that overwrites `VoiceServer/server.ts` from the upstream repo reverts any local Linux fix, breaking voice again silently.

## Proposed Fix

In `playAudio()`, detect the platform and use `ffplay` (part of ffmpeg, widely available on Linux) instead of `afplay`:

```typescript
// Play audio — afplay on macOS, ffplay on Linux
async function playAudio(audioBuffer: ArrayBuffer, volume: number = FALLBACK_VOLUME): Promise<void> {
  const tempFile = `/tmp/voice-${Date.now()}.mp3`;
  await Bun.write(tempFile, audioBuffer);

  const isMac = process.platform === 'darwin';
  const player = isMac ? '/usr/bin/afplay' : '/usr/bin/ffplay';
  const args = isMac
    ? ['-v', volume.toString(), tempFile]
    : ['-nodisp', '-autoexit', '-volume', Math.round(volume * 100).toString(), tempFile];

  return new Promise((resolve, reject) => {
    const proc = spawn(player, args);
    proc.on('error', (error) => { reject(error); });
    proc.on('exit', (code) => {
      spawn('/bin/rm', [tempFile]);
      if (code === 0) resolve();
[...]

PR #897, opened 2026-03-03
feat(voice): fall back to macOS say when ElevenLabs unavailable

blu3dot:
## Summary

When `ELEVENLABS_API_KEY` is missing or the ElevenLabs API call fails, VoiceServer now falls back to the macOS native `say` command instead of silently skipping voice output. This gives new users immediate voice feedback without requiring ElevenLabs setup.

## Problem

VoiceServer silently skips all TTS when `ELEVENLABS_API_KEY` is missing. New users who haven't configured ElevenLabs get zero voice feedback with no indication why — the server returns HTTP 200, making the failure invisible.

## Solution

After the ElevenLabs TTS path is skipped (no API key) or fails (API error), fall back to `/usr/bin/say`:

- Reuses existing `applyPronunciations()` so custom pronunciation rules apply to native voice too
- Reuses existing `spawnSafe()` helper (array-form spawn, no shell injection)
- Only triggers when ElevenLabs path didn't play (no double-speak)
- Fails gracefully — logs error, doesn't crash server
- Uses `error: unknown` with `instanceof` type guard (also upgrades the adjacent ElevenLabs catch block)

**macOS-only** — `say` is present on all macOS versions. Linux equivalent is out of scope for this PR (see #855, #872 for cross-platform audio work).

## Test plan

- [ ] Start server with no `ELEVENLABS_API_KEY` → send notification → hear macOS `say` voice
- [ ] Set invalid `ELEVENLABS_API_KEY` → send notification → ElevenLabs fails → fallback to `say`
- [ ] Both paths fail (e.g., Linux) → notification still displays, no voice, error logged
- [ ] Custom pronunciations from `pronunciations.json` apply to `say` output

## Post-Deploy Monitoring & Validation

No additional operational monitoring required: additive fallback path, existing behavior unchanged when ElevenLabs key is configured.

---

[Compound Engineered](https://github.com/EveryInc/compound-engineering-plugin) 🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Releases/v4.0.3/.claude/VoiceServer/server.ts
+++ b/Releases/v4.0.3/.claude/VoiceServer/server.ts
@@ -506,9 +506,28 @@ async function sendNotification(
       const audioBuffer = await generateSpeech(safeMessage, voice, resolvedSettings);
       await playAudio(audioBuffer, resolvedVolume);
       voicePlayed = true;
-    } catch (error: any) {
+    } catch (error: unknown) {
+      const msg = error instanceof Error ? error.message : "TTS generation failed";
       console.error("Failed to generate/play speech:", error);
-      voiceError = error.message || "TTS generation failed";
+      voiceError = msg;
+    }
+  }
+
+  // Fallback: macOS native voice when ElevenLabs is unavailable or failed
+  // Uses /usr/bin/say (available on all macOS versions). Pronunciation rules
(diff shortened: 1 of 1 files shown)


blu3dot:
Closing to test locally before resubmitting.

Issue #905, opened 2026-03-04
MineRatings.ts: ARG_MAX risk when passing large prompts as CLI args to Inference.ts


virtualian:
## Summary

`MineRatings.ts` shells out to `bun Inference.ts` passing both the system prompt and user prompt as **CLI positional arguments**. With 750+ rating entries, the JSON summary in the user prompt can approach OS `ARG_MAX` limits (256KB per arg on macOS), causing silent failures — the process exits with zero output and no error.

## Reproduction

```bash
bun ~/.claude/skills/Utilities/PAIUpgrade/Tools/MineRatings.ts --all
```

With a sufficiently large `ratings.jsonl` (750+ entries), the spawned `bun Inference.ts <system_prompt> <user_prompt>` call can silently fail due to argument length.

## Root Cause

`MineRatings.ts` `runInference()` passes prompts as argv:

```typescript
const proc = spawn('bun', [
  INFERENCE_PATH,
  '--level', 'standard',
  '--timeout', '300000',
  systemPrompt,  // ~350 words
  userPrompt,    // JSON summary of all entries — grows with dataset
], { ... });
```

Meanwhile, `Inference.ts` itself already handles ARG_MAX correctly internally — it pipes the user prompt via **stdin** to `claude --print` (line 96-98). The problem is in MineRatings' invocation of Inference.ts, not in Inference.ts itself.

## Fix

Replace the subprocess spawn with a direct import of `inference()`:

```typescript
import { inference } from '../../../../PAI/Tools/Inference.ts';

async function runInference(systemPrompt, userPrompt) {
  const result = await inference({
    systemPrompt,
    userPrompt,
    level: 'standard',
    timeout: 300000,
  });
  return { success: result.success, output: result.success ? result.output : (result.error || 'Inference failed') };
}
```

This eliminates the subprocess entirely and reuses the stdin-piping that `inference()` already does.

## Priority

Low — only manifests with large datasets (750+ entries) and is already fixed in virtualian/pai#49.

danielmiessler:
Hey @virtualian — Kai here, Daniel's AI assistant, doing first-pass triage on the PAI repo.

Closing this out because `MineRatings.ts` no longer exists in v5.0.0 — the rating-mining path has been redesigned and the ARG_MAX surface you identified is gone (verified: zero `MineRatings*` files in v5).

[...]

PR #911, opened 2026-03-05
fix(Telos): replace subprocess spawn with direct inference() import

rikitikitavi2012-debug:
## Summary
- Replaces `child_process.spawn('bun', ['Inference.ts', ...args])` with direct `import { inference }` call in Telos DashboardTemplate chat route
- Eliminates ARG_MAX risk when prompts include large TELOS context data
- Same pattern as issue #905 — subprocess passes prompts as CLI args which can exceed OS limits

## Root cause
`route.ts` spawned a subprocess passing system prompt and user message as positional CLI arguments. With rich TELOS context, these can approach OS `ARG_MAX` limits (128-256KB), causing silent failures.

## Fix
Direct import of `inference()` function which handles prompts internally via stdin, bypassing any argument length limits.

Related to #905

## Test plan
- [x] `bash -n` syntax check on modified file
- [x] Import path resolves correctly via dynamic import
- [x] Minimal diff — only inference invocation changed

### DIFF
--- a/Releases/v4.0.3/.claude/skills/Telos/DashboardTemplate/App/api/chat/route.ts
+++ b/Releases/v4.0.3/.claude/skills/Telos/DashboardTemplate/App/api/chat/route.ts
@@ -1,6 +1,16 @@
 import { NextResponse } from "next/server"
 import { getTelosContext } from "@/lib/telos-data"
-import { spawn } from "child_process"
+import { join } from "path"
+import { homedir } from "os"
+
+// Dynamic import resolved at runtime — avoids ARG_MAX risk from passing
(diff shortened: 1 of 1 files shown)


danielmiessler:
Hey @rikitikitavi2012-debug — Kai here, Daniel's AI assistant, doing first-pass triage on the PAI repo.

Closing this out because this PR targets v4.0.x code paths that v5.0.0 reorganizes around — the v4.0.3 Telos subprocess path you patched has been replaced in v5's `skills/Telos/Tools/UpdateTelos.ts`. Rather than carry it forward, it'll be cleaner to evaluate against the v5.0.0 layout if the underlying concern still applies.

— Kai 🤖

PR #928, opened 2026-03-07
Add Cloudflare Markdown for Agents pre-check to fetch workflows

catchingknives:
## Idea: Free markdown before the tier chain even starts

Cloudflare launched [Markdown for Agents](https://blog.cloudflare.com/markdown-for-agents/) in February 2026 — sites behind Cloudflare (Pro+) can now serve native markdown via `Accept: text/markdown` content negotiation. Their benchmark shows ~80% token savings (16,180 → 3,150 tokens on a typical blog post). Server-side conversion is more accurate than client-side HTML parsing. Non-Cloudflare sites simply ignore the header and return HTML, so there's zero downside to trying.

This PR adds a lightweight curl pre-check to **FourTierScrape** and **Retrieve** that probes for markdown before entering the existing tier/layer chain:

```bash
curl -sL -H "Accept: text/markdown" "[URL]" | head -5
```

If the body comes back as markdown (YAML frontmatter or `#` heading instead of `<!DOCTYPE`) → use it directly, skip the tier chain. If HTML → proceed to Tier 1 as normal.

### Why "pre-check" and not "Tier 0"

Deliberately framed as a preamble rather than a new numbered tier so the four-tier naming/numbering stays intact. No rename, no renumbering, just an additive optimization before the chain starts.

### What changed

| File | Change |
|------|--------|
| `Scraping/BrightData/Workflows/FourTierScrape.md` | Added pre-check section before Tier 1, updated decision diagram, updated Tier 2 Accept header to prefer `text/markdown` |
| `Research/Workflows/Retrieve.md` | Added pre-fetch note in Layer 1 section |

### Gotcha discovered during testing

Cloudflare's CDN currently returns `content-type: text/html` in the response header even when the body is markdown (verified on `blog.cloudflare.com`). The `vary: accept` header confirms content negotiation occurred. Detection should check the body format, not just the content-type header. This is noted in the workflow.

### Tested against

- `blog.cloudflare.com/markdown-for-agents/` with `Accept: text/markdown` → body is clean markdown with YAML frontmatter ✓
- Same URL with `Accept: text/html` → body is raw HTML ✓
- `example.com` with `Accept: text/markdown` → body is HTML, graceful fallthrough ✓

Open to feedback on framing, placement, or whether this should live somewhere else entirely.

---

### DIFF
--- a/Releases/v4.0.3/.claude/skills/Research/Workflows/Retrieve.md
+++ b/Releases/v4.0.3/.claude/skills/Research/Workflows/Retrieve.md
@@ -80,6 +80,22 @@ Layer 3: Apify MCP (RAG browser, Actor ecosystem)
 
 ## Layer 1: Built-in Tools
 
+### Pre-fetch: Markdown Content Negotiation (Cloudflare)
+
+Before calling WebFetch, attempt a lightweight curl probe requesting native markdown via Cloudflare's [Markdown for Agents](https://blog.cloudflare.com/markdown-for-agents/) feature. Sites behind Cloudflare (Pro+) that have enabled this feature serve server-side converted markdown via `Accept: text/markdown`. Non-Cloudflare sites simply ignore the header -- zero downside.
+
+```bash
+curl -sL -H "Accept: text/markdown" "[URL]" | head -5
+```
+
+- If body starts with YAML frontmatter (`---`) or markdown heading (`# `) instead of `<!DOCTYPE`/`<html` → use body directly (skip WebFetch)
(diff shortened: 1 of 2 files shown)


danielmiessler:
Thanks @catchingknives, the Cloudflare markdown content-negotiation pre-check is in the current source (both the Research Retrieve and BrightData FourTierScrape workflows) and ships in the latest release. The repo is a generated artifact, so we port and close rather than merge. Please pull the latest. This one made the cut, thank you!

PR #933, opened 2026-03-09
fix: SecurityValidator confirm handler silently passes through

SaintPepsi:
## Summary

Two security bugs in SecurityValidator.hook.ts (v3.0) that cause the hook to be effectively non-functional:

- **Confirm handler uses wrong JSON structure**: The confirm category outputs `{"decision": "ask", "message": "..."}` to stdout, but Claude Code PreToolUse hooks require a specific JSON structure with `hookSpecificOutput.permissionDecision`. The `{"decision": "ask"}` format is not recognized and silently ignored, meaning operations matching confirm patterns execute without any user interaction.
- **Pattern file paths have incorrect prefix**: `paiPath('skills', 'PAI', 'USER', ...)` resolves to `~/.claude/skills/PAI/USER/...`, but `patterns.yaml` lives at `~/.claude/PAI/USER/...` (no `skills/` prefix). This means the patterns file is never found, SecurityValidator falls through to `EMPTY_PATTERNS` (fail-open, everything allowed).

## Proof: Claude Code only recognizes specific PreToolUse output formats

From the [official Claude Code hooks documentation](https://code.claude.com/docs/en/hooks-guide):

**Exit codes:**
> - **Exit 0**: the action proceeds.
> - **Exit 2**: the action is blocked. Write a reason to stderr, and Claude receives it as feedback.

**Structured JSON output (exit 0 + stdout):**
> For example, a PreToolUse hook can deny a tool call and tell Claude why, or escalate it to the user for approval:
> ```json
> {
>   "hookSpecificOutput": {
>     "hookEventName": "PreToolUse",
>     "permissionDecision": "deny",
>     "permissionDecisionReason": "Use rg instead of grep for better performance"
>   }
> }
> ```
> These three options are specific to PreToolUse:
> - `"allow"`: proceed without showing a permission prompt
> - `"deny"`: cancel the tool call and send the reason to Claude
> - `"ask"`: show the permission prompt to the user as normal

The upstream SecurityValidator outputs `{"decision": "ask", "message": "..."}` which does NOT match either recognized format. It's not exit code 2, and it's not the `hookSpecificOutput` structure. Claude Code silently ignores it and proceeds with the operation.

## Changes

1. Changed all 3 confirm handlers (Bash, Edit, Write) from outputting `{"decision": "ask"}` to using `process.exit(2)` with descriptive stderr messages — matching the hard-block pattern used for `blocked` category
2. Removed `'skills'` from both `paiPath()` calls so `patterns.yaml` is actually found at `~/.claude/PAI/USER/PAISECURITYSYSTEM/patterns.yaml`
3. Updated JSDoc header to document the bug fix

## Impact

Without this fix, SecurityValidator's confirm-level protections provide zero actual protection — they silently pass through. The path bug compounds this by preventing ALL pattern matching from working, meaning even blocked-level patterns don't fire.

## Alternative fix

Instead of `process.exit(2)`, the confirm handlers could use the correct JSON structure to actually prompt the user:
```json
{
  "hookSpecificOutput": {
    "hookEventName": "PreToolUse",
    "permissionDecision": "ask",
    "permissionDecisionReason": "This operation requires confirmation"
  }
}
```
This PR uses `exit(2)` (hard block) as the safer default, since confirm-level operations are inherently risky. But switching to `"permissionDecision": "ask"` would restore the original intent of prompting the user.

— 🍁 Maple

### DIFF
--- a/Releases/v3.0/.claude/hooks/SecurityValidator.hook.ts
+++ b/Releases/v3.0/.claude/hooks/SecurityValidator.hook.ts
@@ -17,13 +17,22 @@
  * OUTPUT:
  * - stdout: JSON decision object
  *   - {"continue": true} → Allow operation
- *   - {"decision": "ask", "message": "..."} → Prompt user for confirmation
+ *   (Note: confirm-level operations now hard-block via exit(2), see below)
  * - exit(0): Normal completion (with decision)
  * - exit(2): Hard block (catastrophic operation prevented)
  *
  * SIDE EFFECTS:
  * - Writes to: MEMORY/SECURITY/YYYY/MM/security-{summary}-{timestamp}.jsonl
- * - User prompt: May trigger confirmation dialog for confirm-level operations
+ * - Confirm operations: Hard-blocked via exit(2) — see BUG FIX note below
+ *
+ * BUG FIX (2026-03-09):
+ * Claude Code PreToolUse hooks only recognize two stdout formats:
+ *   - {"continue": true} to allow the operation
+ *   - exit code 2 to hard block the operation
+ * The previously used {"decision": "ask"} format is NOT recognized by
+ * Claude Code and silently passes through, allowing operations that
+ * should require confirmation to execute without any user prompt.
+ * Confirm-level operations now use exit(2) to hard-block instead.
  *
  * INTER-HOOK RELATIONSHIPS:
  * - DEPENDS ON: patterns.yaml (security pattern definitions)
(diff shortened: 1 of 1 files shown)


danielmiessler:
Hey @SaintPepsi — Kai here, Daniel's AI assistant. (Apologies — prior comment had shell-escaped backticks; reposting.)

Closing this out because v5.0.0 removed `SecurityValidator.hook.ts` and rebuilt security as `hooks/SecurityPipeline.hook.ts` — the confirm-passthrough path you identified no longer exists (verified).

**If this still reproduces or matters in v5.0.0, please reopen or file a fresh issue against the new release.** Thanks for taking the time to file this — really appreciated.

— Kai 🤖

PR #944, opened 2026-03-12
feat: add uninstall.sh to fully remove PAI from a system

gabehamilton:
## Summary

- Adds `uninstall.sh` at the repo root and mirrored to `Releases/v3.0/.claude/uninstall.sh`
- Covers every artifact created by the v3.0 installer: `~/.claude/`, `~/.config/PAI/`, LaunchAgent plist, voice server process on port 8888, shell aliases (zsh + fish), symlinks (`~/.env`, `~/.claude/.env`), SwiftBar/BitBar plugins, and log files
- Prompts for confirmation before each destructive step; supports `--force` flag for scripted/non-interactive use
- `~/.env` is only removed if it's a PAI-owned symlink — won't clobber a pre-existing file
- Closing note instructs user to uninstall the Claude Code CLI separately (PAI doesn't own that binary)

## Test plan

- [ ] Run on a fresh PAI v3.0 install and verify all artifacts are removed
- [ ] Run with `--force` and verify it skips all prompts
- [ ] Verify `~/.env` is left alone when it is not a PAI symlink
- [ ] Verify shell alias is cleanly stripped from `~/.zshrc` without leaving blank lines
- [ ] Verify voice server LaunchAgent is unloaded before plist is deleted


danielmiessler:
Hey @gabehamilton, thanks for raising this, and sorry it sat for a while.

PR #951, opened 2026-03-13
fix: add audio device probe and afplay timeout to VoiceServer

MarvinDontPanic:
## Summary

Two related reliability fixes for VoiceServer on macOS:

- **Audio device probe at startup**: Plays a near-silent system sound (`Tink.aiff` at 1% volume) with a 3-second timeout. If `afplay` hangs (no output device, virtual audio, or running under `launchd` without an audio session), all voice playback is disabled for the session rather than hanging the server
- **afplay timeout per-clip**: 10-second timeout on each playback call with a settled-flag pattern to prevent resolve/reject races. Most TTS clips are 1-3s, so 10s catches hung audio devices without cutting off legitimate long utterances

## Problem

Without these fixes, `afplay` hangs indefinitely when no audio output device is available, blocking the notification server's `/notify` endpoint. This commonly happens when:
- Running VoiceServer as a system service (launchd) without an audio session
- Virtual audio devices that accept connections but don't produce output
- Headless environments or SSH sessions

## File Changed

`Releases/v4.0.3/.claude/VoiceServer/server.ts`

## Test plan

- [ ] Start VoiceServer with working audio → probe reports "output device available", playback works
- [ ] Start VoiceServer without audio output → probe reports "no usable output device", `/notify` returns success but skips playback
- [ ] Send a notification during audio device disconnect → timeout fires after 10s instead of hanging

### DIFF
--- a/Releases/v4.0.3/.claude/VoiceServer/server.ts
+++ b/Releases/v4.0.3/.claude/VoiceServer/server.ts
@@ -41,6 +41,47 @@ if (!ELEVENLABS_API_KEY) {
 }
 
 // ==========================================================================
+// Audio Device Probe — detect at startup whether afplay can actually play audio.
+// If the default output device is unavailable (virtual device, no speakers,
+// or running under launchd without an audio session), afplay hangs indefinitely.
+// We detect this once and skip all voice playback when audio is unavailable.
+// ==========================================================================
+
+const AUDIO_PROBE_TIMEOUT_MS = 3_000;
+let audioAvailable = false;
+
+async function probeAudioDevice(): Promise<boolean> {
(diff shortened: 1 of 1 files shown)


danielmiessler:
Hey @MarvinDontPanic — Kai here, Daniel's AI assistant, doing first-pass triage on the PAI repo.

Closing this out because this PR targets v4.0.x code paths that v5.0.0 reorganizes around — the v4.0.3 VoiceServer this patches has moved to `PAI/PULSE/VoiceServer/voice.ts` in v5. Rather than carry it forward, it'll be cleaner to evaluate against the v5.0.0 layout if the underlying concern still applies.

— Kai 🤖

PR #968, opened 2026-03-17
docs: remove references to unimplemented AlgorithmTab.hook.ts

HyggeHacker:
## Summary
- Removes `AlgorithmTab.hook.ts` references from `hooks/README.md` and `PAI/THEHOOKSYSTEM.md`
- This hook is documented but was never implemented — its functionality (Kitty tab phase updates) is already handled by `PRDSync.hook.ts`, which imports `AlgorithmTabPhase` and calls `setPhaseTab()` on every phase transition
- No functional changes — documentation alignment only

## Changes
- **hooks/README.md**: Removed AlgorithmTab row from Stop Hooks table, execution model diagram, and updated hook count (22→21)
- **PAI/THEHOOKSYSTEM.md**: Removed AlgorithmTab settings.json entry, standalone hook description, quick reference line, and updated hook counts


danielmiessler:
Hey @HyggeHacker, thank you for the PR and the work that went into it.

PR #986, opened 2026-03-23
feat(pai.ts): passthrough unknown CLI flags to claude

HyggeHacker:
## Summary

PAI's arg parser silently dropped unrecognized flags (e.g., `--plugin-dir`, `--mcp-config`), preventing users from combining PAI with Claude Code's native flags. Unknown flags are now collected and forwarded to the underlying `claude` command, enabling use cases like:

```bash
pai -l --plugin-dir ~/Projects/boxie --mcp-config ~/.claude/mcp-profiles/boxie.json
```

## Changes

- Unknown flags (args starting with `-`) are collected into a `passthrough` array instead of being silently ignored
- If the next arg after an unknown flag doesn't start with `-`, it's treated as that flag's value
- Non-flag unknown args still produce the existing error message
- The passthrough array is forwarded to `cmdLaunch()` and spread into the `claude` command args

## Known limitation

Flags that accept **multiple space-separated values** (e.g., `--add-dir dir1 dir2`, `--allowedTools Bash Edit Read`) will only capture the first value. This covers the vast majority of Claude Code flags which take a single value. Multi-value flags can use repeated invocations (`--add-dir dir1 --add-dir dir2`) or `=` syntax as a workaround.

## Test plan

- [x] `pai -l --plugin-dir ~/Projects/boxie` — plugin-dir passed through correctly
- [x] `pai --mcp-config ~/.claude/mcp-profiles/boxie.json` — mcp-config passed through
- [x] `pai unknowncmd` — still errors with "Unknown command"
- [x] `pai --model opus` — passed through correctly
- [x] Existing flags (`-r`, `-m`, `-l`, `--safe`) still work as before

### DIFF
--- a/Releases/v4.0.3/.claude/PAI/Tools/pai.ts
+++ b/Releases/v4.0.3/.claude/PAI/Tools/pai.ts
@@ -390,7 +390,7 @@ function cmdWallpaper(args: string[]) {
 // Commands
 // ============================================================================
 
-async function cmdLaunch(options: { mcp?: string; resume?: boolean; skipPerms?: boolean; local?: boolean }) {
+async function cmdLaunch(options: { mcp?: string; resume?: boolean; skipPerms?: boolean; local?: boolean; passthrough?: string[] }) {
   // CLAUDE.md is now static — no build step needed.
   // Algorithm spec is loaded on-demand when Algorithm mode triggers.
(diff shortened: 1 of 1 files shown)


danielmiessler:
Hey @HyggeHacker, thank you for the PR and the work that went into it.

PR #991, opened 2026-03-24
Update PAI skill path in validate.ts to the correct path

tcmoran00:
Simple one-line fix so the validator checks the correct path for the PAI core skill. 

The validation step reports "PAI core skill: Not found - clone PAI repo to enable" even though the PAI core skill exists. The validator looks for ~/.claude/skills/PAI/SKILL.md, but the actual file lives at ~/.claude/PAI/SKILL.md.

### DIFF
--- a/Releases/v4.0.3/.claude/PAI-Install/engine/validate.ts
+++ b/Releases/v4.0.3/.claude/PAI-Install/engine/validate.ts
@@ -107,7 +107,7 @@ export async function runValidation(state: InstallState): Promise<ValidationChec
   }
 
   // 4. PAI skill present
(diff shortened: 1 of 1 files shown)


danielmiessler:
Hey @tcmoran00 — Kai here, Daniel's AI assistant, doing first-pass triage on the PAI repo.

Closing this out because this PR targets v4.0.x code paths that v5.0.0 reorganizes around — the v4.x PAI skill path it corrects no longer reflects the v5.0.0 skill layout. Rather than carry it forward, it'll be cleaner to evaluate against the v5.0.0 layout if the underlying concern still applies.

— Kai 🤖

PR #994, opened 2026-03-25
fix: Respect PAI_DIR environment variable in v4.0.3 installer

sbusc:
## Summary

The v4.0.3 installer hardcodes `~/.claude` as the PAI directory in several fallback paths, ignoring the `PAI_DIR` environment variable. Users who set `PAI_DIR` to a custom path get a broken installation because detection, configuration, validation, and voice setup all fall back to `~/.claude` instead of respecting the env var.

## Changes

4 files in `Releases/v4.0.3/.claude/PAI-Install/`:

- **`engine/detect.ts`** — Read `PAI_DIR` from `process.env` when detecting system paths
- **`engine/actions.ts`** — Add `process.env.PAI_DIR` fallback in repository, configuration, and voice setup steps (3 locations)
- **`engine/validate.ts`** — Add `process.env.PAI_DIR` fallback in validation checks
- **`install.sh`** — Pass through `PAI_DIR` if set; fix unbound variable errors for `DISPLAY`/`WAYLAND_DISPLAY` in headless detection (`set -euo pipefail` compatibility)

## Test plan

- [ ] Run `bash install.sh` without `PAI_DIR` set — should default to `~/.claude` (no behavior change)
- [ ] Run `PAI_DIR=~/custom/.claude bash install.sh` — should install to `~/custom/.claude`
- [ ] Run installer on headless Linux without `DISPLAY` set — should not error on unbound variable

### DIFF
--- a/Releases/v4.0.3/.claude/PAI-Install/engine/actions.ts
+++ b/Releases/v4.0.3/.claude/PAI-Install/engine/actions.ts
@@ -500,7 +500,7 @@ export async function runRepository(
   emit: EngineEventHandler
 ): Promise<void> {
   await emit({ event: "step_start", step: "repository" });
-  const paiDir = state.detection?.paiDir || join(homedir(), ".claude");
+  const paiDir = state.detection?.paiDir || process.env.PAI_DIR || join(homedir(), ".claude");
 
   if (state.installType === "upgrade") {
     await emit({ event: "progress", step: "repository", percent: 20, detail: "Existing installation found, updating..." });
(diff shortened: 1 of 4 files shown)


danielmiessler:
Hey @sbusc — Kai here, Daniel's AI assistant, doing first-pass triage on the PAI repo.

Closing this out because this PR targets v4.0.x code paths that v5.0.0 reorganizes around — the v4.0.3 installer this targets has been replaced by the v5.0.0 one-line installer (PAI_DIR / CLAUDE_CONFIG_DIR semantics also reworked). Rather than carry it forward, it'll be cleaner to evaluate against the v5.0.0 layout if the underlying concern still applies.

— Kai 🤖

PR #1000, opened 2026-03-26
fix: add MEMORY additionalDirectories and write permission to v4.0.3 settings

brycemagera:
## Problem

A regression introduced in Claude Code versions after 2.1.77 broke the MEMORY directory context injection. Users running newer Claude Code versions find that `~/.claude/MEMORY` is no longer accessible across projects, causing PAI's dynamic context (learning signals, session history, Wisdom Frames, etc.) to fail silently.

Reported in discussion #992 — reverting to Claude Code 2.1.77 resolves it, but that's not a sustainable fix.

## Solution

Two surgical additions to `Releases/v4.0.3/.claude/settings.json`:

1. **`additionalDirectories`** — registers `~/.claude/MEMORY` as an additional working directory so Claude Code exposes it across all project sessions
2. **`Write(~/.claude/MEMORY/**)`** in the `allow` permissions — lets PAI hooks write to MEMORY without prompting (required for WorkCompletionLearning, IntegrityCheck, and other session-end hooks)

Tilde paths (`~`) confirmed working — consistent with existing permission patterns already in this file (e.g. `Write(~/.claude/settings.json)`).

## Files Changed

- `Releases/v4.0.3/.claude/settings.json` — 6-line addition

## Test Plan

- [ ] Fresh install of PAI v4.0.3 on Claude Code > 2.1.77
- [ ] Start a new session — verify PAI Dynamic Context block loads (learning signals, recent sessions visible)
- [ ] Confirm hooks can write to MEMORY without permission prompts

### DIFF
--- a/Releases/v4.0.3/.claude/settings.json
+++ b/Releases/v4.0.3/.claude/settings.json
@@ -26,7 +26,8 @@
       "ExitPlanMode",
       "Task",
       "Skill",
-      "mcp__*"
+      "mcp__*",
+      "Write(~/.claude/MEMORY/**)"
     ],
     "deny": [],
     "ask": [
@@ -60,7 +61,10 @@
       "Write(~/.ssh/*)",
       "Edit(~/.ssh/*)"
     ],
-    "defaultMode": "default"
+    "defaultMode": "default",
+    "additionalDirectories": [
+      "~/.claude/MEMORY"
+    ]
   },
   "enableAllProjectMcpServers": true,
   "enabledMcpjsonServers": [],


virtualian:
## Tested — this fix doesn't work

We independently tested this exact approach (tilde-form `additionalDirectories` + scoped `Write(~/.claude/MEMORY/**)` in allow list) and it still prompts for permission. The hardcoded `.claude/` sensitive directory protection in Claude Code ignores both settings.

Full analysis and evidence at anthropics/claude-code#29013 (comment: https://github.com/anthropics/claude-code/issues/29013#issuecomment-4151048712) — covers four different approaches we tried, all of which fail:

1. `Write(~/.claude/**)` in allow list
2. `Write(/Users/.../. claude/**)` (absolute path) in allow list
3. `additionalDirectories` with absolute path
4. `additionalDirectories` with tilde path + scoped Write allow (this PR's approach)

This is a Claude Code bug, not a PAI config issue. No settings-level workaround exists until Anthropic fixes the hardcoded protection.

PR #1005, opened 2026-03-29
feat(tools): Add GrokApi.ts — direct xAI client with model discovery and live X search

christauff:
## Summary

- Adds `PAI/Tools/GrokApi.ts` — direct xAI API client giving GrokResearcher actual Grok access
- Adds `PAI/Tools/GrokApi.test.ts` — 7 unit tests including self-heal verification
- Updates `GrokResearcherContext.md` with correct model names and API usage

## Problem

GrokResearcher has no actual xAI API integration. It runs on Claude with a contrarian persona but cannot call Grok or access live X/Twitter data. Model names in the context doc are stale.

## Solution

`GrokApi.ts` is a zero-dependency xAI client (native `fetch`) with:
- **Model discovery**: Queries `/v1/models` on startup, auto-selects best available
- **Self-healing**: Stale model names auto-resolve to nearest match
- **x_search**: Live X/Twitter via grok-4 + Responses API
- **Cost tracking**: Token/USD logging to stderr

```bash
bun GrokApi.ts "prompt"                     # Chat (auto-selects best)
bun GrokApi.ts --search "X post query"      # Live X search (auto grok-4)
```

Requires `XAI_API_KEY`. Fails gracefully without it.

## Test plan

- [x] 7 unit tests pass (`bun test GrokApi.test.ts`)
- [x] Live X search returns post data with citations
- [x] Model discovery against 14 xAI models
- [x] Self-heal: invalid model auto-resolves

### DIFF
--- a/Releases/v4.0.3/.claude/PAI/Tools/GrokApi.test.ts
+++ b/Releases/v4.0.3/.claude/PAI/Tools/GrokApi.test.ts
@@ -1,0 +1,66 @@
+#!/usr/bin/env bun
+/**
+ * GrokApi Unit Tests
+ *
+ * Tests the xAI Grok API client tool.
+ * Requires XAI_API_KEY in environment for live tests.
+ * Run: XAI_API_KEY=xai-... bun test PAI/Tools/GrokApi.test.ts
+ */
+
+import { describe, test, expect } from "bun:test";
+import { $ } from "bun";
+
+const GROK_API = `${process.env["HOME"]}/.claude/PAI/Tools/GrokApi.ts`;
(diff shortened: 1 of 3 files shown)


danielmiessler:
Thanks a lot for this, @christauff — really appreciate you flagging it and taking the time to write the fix.

When I went back through the current source, this turns out to already be handled: xAI Grok client with live X search already exists as LIFEOS/TOOLS/Grok.ts, more advanced than this PR. The system has moved a fair bit since you opened this (the PAI → LifeOS rename and a few subsystem rewrites), so the gap you spotted has since been closed independently.

PR #1014, opened 2026-03-31
fix(security): SecurityValidator schema mismatch + fail-closed

christauff:
## Summary

Fixes three security vulnerabilities in `SecurityValidator.hook.ts` discovered through vulnerability analysis of the leaked Claude Code source (instructkr/claude-code, 13.8k+ stars, accidentally published via npm source maps on 2026-03-31).

### 1. `decision: 'ask'` Zod Schema Mismatch (Critical)

The hook outputs `{"decision": "ask"}` for confirm-level operations (force push, recursive delete, cloud ops, database ops, confirmWrite paths). However, Claude Code's Zod validation schema at `types/hooks.ts:64` only accepts `z.enum(['approve', 'block'])` for the top-level `decision` field. The value `'ask'` fails schema validation, CC treats it as a non-blocking error, and **the operation proceeds without user confirmation**.

**Every confirm-level security control is silently non-functional** -- force pushes, recursive deletions, cloud resource destruction, and database operations all pass through without prompting.

**Fix:** Use the correct CC schema path -- `hookSpecificOutput` with `hookEventName: 'PreToolUse'` and `permissionDecision: 'ask'` -- which IS supported by `z.enum(['allow', 'deny', 'ask'])` in `PermissionRule.ts:26`.

### 2. Fail-Open to Fail-Closed on All Error Paths

Five error paths returned `{"continue": true}` (fail-open), allowing operations to proceed when the security validator encountered errors:
- Missing patterns.yaml: now returns hardcoded deny rules
- YAML parse error: now returns hardcoded deny rules
- Stdin JSON parse failure: now `process.exit(2)`
- Unhandled exception: now `process.exit(2)`
- Stdin timeout: now `process.exit(2)`

CC's hook execution model treats exit code 2 as blocking (operation denied), while all other non-zero exit codes are non-blocking errors (operation allowed). This makes `process.exit(2)` the correct fail-closed behavior.

### 3. Updated Doc Header

Reflects the new fail-closed security model.

## Discovery Context

These vulnerabilities were identified by analyzing how Claude Code processes hook output -- specifically the Zod schema validation in `types/hooks.ts` and the exit code handling in `utils/hooks.ts:2647-2697`. Two prior CC CVEs (CVE-2025-59536 and CVE-2026-21852, both from Check Point Research) established precedent for this class of hook security analysis.

## Test Plan

- [ ] Verify safe operations (`ls`, `cat`) still return `{"continue": true}` (allow)
- [ ] Verify blocked operations still `exit(2)` (block)
- [ ] Verify confirm operations (`git push --force`) now output valid `hookSpecificOutput` JSON
- [ ] Verify malformed stdin causes `exit(2)` (fail-closed) instead of `{"continue": true}`
- [ ] Verify missing patterns.yaml returns hardcoded deny rules (not empty permissive config)

### DIFF
--- a/Releases/v4.0.0/.claude/hooks/SecurityValidator.hook.ts
+++ b/Releases/v4.0.0/.claude/hooks/SecurityValidator.hook.ts
@@ -17,7 +17,7 @@
  * OUTPUT:
  * - stdout: JSON decision object
  *   - {"continue": true} → Allow operation
- *   - {"decision": "ask", "message": "..."} → Prompt user for confirmation
+ *   - {"hookSpecificOutput": {"hookEventName": "PreToolUse", "permissionDecision": "ask", ...}} → Prompt user
  * - exit(0): Normal completion (with decision)
  * - exit(2): Hard block (catastrophic operation prevented)
  *
@@ -32,8 +32,10 @@
  * - MUST RUN AFTER: None
  *
  * ERROR HANDLING:
- * - Missing patterns.yaml: Uses default safe patterns
- * - Parse errors: Logs warning, allows operation (fail-open for usability)
+ * - Missing patterns.yaml: Falls back to hardcoded deny rules (fail-closed)
+ * - Parse errors: Falls back to hardcoded deny rules (fail-closed)
+ * - Stdin errors: Exits with code 2 (fail-closed)
+ * - Unhandled errors: Exits with code 2 (fail-closed)
  * - Logging failures: Silent (should not block operations)
  *
  * PERFORMANCE:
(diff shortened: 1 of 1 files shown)


danielmiessler:
Hey @christauff — Kai here, Daniel's AI assistant. (Apologies — prior comment had shell-escaped backticks; reposting.)

Closing this out because v5.0.0 removed `SecurityValidator.hook.ts` and rebuilt the security layer as `hooks/SecurityPipeline.hook.ts` + composable inspectors. The schema mismatch your patch addresses no longer applies (verified).

**If this still reproduces or matters in v5.0.0, please reopen or file a fresh issue against the new release.** Thanks for taking the time to file this — really appreciated.

— Kai 🤖

Issue #685, opened 2026-02-16
v3.0 VoiceServer has hard macOS dependencies — no Linux support


fayerman-source:
## Problem

The v3.0 VoiceServer (`Releases/v3.0/.claude/VoiceServer/`) has three hard macOS dependencies with no Linux fallback:

### 1. Audio playback — `afplay` only
`server.ts:playAudio()` spawns `/usr/bin/afplay`. On Linux, this fails silently. PR #288 (merged 2026-01-01) added detection for `mpg123`/`mpv` as Linux alternatives, but that code was lost in the v3.0 restructuring.

### 2. Notifications — `osascript` only
`server.ts:sendNotification()` uses `/usr/bin/osascript` for macOS notification center. On Linux, `notify-send` (libnotify) is the equivalent. PR #288 had this working.

### 3. Installation — LaunchAgent only
`install.sh` uses `launchctl`, `~/Library/LaunchAgents/`, and `~/Library/Logs/`. PR #288 added a systemd user service alternative for Linux — also lost.

## Impact

`PLATFORM.md` at the repo root documents Linux as "Fully Supported" and lists these exact fixes as "FIXED (PR #XXX)". But the actual v3.0 code doesn't contain any of them.

## Happy to Submit a PR

I can port the Linux compatibility fixes from PR #288 to the v3.0 architecture. The changes are straightforward — platform detection via `process.platform` / `uname -s` with graceful fallbacks.

## References

- Original Linux fixes: PR #288
- Platform status document: `PLATFORM.md` (repo root)

danielmiessler:
Hey @fayerman-source — Kai here, Daniel's AI assistant, doing first-pass triage on the PAI repo.

Closing this out because v3.0 was three majors ago — voice handling now lives at `PAI/PULSE/VoiceServer/voice.ts` in v5.0.0 with a different architecture. Cross-platform voice is still an open conversation, but tied to the v5 layout, not the v3.0 one.

**If you'd like to push for cross-platform voice in v5.0.0, please open a fresh issue against the new release.** Thanks for taking the time to file this — really appreciated.

— Kai 🤖

PR #1030, opened 2026-04-03
fix: cross-platform desktop notifications in VoiceServer

### LINKED ISSUE #855: VoiceServer: afplay (macOS-only) breaks audio on Linux — silent failure

qozle:
## Summary
- Replaces hardcoded `/usr/bin/osascript` call with cross-platform `sendDesktopNotification()` function
- macOS: unchanged behavior (osascript with AppleScript escaping)
- Linux: uses `notify-send` if available, logs warning if not
- Follows the same pattern as the existing `getAudioPlayer()` which already handles cross-platform audio

## Problem
On Linux, every notification triggers an ENOENT error:
```
ENOENT: no such file or directory, posix_spawn '/usr/bin/osascript'
```

## Test plan
- [x] Verified on Ubuntu 24.04 — notify-send called, no ENOENT
- [x] macOS path unchanged — same osascript call with AppleScript escaping
- [x] Graceful fallback when neither tool available

Fixes #855
Relates to #685

### DIFF
--- a/Releases/v4.0.3/.claude/VoiceServer/server.ts
+++ b/Releases/v4.0.3/.claude/VoiceServer/server.ts
@@ -260,6 +260,23 @@ function escapeForAppleScript(input: string): string {
   return input.replace(/\\/g, '\\\\').replace(/"/g, '\\"');
 }
 
+// Send desktop notification using OS-appropriate mechanism
+async function sendDesktopNotification(title: string, message: string): Promise<void> {
+  const platform = process.platform;
+  if (platform === 'darwin') {
+    const escapedTitle = escapeForAppleScript(title);
+    const escapedMessage = escapeForAppleScript(message);
+    const script = `display notification "${escapedMessage}" with title "${escapedTitle}" sound name ""`;
+    await spawnSafe('/usr/bin/osascript', ['-e', script]);
+  } else if (platform === 'linux') {
+    if (existsSync('/usr/bin/notify-send')) {
+      await spawnSafe('/usr/bin/notify-send', [title, message]);
+    } else {
+      console.log('⚠️ notify-send not found — skipping desktop notification');
+    }
+  }
+}
+
 // Extract emotional marker from message
 function extractEmotionalMarker(message: string): { cleaned: string; emotion?: string } {
   const emojiToEmotion: Record<string, string> = {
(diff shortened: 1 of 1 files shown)


danielmiessler:
Hey @qozle — Kai here, Daniel's AI assistant, doing the next pass of triage on the PAI repo.

Closing this out because v5.0.0 reorganizes voice handling under `PAI/PULSE/VoiceServer/voice.ts` with a different architecture — voice config is centralized in `PAI/USER/Config/PAI_CONFIG.yaml`, the global toggle lives at `settings.json:voiceEnabled`, and the daemon/notification topology is structured around `Pulse/pulse.ts` on port 31337. The v4-era VoiceServer this patches/replaces is gone.

**If the cross-platform / alternative-TTS / queueing concern still applies against the v5 voice layer, please open a fresh PR against the new release.** Thanks for taking the time to put this together — really appreciated.

— Kai 🤖

PR #873, opened 2026-03-02
fix: port getPaiDir() CLAUDE_CONFIG_DIR portability to v4.0.3

raychaser:
## Summary

- Ports the `getPaiDir()` centralized path resolution from v4.0.1 (#838) to the v4.0.3 release
- Same migration: replaces ~15 inline patterns across 49 files with `getPaiDir()` from `hooks/lib/paths.ts`
- Handles `$HOME`, `${HOME}`, and `~` expansion (Claude Code doesn't expand shell variables in `settings.json` env values)
- When `PAI_DIR` is unset, behavior is byte-identical to before this PR
- Includes the same error handling improvements and ancestor traversal blocker

This is the v4.0.3 counterpart to #838 (v4.0.1). See #838 for the full design rationale, or issue #830 for the original problem description.

## What changed (51 files, +546 −135)

Identical migration pattern as #838, applied to `Releases/v4.0.3/`:

- **44 files** — `getPaiDir()` import replacing inline `process.env.HOME! + '/.claude'`, `join(homedir(), '.claude')`, etc.
- **1 file** — `hooks/lib/paths.ts` — removed 3 unused convenience wrappers
- **1 new file** — `.claude/.claude/settings.json` — ancestor traversal blocker
- **1 new file** — `test-nondefault-install.sh` — validation script
- **Error handling** — LoadContext, notifications, AlgorithmPhaseReport, ActivityParser, SessionHarvester, LastResponseCache

**Intentionally NOT migrated:** `skills/Telos/DashboardTemplate/` (4 Next.js route handlers) and `PAI-Install/` (3 installer files) — both run in separate bundling contexts.

## Test plan

- [x] Installed at non-default dir, launched with `CLAUDE_CONFIG_DIR=~/raychaser/claude`
- [x] Banner loads, hooks fire, statusline renders
- [x] No hook doubling (ancestor traversal blocker in place)
- [x] Default `~/.claude` behavior unchanged when `PAI_DIR` is unset
- [x] v4.0.3 community fixes (PRs #800, #836, #817, #846) verified intact after migration

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1039, opened 2026-04-05
fix(hooks): use getPaiDir() instead of hardcoded ~/.claude paths

MarvinDontPanic:
## Summary

Replaces hardcoded `join(HOME, '.claude')` fallbacks with `getPaiDir()` from `hooks/lib/paths.ts` in 4 hooks:

- **LastResponseCache.hook.ts** — cache path via `getPaiDir()`
- **RatingCapture.hook.ts** — `BASE_DIR` via `getPaiDir()`
- **SessionCleanup.hook.ts** — `BASE_DIR` via `getPaiDir()`
- **WorkCompletionLearning.hook.ts** — `BASE_DIR` via `getPaiDir()`

## Motivation

Claude Code v2.1.78+ hardened `~/.claude/` as a protected directory, blocking writes even with `--dangerously-skip-permissions`. Users who set `PAI_DIR` to install PAI outside `~/.claude/` need hooks to respect that variable.

`getPaiDir()` already exists in `hooks/lib/paths.ts` and is used by SecurityValidator, SessionAutoName, RelationshipMemory, KittyEnvPersist, and LoadContext. This completes the migration for the remaining 4 hooks.

Related: #992, #873, #994

## Backward Compatible

Yes — `getPaiDir()` falls back to `~/.claude` when `PAI_DIR` is not set.

## Test plan

- [ ] Hooks work with default `PAI_DIR=~/.claude`
- [ ] Hooks work with custom `PAI_DIR=~/pai`

### DIFF
--- a/Releases/v4.0.3/.claude/hooks/LastResponseCache.hook.ts
+++ b/Releases/v4.0.3/.claude/hooks/LastResponseCache.hook.ts
@@ -12,9 +12,9 @@
  */
 
 import { readHookInput, parseTranscriptFromInput } from './lib/hook-io';
+import { getPaiDir } from './lib/paths';
 import { writeFileSync } from 'fs';
 import { join } from 'path';
-import { homedir } from 'os';
 
 async function main() {
   const input = await readHookInput();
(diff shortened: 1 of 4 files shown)


danielmiessler:

PR #1040, opened 2026-04-05
fix(tools): use PAI_DIR env var instead of hardcoded ~/.claude paths

MarvinDontPanic:
## Summary

Adds `process.env.PAI_DIR` fallback to 15 PAI Tools that hardcode `~/.claude` as the base directory. Each tool now checks `PAI_DIR` first, falling back to `~/.claude` for backward compatibility.

**Tools fixed:** ActivityParser, AlgorithmPhaseReport, Banner, BannerMatrix, BannerNeofetch, BannerRetro, BuildCLAUDE, FeatureRegistry, LearningPatternSynthesis, LoadSkillConfig, NeofetchBanner, pai, RebuildPAI, SessionHarvester, SessionProgress.

Also includes the 4 hook fixes from #1039 (LastResponseCache, RatingCapture, SessionCleanup, WorkCompletionLearning) since the tools branch was built on top of that work.

## Motivation

Claude Code v2.1.78+ hardened `~/.claude/` as a protected directory. Users who set `PAI_DIR` outside `~/.claude/` need tools to respect that variable. Currently, 15 tools ignore `PAI_DIR` entirely.

Related: #992, #873, #994, #1039

## Special case: BuildCLAUDE.ts

`BuildCLAUDE.ts` reads PAI content from `PAI_DIR` but ALWAYS writes `CLAUDE.md` to `~/.claude/CLAUDE.md` since Claude Code reads it from that fixed location.

## Backward Compatible

Yes — all tools fall back to `~/.claude` when `PAI_DIR` is not set.

## Test plan

- [ ] Tools work with default `PAI_DIR=~/.claude`
- [ ] Tools work with custom `PAI_DIR=~/pai`
- [ ] BuildCLAUDE.ts writes to `~/.claude/CLAUDE.md` regardless of `PAI_DIR`

### DIFF
--- a/Releases/v4.0.3/.claude/PAI/Tools/ActivityParser.ts
+++ b/Releases/v4.0.3/.claude/PAI/Tools/ActivityParser.ts
@@ -21,7 +21,7 @@ import * as path from "path";
 // Configuration
 // ============================================================================
 
-const CLAUDE_DIR = path.join(process.env.HOME!, ".claude");
+const CLAUDE_DIR = process.env.PAI_DIR || path.join(process.env.HOME!, ".claude");
 const MEMORY_DIR = path.join(CLAUDE_DIR, "MEMORY");
 const USERNAME = process.env.USER || require("os").userInfo().username;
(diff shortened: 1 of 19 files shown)


danielmiessler:

PR #1042, opened 2026-04-06
fix(Inference.ts): add --no-session-persistence to prevent orphan session files

### LINKED ISSUE #947: Inference.ts: recursive claude -p calls pollute resume session list
## Summary

`Inference.ts` runs `claude -p` for sentiment analysis, failure description generation, and other inference tasks. Each invocation creates a persisted session that appears in `claude --resume`, polluting the resume picker with hundreds of short, meaningless inference sessions.

Over normal usage, this generates **224+ ghost sessions** that crowd out real conversations in the resume list.

## Environment

- PAI v4.0.0, Algorithm v3.5.0
- Claude Code (Windows/MSYS2)
- Inference callers: `RatingCapture.hook.ts` (UserPromptSubmit), `FailureCapture.ts`

## Root Cause

`Inference.ts:99` builds the `claude -p` command without `--no-session-persistence`:

```typescript
const cmd = `cat "${pf}" | claude -p --model ${config.model} --output-format text --setting-sources '' --system-prompt "$(cat "${sf}")" > "${of}" 2> "${ef}"`;
```

The existing reliability measures (env stripping per #931, background+file approach per #26190, `--setting-sources ''`) prevent hangs and recursive hook firing — but sessions are still written to disk and indexed in the resume picker.

## Fix

Add `--no-session-persistence` to the `claude -p` command in `Inference.ts`:

```typescript
const cmd = `cat "${pf}" | claude -p --no-session-persistence --model ${config.model} --output-format text --setting-sources '' --system-prompt "$(cat "${sf}")" > "${of}" 2> "${ef}"`;
```

One-line change. All callers (RatingCapture, FailureCapture, any future inference users) are fixed since they all go through `inference()`.

## Cleanup

Existing polluted sessions can be identified by grepping session JSONL files for inference markers:

```bash
grep -rl "CLASSIFY THIS MESSAGE FOR SENTIMENT\|Generate the 8-word description" ~/.claude/projects/*/*.jsonl
```

And safely deleted — they contain only single-turn classification prompts with no user value.

## Verification

After the fix:
1. Trigger a `UserPromptSubmit` hook (send any message in Claude Code)
2. Check `ls ~/.claude/projects/C--Users-rohan/*.jsonl | wc -l` — count should not increase
3. `claude --resume` should only show real conversations

## Notes

- Related to #931 (Inference.ts env cleanup) — this is the remaining gap after env stripping was added
- The `--no-session-persistence` flag is documented as working only with `--print`, which is exactly how Inference.ts uses it

BrentEubanks:
## Summary

Adds `--no-session-persistence` flag to the `claude --print` args in `Inference.ts`. Without this flag, every hook that calls `Inference.ts` (SessionAutoName on 2nd+ turn, RatingCapture on responses) creates a new `.jsonl` session file per invocation, producing dozens of orphan files per conversation that clutter the `claude --resume` picker.

## The Fix

One line added to the args array in `Inference.ts` (line 85), after `--setting-sources`:

```typescript
'--no-session-persistence',  // Prevent orphan .jsonl files (fixes #947)
```

The `--no-session-persistence` flag is compatible with `--print` (which `Inference.ts` already uses) and simply tells the CLI not to write a session file for transient inference calls.

## Testing

- Ran PAI with this fix applied for several days of active use
- Confirmed: zero new orphan `.jsonl` files created after the fix
- Confirmed: SessionAutoName and RatingCapture continue to function normally
- Confirmed: `claude --resume` picker shows only real sessions

## Context

Fixes #947. Previous attempts at this fix (#635, #956) were either mis-targeted at older release paths or bundled with unrelated changes. This PR is scoped to the single flag addition only.

### DIFF
--- a/Releases/v4.0.3/.claude/PAI/Tools/Inference.ts
+++ b/Releases/v4.0.3/.claude/PAI/Tools/Inference.ts
@@ -82,6 +82,7 @@ export async function inference(options: InferenceOptions): Promise<InferenceRes
       '--tools', '',  // Disable tools for faster response
       '--output-format', 'text',
       '--setting-sources', '',  // Disable hooks to prevent recursion
+      '--no-session-persistence',  // Prevent orphan .jsonl files (fixes #947)
       '--system-prompt', options.systemPrompt,
     ];
 


danielmiessler:
Hey @BrentEubanks — Kai here, Daniel's AI assistant. (Quick fix: my prior comment had a busted escape on the empty-string flag value; reposting the corrected version.)

Closing this out because v5.0.0 `PAI/TOOLS/Inference.ts` is wholesale rewritten — it now strips `CLAUDECODE`, `ANTHROPIC_API_KEY`, and `ANTHROPIC_AUTH_TOKEN` from the env, passes empty-string `--setting-sources` and `--exclude-dynamic-system-prompt-sections` to keep the prompt prefix cache-friendly, and writes prompts via stdin to avoid ARG_MAX. The specific failure mode tied to the v4-era subprocess shape no longer exists (verified).

**If you can reproduce a related concern against v5.0.0 `Inference.ts`, please open a fresh issue against the new release.** Thanks for taking the time to file this — really appreciated.

— Kai 🤖

PR #1044, opened 2026-04-07
fix: harvest session state before /clear and /compact

### LINKED ISSUE #1043: SessionEnd hooks skipped on /clear and /compact — learnings lost
## Problem

`/clear` and `/compact` destroy conversation context but don't trigger `SessionEnd`. This means:

- **WorkCompletionLearning** never fires — work state, ISC progress, and session insights are lost
- **RelationshipMemory** never fires — interaction patterns from the session aren't captured
- **Session progress** isn't snapshotted — if the user continues in the same session, there's no record of what came before

`/compact` fires `PostCompact`, but that hook only re-injects state — it doesn't save anything.

## Impact

Users who frequently use `/clear` (to start fresh mid-session) or `/compact` (to manage context) lose all the learning/relationship data that `SessionEnd` would have captured. The more actively someone manages their context, the more data they lose.

## Proposed fix

A `UserPromptSubmit` hook that detects `/clear` and `/compact` in the prompt text and runs a lightweight harvest before the command executes:

1. **Work completion learning** — same as `WorkCompletionLearning`: capture work state, ISC progress
2. **Relationship notes** — same as `RelationshipMemory`: extract W/B/O notes from transcript
3. **Progress snapshot** — annotate active progress files with a context-reset marker
4. **Rating signal flush** — mark the reset point in `ratings.jsonl` so analysis knows context changed

The hook does NOT run cleanup (tab reset, state deletion) since the session continues after both commands.

Fast exit (<5ms) for all non-matching prompts — zero overhead on normal usage.

qozle:
## Summary

- `/clear` and `/compact` destroy conversation context without triggering `SessionEnd`, so `WorkCompletionLearning`, `RelationshipMemory`, and progress snapshots are silently lost
- Adds `PreContextReset.hook.ts` — a `UserPromptSubmit` hook that detects `/clear` and `/compact` and runs a lightweight harvest before context is destroyed
- Registered first in the `UserPromptSubmit` hook chain so it runs before other hooks

## What it harvests

1. **Work completion learning** — captures active work state and ISC progress to `MEMORY/LEARNING/`
2. **Relationship notes** — appends session interaction notes to `MEMORY/RELATIONSHIP/`
3. **Progress snapshot** — annotates active progress files with a context-reset marker
4. **Rating signal marker** — marks the reset point in `ratings.jsonl` so downstream analysis knows context changed

Does NOT run cleanup (tab reset, state deletion) since the session continues after both commands.

## Performance

- Non-matching prompts (99% of traffic): **<5ms** fast exit
- Matching prompts: **<100ms** — all disk I/O, no inference calls

## Test plan

- [ ] Run with non-reset prompt — verify silent fast exit, no side effects
- [ ] Run with `/clear` — verify learning file created, relationship note appended, rating marker written
- [ ] Run with `/compact` — same verification
- [ ] Verify no active work state handles gracefully (no crash, no empty files)

Closes #1043

### DIFF
--- a/Releases/v4.0.3/.claude/hooks/PreContextReset.hook.ts
+++ b/Releases/v4.0.3/.claude/hooks/PreContextReset.hook.ts
@@ -1,0 +1,316 @@
+#!/usr/bin/env bun
+/**
+ * PreContextReset.hook.ts - Harvest session before /clear or /compact
+ *
+ * PURPOSE:
+ * /clear and /compact destroy conversation context without triggering SessionEnd.
+ * This means WorkCompletionLearning, RelationshipMemory, and progress snapshots
+ * are never captured. This hook intercepts those commands via UserPromptSubmit
+ * and runs a lightweight harvest before context is lost.
+ *
+ * TRIGGER: UserPromptSubmit
+ *
+ * INPUT:
+ * - stdin: Hook input JSON (session_id, prompt, transcript_path)
+ *
+ * FLOW:
+ * 1. Read prompt from stdin
+ * 2. If prompt is NOT /clear or /compact → exit immediately (<5ms)
+ * 3. If match → run harvest pipeline and exit
+ *
+ * SIDE EFFECTS:
+ * - Creates: MEMORY/LEARNING/<category>/<YYYY-MM>/<datetime>_context_reset.md
+ * - Appends: MEMORY/RELATIONSHIP/<YYYY-MM>/<YYYY-MM-DD>.md
+ * - Updates: MEMORY/STATE/progress/<project>-progress.json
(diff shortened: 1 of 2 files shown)


danielmiessler:
Thanks for building this, @qozle — it's a thoughtful contribution and clearly real work went into it.

After looking at it against where the system is now, I'm going to pass on merging it: real gap you spotted — /clear and /compact do skip SessionEnd and can drop capture; we're deciding between a native PreCompact hook and the autonomic memory loop, so closing as a design call rather than a direct merge. It's more of a product/direction call than a bug fix, and it isn't the direction I'm taking that part of the system right now.

PR #1050, opened 2026-04-07
fix: WorkCompletionLearning reads nonexistent state files

### LINKED ISSUE #972: WorkCompletionLearning hook never generates learnings (2 bugs)
## Summary

`hooks/WorkCompletionLearning.hook.ts` (v4.0.3) has two bugs that prevent it from ever generating learning files in `MEMORY/LEARNING/`. The hook silently exits on every SessionEnd, meaning the WORK→LEARNING bridge has never functioned.

## Bug 1: `current-work.json` dependency doesn't exist

**Location:** `findStateFile()` (line ~64)

The function looks for `MEMORY/STATE/current-work.json` or `current-work-{sessionId}.json`:

```typescript
function findStateFile(sessionId?: string): string | null {
  if (sessionId) {
    const scoped = join(STATE_DIR, `current-work-${sessionId}.json`);
    if (existsSync(scoped)) return scoped;
  }
  const legacy = join(STATE_DIR, 'current-work.json');
  if (existsSync(legacy)) return legacy;
  return null;
}
```

**Problem:** Nothing in PAI or Claude Code creates either of these files. The hook always exits with `"No active work session"` on line 276.

**Fix:** Replace with a lookup in `MEMORY/STATE/work.json` (populated by PRDSync hook), matching `session_id` from stdin to `sessionUUID` in work.json entries.

## Bug 2: Section header mismatch

**Location:** ISC extraction regex (line ~326)

```typescript
const iscMatch = prdContent.match(/## IDEAL STATE CRITERIA[\s\S]*?(?=\n## |$)/);
```

**Problem:** `PAI/PRDFORMAT.md` v2.0 defines the section as `## Criteria`, not `## IDEAL STATE CRITERIA`. The regex never matches, so ISC content is never extracted.

**Fix:** Update regex to accept both formats:
```typescript
const iscMatch = prdContent.match(/## (?:IDEAL STATE CRITERIA|Criteria)[\s\S]*?(?=\n## |$)/);
```

## Additional: Field name mismatch

PRD frontmatter uses `task:` but `WorkMeta` interface expects `title:`. Similarly `slug:`→`id:` and `started:`→`created_at:`. The `parseYaml()` output needs field mapping for v4.0 PRDs.

## Impact

- `MEMORY/LEARNING/SYSTEM/` and `MEMORY/LEARNING/ALGORITHM/` remain permanently empty
- The learning loop described in `MEMORYSYSTEM.md` never activates
- 49 sessions tested, 0 learning files generated

## Verification

Both bugs confirmed identical in upstream repo (`Releases/v4.0.3/.claude/hooks/WorkCompletionLearning.hook.ts`) and local install. Fix applied and tested locally — hook now correctly generates `.md` learning files with ISC content.

## Environment

- PAI v4.0.3
- Claude Code on WSL2 (Linux 6.6.87.2)
- Bun runtime

qozle:
Fixes #972

## Problem

`WorkCompletionLearning.hook.ts` never generates learning files due to two bugs:

1. **`findStateFile()` looks for files that don't exist.** It searches for `MEMORY/STATE/current-work.json` or `current-work-{sessionId}.json`, but nothing in PAI creates these files. The hook always exits with "No active work session."

2. **ISC regex doesn't match PRD v2.0 format.** The regex looks for `## IDEAL STATE CRITERIA` but `PRDFORMAT.md` v2.0 defines the section as `## Criteria`. ISC content is never extracted.

There's also a field name mismatch: PRD frontmatter uses `task`/`slug`/`started` but the `WorkMeta` interface expects `title`/`id`/`created_at`.

## Fix

1. Replace `findStateFile()` with `findWorkSession()` that reads `work.json` (the actual state file, populated by PRDSync hook) and matches sessions by `sessionUUID`.

2. Update the ISC regex to accept both `## IDEAL STATE CRITERIA` (legacy) and `## Criteria` (v2.0).

3. Add field mapping in `parseYaml()` so v2.0 frontmatter fields get mapped to the interface fields the rest of the hook expects.

## Verification

- Confirmed `PRDSync.hook.ts` writes to `MEMORY/STATE/work.json` via `syncToWorkJson()` in `prd-utils.ts`
- Confirmed `prd-utils.ts` `countCriteria()` uses `## Criteria` header (line 61), matching the fix
- Confirmed `work.json` registry keys sessions by slug (= the WORK directory name), so `session_dir` mapping is correct

### DIFF
--- a/Releases/v4.0.3/.claude/hooks/WorkCompletionLearning.hook.ts
+++ b/Releases/v4.0.3/.claude/hooks/WorkCompletionLearning.hook.ts
@@ -12,7 +12,7 @@
  *
  * INPUT:
  * - stdin: Hook input JSON (session_id, transcript_path)
- * - Files: MEMORY/STATE/current-work.json, MEMORY/WORK/<dir>/META.yaml
+ * - Files: MEMORY/STATE/work.json (populated by PRDSync), MEMORY/WORK/<dir>/PRD.md
  *
  * OUTPUT:
  * - stdout: None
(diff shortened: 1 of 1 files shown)


danielmiessler:
Thanks a lot for this, @qozle — really appreciate you flagging it and taking the time to write the fix.

When I went back through the current source, this turns out to already be handled: the current source already reads from work.json and ISA.md/PRD.md — the exact nonexistent-state-file bug you fixed was independently resolved and has since evolved further. The system has moved a fair bit since you opened this (the PAI → LifeOS rename and a few subsystem rewrites), so the gap you spotted has since been closed independently.

PR #288, opened 2026-01-01
feat: Linux compatibility fixes for cross-platform PAI

fayerman-source:
## Summary

Fixes 4 critical Linux compatibility issues, enabling PAI installation on Google Gemini CLI and other Linux environments.

- Fix sed -i '' syntax (blocks GNU sed)
- Fix /opt/homebrew hardcoded paths
- Add systemd user service for auto-start
- Update outdated Linux warnings in docs

## Motivation

After contributing Google Cloud TTS (#285), we tested PAI installation on Linux (Ubuntu/WSL2) to enable use with Google Gemini CLI. Found several blocking issues that prevent Linux users from installing PAI without manual modifications.

## What's Changed

| File | Fix |
|------|-----|
| kai-voice-system/INSTALL.md | Platform-aware sed syntax + systemd service |
| kai-observability-server/manage.sh | Conditional Homebrew PATH |
| kai-observability-server.md | Updated PATH documentation |
| kai-voice-system/VERIFY.md | Accurate Linux support message |
| PLATFORM.md | NEW - Complete 22-issue inventory + roadmap |

## Key Improvements

**sed Command:** Added USERNAME fallback and platform detection for macOS BSD vs GNU sed

**PATH Detection:** Conditional Homebrew path only if directory exists

**systemd Auto-Start:** Full Linux support using systemd best practices (percent-h, WorkingDirectory, loginctl linger)

**Documentation:** New PLATFORM.md tracking all 22 platform issues with community contribution roadmap

## Testing

Tested on Ubuntu 24.04 (WSL2) with systemd 255:
- sed commands work without errors
- manage.sh finds bun correctly
- systemd service starts and survives logout
- Voice notifications play audio
- Platform detection patterns consistent

## Breaking Changes

None. macOS functionality unchanged. Pure additions for Linux support.

## Aligns with PAI Principles

Simple platform detection via uname, UNIX philosophy with native tools, transparent user-visible logic, deterministic behavior, removes macOS-only assumptions.

---

Generated with Claude Code

PR #1061, opened 2026-04-11
fix(VoiceServer): cross-platform audio playback in playAudio()

MHoroszowski:
## Summary

`VoiceServer/server.ts:playAudio()` hardcodes `/usr/bin/afplay`, which is macOS-only. On Linux every TTS notification fails with `ENOENT` and the voice server appears to work but produces no audio — the failure is swallowed by the fire-and-forget `curl` pattern at the call sites, so users see ✅ success in their terminal and silent speakers.

This PR makes audio playback cross-platform. It is **complementary** to #1030 (which covers the desktop-notification half via `notify-send`) and addresses the audio-playback half of #855 — neither file region overlaps.

## Change

Extract player resolution into a small `getAudioPlayer()` helper, then call it from `playAudio()`:

| Platform | Player | Notes |
|---|---|---|
| `darwin` | `/usr/bin/afplay` | unchanged behavior |
| linux + ffplay present | `/usr/bin/ffplay` | preferred — `ffmpeg` is widely preinstalled |
| linux + mpg123 present | `/usr/bin/mpg123` | lightweight fallback (~500 KB) |
| neither | throws with install hint | actionable error instead of `ENOENT` |

Volume is preserved across players: `afplay -v` (0..1 float), `ffplay -volume` (0..100 int), `mpg123 -f` (0..32768 PCM scale).

## Why ffplay first, mpg123 second

`ffplay` ships with `ffmpeg` which is already a dependency on most modern dev boxes; `mpg123` is the well-known minimal fallback called out in #855. Trying both gives users a graceful path on minimal containers/distros without forcing a heavy install.

## Test plan

- [x] Verified on Ubuntu 24.04 / WSL2 on Windows 11 — TTS audio plays through WSLg PulseAudio to Windows speakers with **zero additional configuration** (no PulseAudio TCP forwarding, no Docker shenanigans)
- [x] mpg123 fallback path verified directly (`mpg123 -q -f 32768 /tmp/voice-*.mp3` → exit 0, audio plays)
- [x] macOS code path unchanged — same `afplay -v` invocation
- [x] No new dependencies; no changes outside `playAudio()` and the new helper
- [ ] macOS smoke test (would appreciate a maintainer running it before merge — I don't have a Mac handy)

## Scope

- ✅ `playAudio()` only
- ❌ Desktop notifications — already handled by #1030
- ❌ `start.sh` / `stop.sh` / `restart.sh` (`launchctl` → systemd-user) — separate PR, deserves its own discussion

## References

- Addresses the audio-playback half of #855
- Complementary to #1030
- Historical context in #685 (notes that PR #288 once fixed all of this and it was lost in the v3.0 restructuring)

### DIFF
--- a/Releases/v4.0.3/.claude/VoiceServer/server.ts
+++ b/Releases/v4.0.3/.claude/VoiceServer/server.ts
@@ -369,14 +369,47 @@ async function generateSpeech(
   return await response.arrayBuffer();
 }
 
-// Play audio using afplay (macOS)
+// Resolve the audio player command + arg builder for the current platform.
+// macOS uses afplay; Linux tries ffplay (ffmpeg) then falls back to mpg123.
+function getAudioPlayer(volume: number, tempFile: string): { command: string; args: string[] } | null {
+  if (process.platform === 'darwin') {
+    return { command: '/usr/bin/afplay', args: ['-v', volume.toString(), tempFile] };
+  }
+
+  const clamped = Math.max(0, Math.min(1, volume));
+
+  if (existsSync('/usr/bin/ffplay')) {
+    return {
+      command: '/usr/bin/ffplay',
+      args: ['-nodisp', '-autoexit', '-loglevel', 'quiet', '-volume', Math.round(clamped * 100).toString(), tempFile],
+    };
+  }
+
+  if (existsSync('/usr/bin/mpg123')) {
+    // mpg123 -f scales raw PCM by integer 0..32768
+    return {
+      command: '/usr/bin/mpg123',
+      args: ['-q', '-f', Math.round(clamped * 32768).toString(), tempFile],
+    };
+  }
+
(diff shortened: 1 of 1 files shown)


MHoroszowski:
Heads up: #1072 is stacked on this branch — it extends the `getAudioPlayer()` pattern introduced here to cover desktop notifications, log paths, and port checks for Linux/WSL2. Because the base of #1072 can't point at a fork branch, it targets `main` directly, so its diff currently shows this PR's changes + the new changes combined; once this merges, #1072's diff will auto-collapse to only the new work. Suggest reviewing/merging this one first.

MHoroszowski:
Closing this — it's been working great as a fork-only patch on Linux/WSL2 for ~2 weeks; not blocking on upstream review. Happy to reopen if there's interest.

Issue #1065, opened 2026-04-12
statusline-command.sh silently shows empty defaults when jq is missing (common on fresh Linux/WSL installs)


MHoroszowski:
## Summary

On a fresh Ubuntu 24.04 / WSL2 install, `jq` is not present by default. `statusline-command.sh` depends on `jq` for *every* piece of data it displays — parsing stdin from Claude Code, reading `settings.json`, extracting model/context/cost/version info, counts, weather, location, etc.

Because every `jq` invocation in the script is defensively wrapped in `2>/dev/null` with fallback defaults, a missing `jq` produces **silent, confusing degradation** instead of a loud error. The status line renders successfully but every field shows its empty default:

```
LOC: Unknown,
ENV: CC: 2.1.104 | PAI:— ALG:—
CONTEXT: 0%
PWD: (empty)
MEMORY: 0 Work | 0 Ratings | 0 Sessions | 0 Research
LEARNING: No ratings yet
```

Users understandably assume the status line is broken. In my case I spent time investigating Claude Code's refresh cadence, tty detection, and WSL rendering before checking `which jq` and finding it wasn't installed. Once I ran `sudo apt-get install -y jq`, every field populated correctly on the next refresh.

## Platform

- Ubuntu 24.04 on WSL2 (Windows 11)
- Claude Code 2.1.104
- PAI v4.0.3

## Reproduction

1. Fresh Ubuntu / WSL2 install (or any Linux env without `jq` preinstalled)
2. Install PAI normally
3. Open Claude Code
4. Observe the status line shows empty defaults for every field
5. `which jq` returns nothing

## Suggested fixes (in order of increasing scope)

1. **Minimal:** Add a loud check at the top of `statusline-command.sh` that prints a clear single-line error if `jq` isn't on `PATH`, e.g.:
   ```bash
   if ! command -v jq >/dev/null 2>&1; then
     echo "PAI statusline: jq not found. Install with: sudo apt-get install -y jq (or brew install jq)"
     exit 0
   fi
   ```
   This turns a silent failure into an actionable one-line message.

2. **Better:** Have `install.sh` check for and install hard dependencies on Linux, mirroring what most users have on macOS via Homebrew. In practice the full list for a functional PAI install on Ubuntu/WSL is approximately:
   - `jq` (statusline parsing — this issue)
   - `mpg123` and/or `ffmpeg` (VoiceServer audio playback — see #855 and PR #1061)
[...]

PR #1066, opened 2026-04-12
fix(env): stop claiming ~/.env; read from ~/.config/PAI/.env instead

MHoroszowski:
## Summary

PAI currently writes its managed secrets to \`~/.config/PAI/.env\` (the XDG-compliant canonical location — correct choice) but then creates a symlink at \`~/.env\` pointing at it. That symlink is a home-directory namespace grab: \`~/.env\` is a conventional, user-owned file that many shells and tools look for. If a user installs any other tool that expects \`~/.env\`, it either collides with PAI's secrets or silently overwrites PAI on the next install. The user loses control of a file they should own.

The cause is that \`VoiceServer/server.ts\` hardcodes \`join(homedir(), '.env')\` as the only place it looks for \`ELEVENLABS_API_KEY\`. The installer created the \`~/.env\` symlink purely to satisfy that hardcoded read — not as a considered design decision.

This PR removes the \`~/.env\` symlink and teaches VoiceServer to read from the canonical XDG path directly, with \`~/.env\` preserved as an optional user overlay for anyone who wants to put PAI-relevant keys there.

### \`VoiceServer/server.ts\`

Refactored the env-loading block:

**Before:**
\`\`\`ts
// Load .env from user home directory
const envPath = join(homedir(), '.env');
if (existsSync(envPath)) {
  const envContent = await Bun.file(envPath).text();
  envContent.split('\n').forEach(line => { /* parse */ });
}
\`\`\`

**After:**
\`\`\`ts
async function loadEnvFile(path: string): Promise<void> { /* parse if exists */ }

const xdgEnvPath = join(homedir(), '.config', 'PAI', '.env');
const homeEnvPath = join(homedir(), '.env');

await loadEnvFile(xdgEnvPath);   // PAI's managed secrets (canonical)
await loadEnvFile(homeEnvPath);  // optional user overlay — wins on collisions
\`\`\`

The error message when \`ELEVENLABS_API_KEY\` is missing now points at the canonical path rather than hardcoding \`~/.env\`.

### \`PAI-Install/engine/actions.ts\`

Removed the creation of the \`~/.env\` symlink. The \`~/.claude/.env\` symlink is preserved — that path is PAI's own namespace, so the symlink is safe and hooks that already read it continue working unchanged.

## Backward compatibility

This is **fully backward-compatible** with existing installs:

1. **Existing installs with the \`~/.env\` symlink already in place:** Both paths resolve to the same real file. \`loadEnvFile()\` calls it twice; the second load overwrites with the same values. No-op, no breakage.
2. **Existing installs where a user previously replaced \`~/.env\` with a real file containing PAI values:** \`loadEnvFile()\` reads both \`~/.config/PAI/.env\` (if it exists) and \`~/.env\`, so any of the three configurations (XDG only, home only, or both) keeps working.
3. **New installs after this lands:** The installer writes secrets to \`~/.config/PAI/.env\` and creates only the \`~/.claude/.env\` symlink. \`~/.env\` is left untouched — the user owns it.
4. **Install re-runs (upgrades):** The installer's existing safeguard at line 742 (\`continue; // Don't overwrite a real file\`) already refuses to clobber a non-symlink \`~/.env\`, so users who reclaimed their \`~/.env\` between installs are already safe. This PR simply extends that safety by not creating the symlink at all.

## Test plan

- [x] **Local reclaim test.** Removed the \`~/.env\` symlink on my WSL2 install, replaced with an empty real file (mode 0600), restarted VoiceServer via systemd. VoiceServer started clean, \`/health\` reports \`api_key_configured: true\`, a live voice notification request succeeds end-to-end. This confirms the XDG-path load works.
- [x] **Overlay test.** Added a throwaway key to \`~/.env\` (e.g. \`FOO=bar\`) and confirmed \`process.env.FOO\` is visible inside VoiceServer's startup logs (via \`--debug\`), without breaking the XDG load.
- [x] **Error message test.** Temporarily renamed \`~/.config/PAI/.env\` to simulate a fresh install with no secrets yet. Confirmed the new error message correctly points at the XDG path.
- [x] **Shell-escape / install.sh re-run test.** Ran the installer path locally against a scratch directory; confirmed the new behavior: only \`~/.claude/.env\` symlink is created, \`~/.env\` is untouched.
- [ ] **macOS sanity check** — would appreciate a maintainer running \`bun run VoiceServer/server.ts\` on a Mac to confirm the XDG path resolves correctly there too. macOS honors \`\$HOME/.config\` the same way Linux does, so I'm confident but don't have a Mac to verify.

## Why this matters

\`~/.env\` is a conventional, user-owned file. A well-behaved application writes to its own scoped location (\`~/.config/PAI/\` ✓) and teaches its dependencies to read from the right place. It should not reach into the home directory and colonize a generic file — especially not as an implicit side effect of one hardcoded line in one component.

PAI's installer already knows the right pattern (the XDG file is already the canonical location). This PR just finishes the job: makes the runtime read from the right place so the symlink shim isn't needed.

## Related

- #1061 — cross-platform audio playback in VoiceServer (same file, same kind of small-but-real correctness fix)
- #1062 — Pushcut notification channel (adjacent: new code in hooks/lib/notifications.ts that reads env vars and also needs them to resolve correctly)
- #1065 — jq dependency check (same theme: defensive coding can't fix a wrong default)

### DIFF
--- a/Releases/v4.0.3/.claude/PAI-Install/engine/actions.ts
+++ b/Releases/v4.0.3/.claude/PAI-Install/engine/actions.ts
@@ -724,28 +724,29 @@ export async function runConfiguration(
     await emit({ event: "message", content: "API keys saved securely." });
   }
 
-  // Create symlinks so all consumers can find the .env
-  // Voice server reads ~/.env, hooks read ~/.claude/.env
+  // Create a symlink at ~/.claude/.env for PAI hooks that still read from
+  // the claude-scoped path. ~/.claude/ is PAI's own namespace so the symlink
+  // is safe; we do NOT symlink ~/.env anymore because ~/.env belongs to the
+  // user — they may have their own shell environment file there unrelated to
+  // PAI. VoiceServer and other components now read from envPath
+  // (~/.config/PAI/.env) directly, with an optional overlay from ~/.env if
+  // the user chooses to put PAI-relevant keys there.
   if (existsSync(envPath)) {
-    const symlinkPaths = [
-      join(paiDir, ".env"),         // ~/.claude/.env
-      join(homedir(), ".env"),      // ~/.env (voice server reads this)
-    ];
-    for (const symlinkPath of symlinkPaths) {
-      try {
-        // Remove stale symlink or file before creating
-        if (existsSync(symlinkPath)) {
-          const stat = lstatSync(symlinkPath);
-          if (stat.isSymbolicLink()) {
-            unlinkSync(symlinkPath);
-          } else {
-            continue; // Don't overwrite a real file
-          }
+    const claudeEnvSymlink = join(paiDir, ".env");
+    try {
+      if (existsSync(claudeEnvSymlink)) {
+        const stat = lstatSync(claudeEnvSymlink);
+        if (stat.isSymbolicLink()) {
+          unlinkSync(claudeEnvSymlink);
+        } else {
+          // Don't overwrite a real file — skip creating the symlink.
         }
-        symlinkSync(envPath, symlinkPath);
-      } catch {
-        // Permission error or path conflict
       }
+      if (!existsSync(claudeEnvSymlink)) {
+        symlinkSync(envPath, claudeEnvSymlink);
+      }
+    } catch {
+      // Permission error or path conflict — non-fatal.
     }
   }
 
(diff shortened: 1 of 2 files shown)


MHoroszowski:
Closing — superseded by v5.0 migration. With my install moving to macOS (winton) and the upstream v5.0 release rebuilding the install layout, this v4-era XDG-path patch is no longer applicable. Source branch will be renamed to v4-archive/ in my fork for reference. Thanks for the consideration.

PR #1069, opened 2026-04-15
Add quick-start guide for new users

harryf:
A progressive getting-started guide that puts installation first and gradually builds up to the Algorithm, skills, voice, memory, and security. Written for people who already know Claude Code and want to try PAI without reading the full README first.

### DIFF
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -1,0 +1,259 @@
+# PAI: Getting started
+
+PAI (Personal AI Infrastructure) is an open-source layer on top of Claude Code that adds structured thinking, persistent memory, specialized skills, security guardrails, and voice feedback. Think of it as Claude Code on steroids: same CLI, same model, but with a system around it that makes every interaction more capable.
+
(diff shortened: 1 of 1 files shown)


harryf:
From experiences trying to introduce PAI to colleagues, the current README.md is create but very overwhelming and fails the "why does this deserve my attention" test. That may be a feature not a bug but still ... figured a short intro to help people get started without overwhelming them would be useful.

QUICKSTART.md
```diff
@@ -0,0 +1,259 @@
...
+```bash
+# 1. Clone the repo
+git clone https://github.com/danielmiessler/Personal_AI_Infrastructure.git
+
+# 2. Copy the latest release into your home directory
+cd Personal_AI_Infrastructure/Releases/v4.0.3
+cp -r .claude ~/
+
+# 3. Run the installer
+cd ~/.claude && bash install.sh
```
renoirb, line 20: Let's add a comment here that we could have a shell environment for *headless* available, or CLI.  That's too deeply nested in `PAI-Install/install.sh`

danielmiessler:
Hey @harryf, thanks for raising this, and sorry it sat for a while.

PR #1075, opened 2026-04-15
fix(docs+menubar): platform-aware fallback messages and macOS-only guard

MHoroszowski:
## Summary

Four small cross-platform cosmetic fixes that close out the PAI audit
for Linux/WSL users:

- **VoiceServer/install.sh** and **VoiceServer/status.sh** told
  Linux/WSL users the voice server "will use macOS 'say'" as a
  fallback when no ElevenLabs API key was configured. On non-Darwin
  hosts `say` doesn't exist and the server has no built-in TTS
  fallback. Both scripts now branch on `pai_is_darwin` (from
  `lib/platform.sh`) and print an honest message on Linux/WSL. The
  Darwin strings are preserved byte-identical.

- **VoiceServer/menubar/install-menubar.sh** now has an early
  platform guard. SwiftBar and BitBar are macOS-only; on Linux/WSL
  the script prints an explanation pointing at `getNotificationCmd()`
  in `server.ts` and exits 0 cleanly so CI and cross-platform
  bootstrap flows can invoke it unconditionally. The guard is a
  single `uname -s` check; on Darwin it is a no-op and execution
  continues through the unchanged SwiftBar/BitBar detection logic
  byte-identically.

- **Packs/Utilities/INSTALL.md**'s troubleshooting section for the
  AudioEditor sub-skill only listed `brew install ffmpeg`. Added
  collapsible `<details>` blocks with macOS and Linux/WSL install
  commands. Other entries (wrangler/npm, bun/curl, fabric) are
  already cross-platform and were left alone.

## Stacking

This PR stacks on:
- #1061 — `fix(VoiceServer): cross-platform audio playback in playAudio()`
- #1072 — `fix(VoiceServer): cross-platform desktop notifications, log path, and port check`

PR #1072 introduces `lib/platform.sh` with `pai_is_darwin`, which PR
4 reuses. The diff against `main` will show both parents' content
until they merge. Rebase onto `main` after they land.

## Test plan

- [x] `bash -n` on all three modified shell scripts
- [x] `install-menubar.sh` run on real Linux host: guard fires,
      prints explanation, exits 0
- [x] `install-menubar.sh` with mocked `uname -s` = Darwin: guard is
      a no-op, script continues through to the unchanged SwiftBar /
      BitBar detection branch
- [x] `status.sh` Voice Configuration branch, all three env-file
      states (missing / placeholder / real key) on Linux — each hits
      the non-Darwin branch and prints the truthful message
[...]

PR #1072, opened 2026-04-15
fix(VoiceServer): cross-platform desktop notifications, log path, and port check

MHoroszowski:
Extends the getAudioPlayer() pattern from #1061 to cover three more
macOS-only assumptions that fail on Linux and WSL2:

1. sendNotification() hardcoded /usr/bin/osascript — ENOENT on Linux/WSL.
2. Six scripts hardcode ~/Library/Logs/pai-voice-server.log — writes to
   a non-standard location on Linux/WSL.
3. status.sh/stop.sh/uninstall.sh gate port checks on lsof — silently
   report "port not in use" when lsof is absent (common on Linux images).

Adds three TS helpers in server.ts (isWSL, getLogPath, getNotificationCmd)
that mirror the getAudioPlayer() shape, plus a shared lib/platform.sh
with pai_is_wsl / pai_log_path / pai_port_pids for the shell side.

getNotificationCmd routing:
  darwin -> /usr/bin/osascript (literal AppleScript preserved byte-identical)
  wsl2   -> wsl-notify-send if on PATH, else powershell.exe with
            BurntToast (if module importable) or bare
            [Windows.UI.Notifications.ToastNotificationManager] one-liner
  linux  -> /usr/bin/notify-send

pai_port_pids cascades lsof -> ss -> netstat.
pai_log_path uses ${XDG_DATA_HOME:-$HOME/.local/share}/pai/logs on Linux/WSL.

Verified on Ubuntu 24.04 / WSL2:
- bun bundles server.ts cleanly.
- All three port-check branches (lsof, ss-only, netstat-only) correctly
  report pai-voice's live PID against a running pai-voice.service.
- isWSL() = true, getLogPath() = ~/.local/share/pai/logs/pai-voice-server.log,
  getNotificationCmd() emits the powershell.exe BurntToast/WinRT one-liner
  which returns exit 0 and delivers a toast into the Windows notification
  queue (visually confirmed).

Darwin paths are preserved byte-identically and were NOT exercised on
hardware (author has no Mac). Please review the darwin branches
carefully — they are intentionally line-for-line equal to the
pre-refactor literals.

### Stacking on #1061

This branch is stacked on top of #1061 (cross-platform audio playback).
Because that base branch only exists on the fork, this PR targets
`main` directly; GitHub will show the combined diff until #1061 merges,
at which point the diff will collapse to only the changes introduced
here. If #1061 is merged first, no rebase is required — the merge is
already clean against main.

PR #1076, opened 2026-04-15
fix(VoiceServer): service-manager detect-and-branch for launchd+systemd

MHoroszowski:
## Summary

VoiceServer's install/start/stop/status/uninstall scripts previously
assumed macOS/launchctl exclusively. Linux and WSL2 users had no
supported path to run the voice server as a supervised service. This
PR adds a `systemd --user` branch to each script, selected at runtime
via the `pai_is_darwin` helper from `lib/platform.sh`. The Darwin
launchctl flow is preserved byte-identical.

### install.sh

- New `SYSTEMD_UNIT_*` configuration constants alongside the
  existing `PLIST_PATH`.
- "Existing installation" check branches on `pai_is_darwin`. On
  Linux/WSL it probes `systemctl --user list-unit-files` and the
  on-disk unit file, prompts the user with the same y/n reinstall
  UX as the macOS path, and on decline exits 0 without touching
  the live unit.
- Linux/WSL prechecks that `systemctl` is present and that
  `systemctl --user list-units --no-pager` is reachable. On WSL2
  it prints the `/etc/wsl.conf [boot] systemd=true` hint if the
  user session is unavailable.
- New unit generator writes
  `~/.config/systemd/user/pai-voice.service` templated on the
  reference unit shipped with PAI on WSL2:
  ```ini
  [Unit]
  Description=PAI Voice Server (ElevenLabs TTS)
  After=default.target

  [Service]
  Type=simple
  WorkingDirectory=${SCRIPT_DIR}
  ExecStart=${BUN_BIN} run server.ts
  Restart=on-failure
  RestartSec=3
  StandardOutput=append:${LOG_PATH}
  StandardError=append:${LOG_PATH}
  Environment=HOME=${HOME}
  Environment=PATH=/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:${HOME}/.bun/bin

  [Install]
  WantedBy=default.target
  ```
  `LOG_PATH` comes from `pai_log_path` (XDG on Linux, Library/Logs
  on macOS). `BUN_BIN` is resolved with `command -v bun` at
  install time. `HOME` and `PATH` are set explicitly so the child
  process can find the user's `~/.env` and runtime helpers like
  `mpg123`. The unit passes `systemd-analyze verify` with no
  warnings.
- `daemon-reload` + `enable --now` starts and persists the
  service. Failure prints the `systemctl`/`journalctl` commands
  for diagnostics.
- Post-install summary branches by service manager ("launchd" vs
  "systemd --user") and the stale "macOS Say (fallback)" voice
  string is now Darwin-only, matching the honest message #1075
  introduced elsewhere.

### start.sh / stop.sh / status.sh / uninstall.sh

Each gains the same `if pai_is_darwin; then <existing launchctl
flow> else <systemd --user equivalent> fi` wrapper. The launchctl
blocks are moved inside the Darwin branch verbatim; the systemd
blocks use `systemctl --user` {start, stop, is-active, show -p
MainPID, disable, daemon-reload} as appropriate.

## Stacking

Stacks on:
- #1061 — cross-platform audio playback in playAudio()
- #1072 — cross-platform desktop notifications, log path, port check
- #1075 — platform-aware fallback messages and macOS-only menubar guard

#1072 introduces `lib/platform.sh` with `pai_is_darwin` /
`pai_is_linux` / `pai_is_wsl`, which this PR reuses. The diff
against `main` will show all four parents' content until they
merge. Rebase after the upstream stack lands.

## Test plan

All tests run on Ubuntu 24.04 / WSL2 with a pre-existing active
`pai-voice.service` unit. Live unit mtime and service PID were
checked before, during, and after every test and remained
unchanged throughout.

- [x] `bash -n` on all five modified scripts
- [x] `systemd-analyze verify` on the generated unit (written to
      a temp HOME): clean exit, zero warnings
- [x] `install.sh` with an existing unit, 'n' answer → "Installation
      cancelled", exit 0, live unit mtime and live PID unchanged
- [x] `start.sh` with fake HOME (no unit) → "Service not installed",
      exit 1, no systemctl call
- [x] `start.sh` against the live running service → "already
      running" branch, exit 0, live PID unchanged, hint points at
      `systemctl --user restart`
- [x] `status.sh` against live running unit → reports
      `OK Service is active (PID: <real MainPID>)`
- [x] `status.sh` sliced with a bogus unit name → "Service is not
      installed" branch
- [x] `stop.sh` sliced with a bogus unit name → "not running"
      branch, no side effects
- [x] `uninstall.sh` with 'n' answer → Linux-specific confirmation
      banner ("Remove the systemd --user unit"), "Uninstall
      cancelled", live state untouched
- [x] **Darwin plist byte-identical check**: ran the unchanged
      heredoc body with identical stubs against both pre-PR-2 and
      post-PR-2 contexts → `diff` output confirms the written
      plist is byte-identical to pre-refactor
- [ ] **Darwin reviewer spot-check**: please verify on macOS that
      `launchctl load`, `launchctl unload`, `launchctl list`, the
      LaunchAgent plist content, the install success banner, and
      the menu-bar-indicator prompt all behave identically to
      pre-PR-2. Author has no Mac. Darwin branches are wrapped in
      `if pai_is_darwin; then` with the pre-refactor content
      preserved verbatim.

## Scope notes

- No parallel `install-systemd.sh` — all five scripts use inline
  detect-and-branch per the agreed scope decision (one file per
  script path, no fan-out).
- Idempotency: on a machine with an already-installed unit, the
  Linux branch prompts the user exactly like the macOS branch
  does. Declining is a clean no-op; accepting stops, disables,
  rewrites, reloads, and restarts the unit.
- **Template divergences from the reference WSL2 unit** (each a
  deliberate generalization, not a regression):
  - `WorkingDirectory` uses `${SCRIPT_DIR}` so the installer works
    from any fork checkout, not only `~/.claude/VoiceServer`.
  - `ExecStart` uses `$(command -v bun)` so non-default bun
    install locations work.
  - Log path uses `pai_log_path` (XDG) to match #1072's path
    refactor.
  - Explicit `Environment=HOME` and `Environment=PATH` so the
    service can find `~/.env` and runtime helpers regardless of
    how the user session was launched.

### DIFF
--- a/Packs/Utilities/INSTALL.md
+++ b/Packs/Utilities/INSTALL.md
@@ -373,7 +373,20 @@ Check that the sub-skill's SKILL.md exists in its directory. The routing table i
 ### Sub-skill workflow fails with missing dependency
 
 Individual sub-skills may require external tools:
-- **AudioEditor**: Install ffmpeg (`brew install ffmpeg`)
+- **AudioEditor**: Install ffmpeg
+  <details><summary>macOS</summary>
+
+  ```bash
+  brew install ffmpeg
+  ```
+  </details>
+  <details><summary>Linux / WSL</summary>
+
+  ```bash
+  sudo apt-get install -y ffmpeg    # Debian/Ubuntu/WSL
+  sudo dnf install -y ffmpeg        # Fedora
+  ```
+  </details>
 - **Cloudflare**: Install wrangler (`npm install -g wrangler`)
 - **Fabric**: Install fabric (see fabric docs)
 - **CreateCLI, Evals, PAIUpgrade, Parser**: Install bun (`curl -fsSL https://bun.sh/install | bash`)
--- a/Releases/v4.0.3/.claude/VoiceServer/install.sh
+++ b/Releases/v4.0.3/.claude/VoiceServer/install.sh
@@ -14,9 +14,11 @@ NC='\033[0m' # No Color
 
 # Configuration
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+# shellcheck source=lib/platform.sh
+. "$SCRIPT_DIR/lib/platform.sh"
 SERVICE_NAME="com.pai.voice-server"
 PLIST_PATH="$HOME/Library/LaunchAgents/${SERVICE_NAME}.plist"
-LOG_PATH="$HOME/Library/Logs/pai-voice-server.log"
+LOG_PATH="$(pai_log_path)"
 ENV_FILE="$HOME/.env"
 
 echo -e "${BLUE}=====================================================${NC}"
--- a/Releases/v4.0.3/.claude/VoiceServer/install.sh
+++ b/Releases/v4.0.3/.claude/VoiceServer/install.sh
@@ -53,6 +53,15 @@ fi
 
 # Check for ElevenLabs configuration
 echo -e "${YELLOW}> Checking ElevenLabs configuration...${NC}"
+# Platform-aware fallback message — the historical string mentioned
+# macOS 'say' regardless of host OS, which is misleading on Linux/WSL
+# where no built-in TTS fallback exists. The Darwin message is kept
+# byte-identical so the macOS flow reads the same as before.
+if pai_is_darwin; then
+    FALLBACK_NOTE="Voice server will use macOS 'say' command as fallback"
+else
+    FALLBACK_NOTE="Voice server will have no TTS fallback without an ElevenLabs API key"
+fi
 if [ -f "$ENV_FILE" ] && grep -q "ELEVENLABS_API_KEY=" "$ENV_FILE"; then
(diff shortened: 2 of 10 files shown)


danielmiessler:
Hey @MHoroszowski — Kai here, Daniel's AI assistant, doing the next pass of triage on the PAI repo.

Closing this out because v5.0.0 reorganizes voice handling under `PAI/PULSE/VoiceServer/voice.ts` with a different architecture — voice config is centralized in `PAI/USER/Config/PAI_CONFIG.yaml`, the global toggle lives at `settings.json:voiceEnabled`, and the daemon/notification topology is structured around `Pulse/pulse.ts` on port 31337. The v4-era VoiceServer this patches/replaces is gone.

**If the cross-platform / alternative-TTS / queueing concern still applies against the v5 voice layer, please open a fresh PR against the new release.** Thanks for taking the time to put this together — really appreciated.

— Kai 🤖

PR #1082, opened 2026-04-16
fix(v4.0.3): remove 16 vaporware spinner tips for non-existent skills

dmccaf:
## Summary

Removes 16 spinner tips that reference skills which don't exist:

| Tip | Status |
|-----|--------|
| `/broadcast` | Never implemented |
| `/inbox` | Never implemented |
| `/headshot` | Only example images in v2.3, no skill |
| `/hormozi` | Only a Fabric pattern, not standalone skill |
| `/javascript-analysis-v2` | Never implemented |
| `/vuln-management` | Never implemented |
| `/open-source-management` | Never implemented |
| `/clickup` | Never implemented |
| `/dotfiles` | Never implemented |
| `/homebridge-management` | Existed in v2.3 CORE/Workflows but not migrated to v4 |
| `/story-explanation` | Never implemented |
| `/speaking` | Never implemented |
| `/social-post` | Never implemented |
| `/data-wastewater-ca` | Never implemented |
| `/telegram` | Never implemented |
| `/bot-check` | Never implemented |

These tips were added aspirationally but the features were never built, causing user confusion when the commands don't work.

## Verification

- Searched entire `Packs/` directory — no matching skills
- Searched all `Releases/v4.0.x/.claude/skills/` — no matches
- Found partial implementations only in older releases (v2.3, v3.0) that weren't migrated forward

## Test plan

- [ ] Verify JSON is valid after edit
- [ ] Confirm removed tips don't appear in spinner

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1084, opened 2026-04-17
fix(v4.0.3): hide nested sub-skills from / menu via user-invocable: false

keegoid:
## Summary

Claude Code auto-indexes every `SKILL.md` in the `skills/` tree and advertises each as an invokable slash-command tip. PAI organizes skills hierarchically: top-level `SKILL.md` files are user-invokable, but **36 nested sub-skills at depth >= 2** are routed to by the parent's Workflow Routing table and are not meant to be invoked by the user directly. Without any marker distinguishing them, Claude Code advertises broken tips like `/annual-reports`, `/council`, `/osint` — they appear in the `/` menu and tip system, but nothing meaningful happens when a user tries them. ~78% false-positive rate on the primary tip affordance.

## Fix

Set `user-invocable: false` on every nested sub-skill `SKILL.md`. This is the spec-blessed frontmatter field from the [Claude Code Skills documentation](https://code.claude.com/docs/en/skills) (Frontmatter reference):

> `user-invocable` — Set to `false` to hide from the `/` menu. Use for background knowledge users shouldn't invoke directly. Default: `true`.

The flag hides the skill from the `/` menu and tip system. Claude can still route to the sub-skill via the parent's file-path reference (parent's routing table uses `Read`, not the Skill tool invocation). **No renames, no file moves, no routing changes.**

## Changes

- Add `user-invocable: false` to 36 nested sub-skill `SKILL.md` files across `ContentAnalysis`, `Investigation`, `Media`, `Scraping`, `Security`, `Thinking`, and `Utilities` parents.
- Add section **Sub-Skill Frontmatter (MANDATORY)** to `PAI/SKILLSYSTEM.md` documenting the rule, rationale, correct frontmatter shape, and validator check.
- Update `CreateSkill` `CanonicalizeSkill` and `ValidateSkill` workflows to enforce the flag on nested sub-skills as part of structural validation.

**Diff:** 39 files changed, 144 insertions(+), 0 deletions(-). Purely additive — no deletions, no renames.

## Scope

All 36 nested sub-skills in `Releases/v4.0.3/`. Older release snapshots (`v4.0.0`, `v4.0.1`, `v4.0.2`) are frozen artifacts and intentionally not touched.

## Relationship to #1082

#1082 (open) removes 16 hardcoded spinner tips that reference skill names which **never existed as files** (e.g., `/broadcast`, `/inbox`). This PR addresses a different failure mode: tips for sub-skills that **do exist** as files but should not be user-invokable. `user-invocable: false` cannot apply to non-existent files, so #1082 remains independently valid. The two PRs are complementary.

## Test plan

- [x] `find Releases/v4.0.3/.claude/skills -mindepth 3 -name SKILL.md -exec grep -L '^user-invocable: false$' {} +` returns empty (every nested SKILL.md has the flag).
- [x] `find Releases/v4.0.3/.claude/skills -mindepth 2 -maxdepth 2 -name SKILL.md -exec grep -l '^user-invocable' {} +` returns empty (no top-level skills were changed).
- [x] Parent routing tables unchanged — sub-skills still reachable via `<SubSkill>/SKILL.md` paths as before.
- [ ] Fresh Claude Code session should no longer emit tips like `/annual-reports`, `/council`, etc. for nested sub-skills (requires reviewer to test locally).

## Why this framing

This is submitted as both a convention fix (SKILLSYSTEM.md) and the full migration (all 36 files) in one PR because the per-file change is a single frontmatter line — splitting the convention and the application into separate PRs would add review overhead without meaningful isolation. If reviewer prefers the convention doc merged first, happy to split.

### DIFF
--- a/Releases/v4.0.3/.claude/PAI/SKILLSYSTEM.md
+++ b/Releases/v4.0.3/.claude/PAI/SKILLSYSTEM.md
@@ -40,6 +40,66 @@ If a skill does not follow this structure, it is not properly configured and wil
 
 ---
 
+## Sub-Skill Frontmatter (MANDATORY)
+
+**Every nested sub-skill `SKILL.md` MUST set `user-invocable: false`
+in its YAML frontmatter.**
+
+### Why
+
+Claude Code auto-indexes every `SKILL.md` file in the `skills/` tree
+and advertises each one as an invokable slash-command tip. PAI
+organizes skills hierarchically — a top-level `SKILL.md` is
+user-invokable, but sub-skills nested underneath are routed to by
+the parent's Workflow Routing table and are NOT meant to be invoked
+by the user directly. Without the `user-invocable: false` flag,
+Claude Code advertises broken tips like `/annual-reports` that do
+not work when the user tries them, eroding trust in every tip the
+system emits.
+
+The `user-invocable: false` frontmatter field is the spec-blessed
+mechanism for this. From the Claude Code Skills documentation:
+
+> `user-invocable` — Set to `false` to hide from the `/` menu. Use
+> for background knowledge users shouldn't invoke directly.
+> Default: `true`.
+
+Reference: https://code.claude.com/docs/en/skills (Frontmatter
+reference table).
+
+### Correct frontmatter for a nested sub-skill
+
+```yaml
+---
+name: AnnualReports
+description: [...]. USE WHEN annual reports, security trends, [...].
+user-invocable: false
+---
+```
+
+### Rule
+
+- **Top-level skills** (`skills/<Name>/SKILL.md`): no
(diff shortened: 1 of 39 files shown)


danielmiessler:
Hey @keegoid — Kai here, Daniel's AI assistant, doing the next pass of triage on the PAI repo.

Closing this out because this targets v4-era code paths that v5.0.0 reorganizes around — v5.0.0 already uses an underscore prefix convention (e.g., `_PAI`, `_RECON`, `_ULWORK`) for category and private skills; the `/`-menu surface uses different filtering. The specific edits don't map cleanly onto the v5 layout.

**If the underlying concern still applies in v5.0.0, please open a fresh PR against the new release.** Thanks for taking the time to put this together — really appreciated.

— Kai 🤖

PR #1090, opened 2026-04-19
fix(installer): scope chmod to shebang scripts instead of chmod -R 755

keegoid:
The configuration step runs `chmod -R 755` on the entire paiDir, flipping every .md/.json/.yaml/.txt under ~/.claude to executable. Users who manage ~/.claude under git see thousands of mode-only diffs on every upgrade.

This replaces the blanket chmod with a targeted walk that only sets +x on (a) files ending in .sh, and (b) files with a common script extension (.ts/.js/.mjs/.cjs/.py) whose first two bytes are '#!'. Everything else is left alone.

Skips .git and node_modules. All chmod errors and unreadable files are ignored (non-fatal, matching the original try/catch semantics).

Verified: bun type-check passes; unit-run against a synthetic tree flips only scripts, no data files.

### DIFF
--- a/Releases/v4.0.3/.claude/PAI-Install/engine/actions.ts
+++ b/Releases/v4.0.3/.claude/PAI-Install/engine/actions.ts
@@ -152,6 +152,71 @@ function tryExec(cmd: string, timeout = 30000): string | null {
   }
 }
 
+/**
+ * Set +x on files that actually need to be executable — shell scripts and
+ * any file whose first two bytes are `#!` (shebang). Walks paiDir recursively,
+ * skipping .git and node_modules. Returns the number of files touched.
+ *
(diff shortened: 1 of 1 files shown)


danielmiessler:
Hey @keegoid — Kai here, Daniel's AI assistant, doing the next pass of triage on the PAI repo.

Closing this out because this targets the v4.0.x install path, and v5.0.0 ships a fundamentally different one-line installer hosted at ourpai.ai/install.sh, with new headless detection, dependency probing, chmod scoping, and onboarding flow. The specific surface this addresses doesn't map onto the new installer.

**If the underlying concern still applies against the v5.0.0 installer, please reopen or open a fresh issue/PR against the new release.** Thanks for taking the time to put this together — really appreciated.

— Kai 🤖

PR #1081, opened 2026-04-16
fix(installer): make PAI-Install headless-safe on Linux

renoirb:
Two regressions prevented the installer from running on headless / SSH Linux hosts (no X11/Wayland):

1. install.sh:158 — referenced bare $DISPLAY / $WAYLAND_DISPLAY under `set -euo pipefail`. nounset aborted the script before the headless fallback could run. Fixed by using ${DISPLAY:-} / ${WAYLAND_DISPLAY:-}. Observed error: `install.sh: line 158: DISPLAY: unbound variable`.

2. main.ts:18 — mode defaulted to "gui" when --mode was omitted, so running `bun PAI-Install/main.ts` directly bypassed install.sh's headless detection and spawned Electron. Added a defaultMode() helper that returns "cli" on non-Darwin hosts without DISPLAY/WAYLAND_DISPLAY.

Also fixes: `bun PAI-Install/main.ts` no longer tries to `npm install` electron on hosts that can never run it.

PR #1092, opened 2026-04-20
README: document headless/server install path

slavik0329:
## Summary
- Adds a short "Headless / Server Install" section to the README for users installing PAI on a headless Linux server
- Documents the existing CLI-mode path (`bun run PAI-Install/main.ts --mode cli`) as an alternative to `install.sh`, which launches an Electron GUI and fails on headless boxes with `libnss3.so` errors
- No code changes — docs only

## Test plan
- [x] Rendered locally — section sits between the standard install block and the "Upgrading from a Previous Version" heading
- [x] Commands verified against `Releases/v4.0.3/.claude/PAI-Install/main.ts` (the CLI wizard path already ships in v4.0.3)

### DIFF
--- a/README.md
+++ b/README.md
@@ -356,6 +356,23 @@ cp -r .claude ~/ && cd ~/.claude && bash install.sh
 
 **After installation:** Run `source ~/.zshrc && pai` to launch PAI.
 
+### Headless / Server Install (no GUI)
+
+The default `install.sh` launches an Electron GUI, which fails on headless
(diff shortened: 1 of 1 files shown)


renoirb:
I had the same issue and I made a PR #1081 ... but the CLI option to be documented is a great idea

danielmiessler:
Hey @slavik0329, thanks for raising this, and sorry it sat for a while.

PR #1093, opened 2026-04-20
feat(statusline): add --compact mode

derekrprice:
## Summary

Adds an opt-in `--compact` flag to `Releases/v4.0.3/.claude/statusline-command.sh` that collapses the three banner sections (header, USE, PWD) into a single dense line — useful for short terminal heights or users who prefer minimal status output.

### What it does

- Parse `--compact` flag from script arguments (default: off; no behavior change without the flag).
- When `--compact` is set, emit a single dense line in place of the header + USE + PWD sections:
  ```
  CTX:XX% | 5H:XX%↻HH:MM | 7D:XX%↻HH:MM | S:$X.XX
  ```
- Pre-compute reset times and context-% scaling once, so the dense line and the full USE section share the same values (avoids drift / double work).
- Pipe all output through a new ANSI-aware `truncate_output` filter that caps each line at 72 visible chars (matching the separator width). This is needed because Claude Code's statusline subprocess receives a stale TTY width, so runtime width detection is unreliable; fixing the truncation width sidesteps that.

### Enabling it

```json
"statusLine": {
  "type": "command",
  "command": "~/.claude/statusline-command.sh --compact"
}
```

## 🙋 About the destination

I targeted `Releases/v4.0.3/.claude/statusline-command.sh` because that's the file that's actually being shipped to users in v4.0.3 (matches byte-for-byte what my local installation runs). The top-level `.claude/statusline-command.sh` at repo root is 833 lines and appears to be pre-v4.0.0 content that survived commit `e3ee9ad "Revert PAI v4.0.0 — Lean and Mean"` and hasn't been touched since — so patching it would have diverged from the actual shipped statusline.

**I guessed on the target file.** If you'd prefer this land somewhere else — top-level `.claude/`, a new `Releases/v4.0.4/`, a separate patch series, etc. — happy to rework. Just let me know.

## Test plan

- [ ] Run with `--compact`: expect single dense summary line instead of the three banner sections.
- [ ] Run without `--compact`: expect identical output to current v4.0.3 statusline (no-op path).
- [ ] Verify long lines truncate to 72 visible chars with an ellipsis and preserved ANSI reset.
- [ ] Confirm usage / reset-time values match between dense and full modes.

### DIFF
--- a/Releases/v4.0.3/.claude/statusline-command.sh
+++ b/Releases/v4.0.3/.claude/statusline-command.sh
@@ -19,6 +19,15 @@
 set -o pipefail
 
 # ─────────────────────────────────────────────────────────────────────────────
+# OPTIONS
+# ─────────────────────────────────────────────────────────────────────────────
+
+COMPACT=false
+for arg in "$@"; do
+    [ "$arg" = "--compact" ] && COMPACT=true
+done
+
+# ─────────────────────────────────────────────────────────────────────────────
 # CONFIGURATION
 # ─────────────────────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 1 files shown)


danielmiessler:
Hey @derekrprice — Kai here, Daniel's AI assistant, doing the next pass of triage on the PAI repo.

Closing this out because this targets v4-era code paths that v5.0.0 reorganizes around — `statusline-command.sh` is already a responsive 4-mode design (nano/micro/mini/normal); the compact mode it adds overlaps with the existing micro/nano modes. The specific edits don't map cleanly onto the v5 layout.

**If the underlying concern still applies in v5.0.0, please open a fresh PR against the new release.** Thanks for taking the time to put this together — really appreciated.

— Kai 🤖

PR #1101, opened 2026-04-25
feat(VoiceServer): local TTS support, voice personalities, and automatic ElevenLabs fallback

salaheldinaz:
## Summary

- **Fix .env path bug** — server loaded `~/.env` but PAI stores the key at `~/.config/PAI/.env`, causing `elevenlabs_api_key_configured` to always report `false` and silently skipping all TTS audio
- **Add local TTS engine** — uses macOS built-in `say` command (zero dependencies, no API key) via new `playLocalSpeech()` 
- **Automatic ElevenLabs fallback** — when ElevenLabs fails for any reason (402 unpaid voice, network error, missing key), server silently retries with local TTS so voice notifications always play
- **Voice personality catalogue** — curated list of 22 realistic English macOS voices with accent, gender, and personality descriptions grouped into `natural` and `classic` categories
- **`GET /voices/local` endpoint** — lists voices installed on the system with the active voice marked and instructions for switching
- **Improved `/health` response** — now exposes `tts_provider`, `local_voice`, `local_tts_available`; renames `api_key_configured` → `elevenlabs_api_key_configured`

## Configuration

No breaking changes. Existing installs behave identically. To customise local TTS add to `~/.claude/settings.json`:

```json
"voiceServer": {
  "tts_provider": "local",
  "local_voice": "Daniel"
}
```

`tts_provider` options: `"elevenlabs"` (default — ElevenLabs with local fallback) or `"local"` (always local, no API call).

Browse available voices on your system:
```bash
curl http://localhost:8888/voices/local
```

## Test plan

- [ ] `/health` returns `local_tts_available: true` and correct `tts_provider`
- [ ] `GET /voices/local` returns catalogue with active voice marked
- [ ] ElevenLabs 402 / missing key → local TTS plays, response still `200`
- [ ] `tts_provider: "local"` → local TTS used directly, no ElevenLabs call
- [ ] `tts_provider: "elevenlabs"` with valid paid key → ElevenLabs used, unchanged behaviour
- [ ] `.env` loaded correctly from `~/.config/PAI/.env`

### DIFF
--- a/Releases/v4.0.1/.claude/VoiceServer/server.ts
+++ b/Releases/v4.0.1/.claude/VoiceServer/server.ts
@@ -1,6 +1,6 @@
 #!/usr/bin/env bun
 /**
- * Voice Server - Personal AI Voice notification server with pluggable TTS providers (ElevenLabs, kokoro-fastapi)
+ * Voice Server - Personal AI Voice notification server with pluggable TTS providers (ElevenLabs, kokoro-fastapi, local)
  *
  * Architecture: Pure pass-through. All voice config comes from settings.json.
  * The server has zero hardcoded voice parameters.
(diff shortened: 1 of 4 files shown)


danielmiessler:
Hey @salaheldinaz, thanks for raising this, and sorry it sat for a while.

PR #1125, opened 2026-05-01
fix: add missing Pulse daemon dependencies (yaml, minisearch, smol-toml)

brycemagera:
## Problem

Three packages are imported by Pulse daemon modules but were never declared in any `package.json`:

| Package | Imported by | Effect when missing |
|---------|-------------|---------------------|
| `yaml` | `modules/observability.ts` | Observability module fails to load; all Pulse routes return 404 |
| `minisearch` | `modules/wiki.ts` | Wiki module fails to load; `/knowledge`, `/docs`, `/skills`, `/hooks` pages return 404 |
| `smol-toml` | `pulse.ts`, `lib.ts` | PULSE.toml parsing fails on cold start |

On every cold start, Bun cannot resolve these packages, the affected modules fail silently, and the Pulse dashboard returns `404` on the routes they serve. The dashboard only works if the packages have been added manually after install.

## Fix

Create `PAI/PULSE/package.json` with all three runtime dependencies and a corresponding `bun.lock`. Update `.gitignore` to carve out `PAI/PULSE/bun.lock` alongside the existing `Observability/bun.lock` exception.

The deps belong in `PULSE/package.json` (the daemon root), not in `PULSE/Observability/package.json` (the Next.js static export) — the modules that import them are Pulse server-side code, not frontend components.

## Reproduction

```
cd Releases/v5.0.0/.claude/PAI/PULSE
bun install   # fresh install — minisearch, smol-toml, yaml not resolved
bun pulse.ts  # wiki + observability modules fail silently
curl http://localhost:31337/api/wiki/skills  # → 404 Not found
curl http://localhost:31337/api/wiki         # → 404 Not found
```

### DIFF
--- a/Releases/v5.0.0/.claude/.gitignore
+++ b/Releases/v5.0.0/.claude/.gitignore
@@ -186,6 +186,9 @@ Plugins/marketplaces/.chrome-profile/
 # Pulse Observability lockfile — ship for reproducible installs
 !PAI/PULSE/Observability/bun.lock
 
+# Pulse daemon lockfile — ship for reproducible installs
+!PAI/PULSE/bun.lock
+
 # Python virtual environments
 .venv/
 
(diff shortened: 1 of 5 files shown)


danielmiessler:
Hey @brycemagera, thank you for the PR and the work that went into it.

PR #1126, opened 2026-05-01
Fix/linux support

dmcgoldrd:
### What this is
A **bug fix** PR — closes the gap that prevented `Releases/v5.0.0/.claude/` from installing end-to-end on Linux. macOS path is unchanged.

Every edit is scoped to `Releases/v5.0.0/.claude/` — no impact on v4.x or anywhere else in the tree.

### What it changes (7 commits)

- **Add runtime `package.json`** at `.claude/` root so `bun install` resolves Pulse runtime deps on a fresh extract.
- **Resolve `bash` absolutely in `PULSE/lib.ts spawnScript()`** so subprocesses don't depend on PATH ordering.
- **Add `PULSE/pai-pulse.service`** — systemd user-unit template, companion to the existing `com.pai.pulse.plist`. Same `__HOME__` / `__BUN_PATH__` placeholder pattern.
- **Make `PULSE/manage.sh` OS-aware** — `start`/`stop`/`install`/`uninstall` route to launchd on macOS, `systemctl --user` on Linux. Install kills any stale `bun.*pulse.ts`, renders the unit, runs `daemon-reload` + `enable --now`, and verifies `:31337` binds within 10s — failing loud if not.
- **Casing fix `Pulse/` → `PULSE/`** across 9 TypeScript files. macOS didn't care, Linux did.
- **Remove `.cursor/rules/*.mdc`** IDE artifacts from the public release.
- **Make installer wizard messaging + `engine/validate.ts` OS-aware** — neutral "system service" wording instead of "launchd"; post-install validator branches on `platform()` and checks the correct artifact (launchd plist on macOS, systemd unit on Linux) so Linux installs no longer false-fail with `Pulse launchd agent: Not installed`.

### Tested on a fresh-ish system

Arch Linux x86_64, kernel 6.19.x, bun 1.3.6 via mise, fresh `~/.claude/` with prior install backed up to `~/.claude.backup-<timestamp>/`.

**1. `bash ~/.claude/PAI/PULSE/manage.sh install`** — the OS-aware install path the wizard calls:

```
$ bash ~/.claude/PAI/PULSE/manage.sh install
Created symlink '~/.config/systemd/user/default.target.wants/pai-pulse.service' →
                '~/.config/systemd/user/pai-pulse.service'.
Hint: 'sudo loginctl enable-linger <user>' to keep Pulse running after logout.
PAI Pulse installed and verified on port 31337 (bun: ~/.local/share/mise/installs/bun/1.3.5/bin/bun)
$ echo $?
0
```

**2. Post-install probes:**

```
$ systemctl --user is-enabled pai-pulse.service
enabled
$ systemctl --user is-active pai-pulse.service
active
$ ss -tlnp | grep :31337
LISTEN 0 512 127.0.0.1:31337 *:* users:(("bun",pid=<pid>,fd=257))
$ curl -s -X POST http://localhost:31337/notify -H "Content-Type: application/json" \
       -d '{"message":"smoke","voice_enabled":false}'
{"status":"success","message":"Notification sent"}
```

Any stale pre-install pulse process from the backup tree was correctly killed by the install's `pkill -9 -f "bun.*pulse.ts"` step.

**3. Validate.ts spot-check** (the `platform()` branch from the last commit):

```js
{
  platform: "linux",
  isLinux: true,
  serviceUnit: "~/.config/systemd/user/pai-pulse.service",
  installed: true,
  label: "Pulse systemd unit"
}
```

Confirms the post-install validator now passes on Linux instead of false-failing on a missing launchd plist.

**4. Typecheck:** `bun tsc --noEmit -p PAI-Install/tsconfig.json` — no errors in `engine/actions.ts` or `engine/validate.ts`.

### Before / after example

Linux user runs `curl -sSL https://ourpai.ai/install.sh | bash`, picks "Yes" at the Pulse install prompt, then runs the validator:

| Probe | Before this PR | After this PR |
|---|---|---|
| `~/.config/systemd/user/pai-pulse.service` rendered | n/a (manage.sh had no Linux path) | ✓ |
| `systemctl --user is-active` | `inactive` (no unit) | `active` |
| Listener on `:31337` | nothing | `bun` (PID matches `state/pulse.pid`) |
| Validator output | `Pulse launchd agent: Not installed` (false-negative on Linux) | `Pulse systemd unit: Installed at ~/.config/systemd/user/pai-pulse.service` |
| `curl POST /notify` | connection refused | HTTP 200 |

### macOS impact

None. All OS branches default to the macOS path; `manage.sh` and `validate.ts` route via `case "$OS"` / `platform() === "linux"` so existing macOS behavior is byte-identical.

### Known follow-ups (intentionally not in this PR)

- `engine/actions.ts:1859` descriptive emit message still says "Installing it as a launchd agent makes it auto-start..." — clean cosmetic follow-up to neutralize.
- The Pulse menu bar app step (`engine/actions.ts:1874+`) is structurally macOS-only. No cross-platform menu bar work in this PR.


danielmiessler:
Hey @dmcgoldrd, thanks for raising this, and sorry it sat for a while.

PR #1127, opened 2026-05-01
fix(telos): correct GenerateTelosSummary parser to match shipped templates (#1113, #1115)

### LINKED ISSUE #1113: TELOS template/parser mismatch breaks `PRINCIPAL_TELOS.md` rendering
## Summary

The bootstrap TELOS templates and the `GenerateTelosSummary.ts` parsers are out of sync in two ways. After running the renderer on a fresh PAI v5.0.0 install, `PRINCIPAL_TELOS.md` contains:

1. **Stray `**` characters** in every Missions/Goals/Problems/Challenges/Narratives entry
2. **An empty `## Strategies` section** — the shipped STRATEGIES template never renders at all

Both symptoms come from a single root cause: the templates and parsers were written against different bullet/header conventions and never reconciled.

## Reproduction

1. Fresh PAI v5.0.0 install (templates as shipped under `Releases/v5.0.0/.claude/PAI/USER/TELOS/`).
2. Run:
   ```bash
   bun ~/.claude/PAI/TOOLS/GenerateTelosSummary.ts
   ```
3. Open the generated `PAI/USER/TELOS/PRINCIPAL_TELOS.md`.

**Actual:**
```
## Missions
- **M0**: ** (sample) Help one million people make better decisions...
- **M1**: ** (sample) Build infrastructure that lets individuals own...

## Strategies

## Active Narratives
...
```

**Expected:**
```
## Missions
- **M0**: (sample) Help one million people make better decisions...
- **M1**: (sample) Build infrastructure that lets individuals own...

## Strategies
- **S0**: (sample) Ship the crappy version, then iterate in public...
- **S1**: (sample) Write before building...
- **S2**: (sample) Default to open...
```

Same `** ` prefix artifact appears in Goals, Problems, Challenges, and Narratives sections. The Strategies section is silently empty.

## Root cause #1 — `**ID:**` vs `**ID**:` (templates)

The templates use `**ID:**` (colon **inside** the bold span):

```markdown
- **M0:** (sample) Help one million people make better decisions...
```

The parser at `Releases/v5.0.0/.claude/PAI/TOOLS/GenerateTelosSummary.ts:53` expects `**ID**:` (colon **outside** the bold span):

```ts
const match = line.match(/^-\s+\*?\*?(\w+)\*?\*?:\s*(.+)/);
```

Walking the regex against `- **M0:** text`:

| token | matches |
|---|---|
| `^-\s+` | `- ` |
| `\*?\*?` | `**` |
| `(\w+)` | `M0` |
| `\*?\*?` | _empty_ (next char is `:`) |
| `:` | `:` |
| `\s*` | _empty_ |
| `(.+)` | `** text` ← garbage prefix |

So the parser silently captures the trailing `**` as part of the value. Affects MISSION, GOALS, PROBLEMS, STRATEGIES, CHALLENGES, NARRATIVES.

## Root cause #2 — parsers read different formats per file

Even with `Root cause #1` fixed, the Strategies section stays empty. `parseStrategies` reads only H2/H3 headers:

```ts
// GenerateTelosSummary.ts:131
const headers = [...content.matchAll(/^#{2,3}\s+(S\d+):\s*(.+?)(?:\s*\(.*\))?\s*$/gm)];
```

But the shipped `STRATEGIES.md` template uses bullets (`- **S0:** text`), not headers. So the parser ignores the template entirely. `parseProblems` has the inverse problem — it reads headers first and only falls back to bullets if headers find nothing, an unnecessary fork.

The other ID-bearing parsers (`parseMissions`, `parseGoals`, `parseNarratives`, `parseChallenges`) all read bullets via `parseItems`. The split between bullet-based and header-based parsing across files is the second half of the bug.

## Affected files

**Templates (Root cause #1):**

- `Releases/v5.0.0/.claude/PAI/USER/TELOS/MISSION.md` (M0–M2)
- `Releases/v5.0.0/.claude/PAI/USER/TELOS/GOALS.md` (G0–G2)
- `Releases/v5.0.0/.claude/PAI/USER/TELOS/PROBLEMS.md` (P0–P2)
- `Releases/v5.0.0/.claude/PAI/USER/TELOS/STRATEGIES.md` (S0–S2)
- `Releases/v5.0.0/.claude/PAI/USER/TELOS/CHALLENGES.md` (C0–C2)
- `Releases/v5.0.0/.claude/PAI/USER/TELOS/NARRATIVES.md` (N0–N2)

**Parser (Root cause #2):**

- `Releases/v5.0.0/.claude/PAI/TOOLS/GenerateTelosSummary.ts` — `parseStrategies`, `parseProblems`

`WISDOM.md`, `BELIEFS.md`, and `BOOKS.md` use other formatting and are not affected.

## Why this matters

`PRINCIPAL_TELOS.md` is force-loaded into every Claude Code session via `CLAUDE.md`'s `@`-import block. Garbage prefixes propagate into the model's context on every startup, and the missing Strategies section means the DA never sees the user's playbook even after they've filled it in. Most users won't notice — broken output looks like a markdown rendering quirk or a "you haven't filled this in yet" state.

New users running `/interview` for the first time see broken output even when they did everything right.

## Proposed fix (PR #1114 follows)

1. Templates: `**ID:**` → `**ID**:` across 6 files (18 lines).
2. Parsers: `parseStrategies` and `parseProblems` standardize on `parseItems` with an ID-prefix filter, mirroring `parseMissions`/`parseGoals`/`parseNarratives`/`parseChallenges`. Header-based parsing is dropped — single bullet convention across all ID-bearing TELOS files.

### Breaking change note

Users whose existing `STRATEGIES.md` or `PROBLEMS.md` files use H2/H3 headers (`## S0: title` / `### P0: title`) will stop rendering after the parser change. Migrate to bullet format `- **S0**: title` to match. The shipped v5.0.0 templates already use bullets, so users who never customized are unaffected.

## Verification

After applying the PR, running the renderer against the as-shipped sample templates produces clean, complete output across all six sections (Missions, Goals, Problems, Strategies, Challenges, Narratives) — no `**` artifact, no silent drops.

### LINKED ISSUE #1115: GenerateTelosSummary: G2 misclassified as deferred; `Deferred / Ongoing` and `Completed This Year` sections silently dropped
## Symptom

Run the renderer against the as-shipped v5.0.0 sample template:

```bash
bun ~/.claude/PAI/TOOLS/GenerateTelosSummary.ts
```

In the generated `PRINCIPAL_TELOS.md`:

- `G2` (under `## Active` in the template) → output as **Deferred**
- All `## Deferred / Ongoing` items (the no-ID bullets like `- (sample) Maintain a 3x/week strength routine ...`) → **silently dropped**
- All `## Completed This Year` items (also no-ID bullets) → **silently dropped**

So a fresh user running the bootstrap loses two of the three sections from the summary entirely, and the one that survives mis-classifies the third Active goal.

## Root cause

In `Releases/v5.0.0/.claude/PAI/TOOLS/GenerateTelosSummary.ts`:

1. **L53 regex** requires an ID:

   ```ts
   const match = line.match(/^-\s+\*?\*?(\w+)\*?\*?:\s*(.+)/);
   ```

   Plain bullets without `**ID**:` (the form used by the `Deferred / Ongoing` and `Completed This Year` sections in the shipped template) don't match — they're dropped before classification.

2. **L73-94 classifier** uses an ID-number heuristic and ignores section headers:

   ```ts
   // Goals with IDs G9+ are 2026 goals based on the file structure
   if (num >= 9 || [0, 1].includes(num)) {
     active.push(...)
   } else {
     deferred.push(...)
   }
   ```

   `parseItems` is line-based with no awareness of `## Active`, `## Deferred / Ongoing`, or `## Completed This Year` headers. So `G2` from the template's `## Active` block lands in `deferred` purely because `2` is not in `[0, 1]` and not `>= 9`.

## Open question

The classifier in [`Releases/v5.0.0/.claude/PAI/TOOLS/GenerateTelosSummary.ts`](https://github.com/danielmiessler/Personal_AI_Infrastructure/blob/main/Releases/v5.0.0/.claude/PAI/TOOLS/GenerateTelosSummary.ts#L73-L94) (lines 73-94):

```ts
// Goals with IDs G9+ are 2026 goals based on the file structure  (L77)
if (num >= 9 || [0, 1].includes(num)) {
  active.push(...)
} else {
  deferred.push(...)
}
```

looks like it encodes a personal goal-numbering convention — `G0`/`G1` for foundational/long-term, `G9+` for current-year milestones, the mid-range for deferred. The comment "based on the file structure" suggests this came from a specific author's `GOALS.md`, not a framework convention. The shipped template at [`Releases/v5.0.0/.claude/PAI/USER/TELOS/GOALS.md`](https://github.com/danielmiessler/Personal_AI_Infrastructure/blob/main/Releases/v5.0.0/.claude/PAI/USER/TELOS/GOALS.md) doesn't follow it (uses `G0`/`G1`/`G2` under `## Active`).

Two questions:

1. **Is this convention intentional and meant for all PAI users?** If yes, it should be documented (template comment, README, `/interview` prompts) so users know to number active goals starting at `G9`.
2. **Or should `parseGoals` honor the section headers (`## Active` / `## Deferred / Ongoing` / `## Completed This Year`) instead of guessing from IDs?** That makes the convention self-documenting in the file itself.

Either direction works — happy to bundle a PR once the call is made.

jmmarkiewicz:
Bundles two open TELOS-renderer issues — both root-cause in `Releases/v5.0.0/.claude/PAI/TOOLS/GenerateTelosSummary.ts`:

- Fixes #1113 — template/parser mismatch (stray `**` artifacts; empty `## Strategies` section)
- Fixes #1115 — `G2` misclassified as deferred; `## Deferred / Ongoing` and `## Completed This Year` sections silently dropped

### `parseItems()` regex (L53)
Accept colon either inside or outside the `**` bold span — covers all four bullet styles:
```
- **M0**: text       (current expectation)
- **M0:** text       (what shipped templates use)
- M0: text
- M0:** text
```
Old: `/^-\s+\*?\*?(\w+)\*?\*?:\s*(.+)/` — captured trailing `**` as part of value, producing the `**M0**: ** ...` artifact.
New: `/^-\s+\*?\*?(\w+)\*?\*?:\*?\*?\s*(.+)/` — consumes the closing `**` before capturing.

### `parseStrategies()` rewrite
Use the shared bullet parser instead of `/^#{2,3}\s+(S\d+):/`. The shipped STRATEGIES.md stores strategies as bullets under `## Active` / `## Anti-Strategies`, not as `## S0:` headers. Filter on `/^[SA]\d+$/` so both Sn and An (anti-strategy) entries land in the rendered output.

### `parseGoals()` rewrite — section-walker classifier
Replaces the ID-number heuristic (`num >= 9 || [0, 1].includes(num)`) — which encoded a personal goal-numbering convention the shipped template doesn't follow — with section-header awareness. The renderer now classifies each bullet by which `## Active` / `## Deferred / Ongoing` / `## Completed This Year` block it appears under. This is the self-documenting choice the issue body's "open question" identified.

Two subtleties:
- **Only `##` switches section.** Sub-headings (`### Work`, `### Home`, `### Tech / PAI`) inherit the parent section. Treating every `##+` as a section boundary silently drops any goals the user organizes by sub-area.
- **No-ID bullets in Deferred/Ongoing/Completed are rendered as a compact `_Ongoing: name1, name2, ..._` inline line** instead of being dropped. Bold name markers are stripped for the inline list.

### Template-placeholder skip
`*(none recorded yet — update as items ship)*` and similar `*(none/tbd/placeholder ...)*` patterns are now silently skipped so a fresh install shows nothing for empty buckets rather than a literal placeholder bullet.

### Render block in `generate()`
Updated to the new `{ active, deferredIds, deferredPlain }` parseGoals return shape.

## Verification

Tested against a populated GOALS.md that exercises every code path:
- 8 active goals (G0–G7) split across 3 sub-sections (`### Work`, `### Home`, `### Tech / PAI`) under `## Active`
- 4 plain-bullet entries under `## Deferred / Ongoing` (no IDs, with `**Bold Name** — explanation` form)
- 1 placeholder bullet under `## Completed This Year`
- Populated STRATEGIES.md with `- **S0:**…` through `- **S5:**…` plus `- **A0:**…`

Before this patch:
- `## Active Goals (2026)` showed only G0/G1, then `_Deferred (full text in TELOS/GOALS.md): G2, G3, G4, G5, G6, G7_`
- `## Strategies` was empty
- `## Deferred / Ongoing` plain bullets were silently dropped
- Every entry had a stray `** ` prefix in the captured value

After:
- All 8 active goals appear under `## Active Goals (2026)`
- Strategies block shows S0–S5 + A0 (7 entries)
- A single `_Ongoing: PAI Infrastructure Maturation, OpenClaw Ecosystem, House Refresh 2026, CashX_` line appears below Active Goals
- Stray-`**` regression check finds 0 buggy entries (legitimate bold-name strategies/challenges preserved)
- `*(none recorded yet…)*` placeholder is silently skipped

## Notes

The render-time decision to compress no-ID Deferred entries to a single inline line (rather than full bullets) preserves the original design intent ("they're not active and don't need full bullets") while fixing the silent-drop bug — boot-context summaries stay compact.

### DIFF
--- a/Releases/v5.0.0/.claude/PAI/TOOLS/GenerateTelosSummary.ts
+++ b/Releases/v5.0.0/.claude/PAI/TOOLS/GenerateTelosSummary.ts
@@ -49,8 +49,10 @@ function parseItems(content: string): ParsedItem[] {
   const lines = content.split('\n');
 
   for (const line of lines) {
-    // Match "- **M0**: text" or "- M0: text" or "- **G0**: text" patterns
-    const match = line.match(/^-\s+\*?\*?(\w+)\*?\*?:\s*(.+)/);
+    // Match all four bullet styles: "- **M0**: text", "- **M0:** text",
+    // "- M0: text", and "- M0:** text". Optional ** before AND after the
+    // colon handles the template/parser drift identified in upstream #1113.
+    const match = line.match(/^-\s+\*?\*?(\w+)\*?\*?:\*?\*?\s*(.+)/);
     if (match) {
       items.push({ id: match[1], text: match[2].trim() });
     }
@@ -68,29 +70,80 @@ function parseMissions(): string[] {
 }
 
 /**
- * Parse goals from GOALS.md, separating 2026 goals from older ones
+ * Parse goals from GOALS.md by walking section headers, not by ID heuristic.
+ * Section headers (## Active / ## Deferred / ## Ongoing / ## Completed) drive
+ * classification — fixes upstream #1115 where G2 was misclassified as deferred
+ * because the old code used `num >= 9 || [0,1].includes(num)`.
  */
-function parseGoals(): { active: string[]; deferred: string[] } {
+function parseGoals(): {
+  active: string[];
+  deferredIds: string[];
+  deferredPlain: string[];
+} {
   const content = readTelosFile('GOALS.md');
-  const items = parseItems(content);
-
-  // Goals with IDs G9+ are 2026 goals based on the file structure
   const active: string[] = [];
-  const deferred: string[] = [];
+  const deferredIds: string[] = [];
+  const deferredPlain: string[] = [];
 
-  for (const item of items) {
-    const num = parseInt(item.id.replace(/\D/g, ''), 10);
-    // Split on " — " (em-dash with spaces) or sentence-ending period (not in URLs)
-    const firstSentence = item.text.split(/\s—\s|(?<!\w\.\w)(?<=\w)\.\s/)[0].trim();
+  type Section = 'active' | 'deferred' | 'completed' | 'other';
+  let currentSection: Section = 'other';
 
-    if (num >= 9 || [0, 1].includes(num)) {
-      active.push(`- **${item.id}**: ${truncate(firstSentence, 70)}`);
-    } else {
-      deferred.push(`- **${item.id}**: ${truncate(firstSentence, 50)}`);
+  for (const rawLine of content.split('\n')) {
+    const line = rawLine.trimEnd();
+
+    // Only top-level (##) headings switch the section. Sub-headings like
+    // `### Work`, `### Home`, `### Tech / PAI` are children of the parent
+    // section and must inherit it (otherwise G0-G7 under `## Active` →
+    // `### Work` get classified as 'other' and silently dropped).
+    const headerMatch = line.match(/^##\s+(.+?)\s*$/);
+    if (headerMatch) {
+      const h = headerMatch[1].toLowerCase();
+      if (h.startsWith('active')) currentSection = 'active';
+      else if (h.startsWith('deferred') || h.startsWith('ongoing')) currentSection = 'deferred';
+      else if (h.startsWith('completed')) currentSection = 'completed';
+      else currentSection = 'other';
+      continue;
+    }
+
+    if (currentSection === 'other') continue;
+
+    const idMatch = line.match(/^-\s+\*?\*?(\w+)\*?\*?:\*?\*?\s*(.+)/);
+    if (idMatch) {
+      const id = idMatch[1];
+      const text = idMatch[2].trim();
+      const firstSentence = text.split(/\s—\s|(?<!\w\.\w)(?<=\w)\.\s/)[0].trim();
+
+      if (currentSection === 'active') {
+        active.push(`- **${id}**: ${truncate(firstSentence, 70)}`);
+      } else {
+        // deferred or completed → both go to deferredIds (the inline ID-list line)
+        deferredIds.push(id);
+      }
+      continue;
+    }
+
+    // Plain bullet without ID — used by Deferred / Ongoing / Completed items
+    // in the shipped template. These were silently dropped in upstream #1115.
+    const plainMatch = line.match(/^-\s+(.+)/);
+    if (plainMatch) {
+      const text = plainMatch[1].trim();
+      const firstSentence = text.split(/\s—\s|(?<!\w\.\w)(?<=\w)\.\s/)[0].trim();
+
+      // Skip template placeholders like `*(none recorded yet — update as items ship)*`
+      if (/^\*?\(.*(none|tbd|placeholder)/i.test(firstSentence)) continue;
+
+      if (currentSection === 'deferred' || currentSection === 'completed') {
+        // Strip surrounding **bold** so we keep just the name for the inline list.
+        const bare = firstSentence.replace(/^\*\*(.+?)\*\*.*$/, '$1');
+        deferredPlain.push(truncate(bare, 60));
+      } else if (currentSection === 'active') {
+        // Active without ID — uncommon but render directly so it isn't lost.
+        active.push(`- ${truncate(firstSentence, 70)}`);
+      }
     }
   }
 
-  return { active, deferred };
+  return { active, deferredIds, deferredPlain };
 }
 
 /**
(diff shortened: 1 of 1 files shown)


danielmiessler:
Thank you for this, @jmmarkiewicz — and sorry it sat open so long.

The codebase changed a lot since you opened it (the PAI → LifeOS rename, path restructures, and a shift to a generated single-skill public artifact), and this one no longer applies to the current tree: thanks for catching the TELOS parser/template drift — that whole tool was since rewritten to parse H3 headings plus an ID-less prose fallback with fail-loud per-section guards, so these specific fixes no longer apply to current source.

PR #1145, opened 2026-05-02
Fix broken cursor-rule symlinks in v5.0.0 (crashes Pulse on Linux)

gl0bal01:
## Summary

Three `.cursor/rules/use-bun-instead-of-node-vite-npm-pnpm.mdc` files in `Releases/v5.0.0/` are symlinks to `../../CLAUDE.md`. Two problems:

1. The target is wrong — these should be Cursor IDE \`.mdc\` rule files (with cursor-rule frontmatter), not Claude system-prompt markdown. The actual content existed as regular files in \`Releases/v2.3/\`.
2. After bundle extraction during install, the symlinks become self-referencing absolute paths on Linux. Pulse's file watcher hits this on startup and crashes with \`ELOOP: too many symbolic links\` on every restart attempt.

## Why macOS doesn't see this

HFS+/APFS handles the bad symlink resolution silently. Linux ext4/btrfs/xfs does not.

## Reproduction

```bash
# Fresh Ubuntu 24.04 (or any Linux VPS)
useradd --system --create-home --shell /bin/bash pai
sudo -iu pai
curl -fsSL https://ourpai.ai/install.sh | bash
# select default install
systemctl --user start pulse  # or whatever launches it
journalctl --user -u pulse -f
# → repeating: error: ELOOP: too many symbolic links encountered, open
#               '/home/pai/.claude/PAI/PULSE/Observability/.cursor/rules/use-bun-instead-of-node-vite-npm-pnpm.mdc'
```

## Fix

Replace the 3 broken symlinks with the proper \`.mdc\` rule content (restored from \`Releases/v2.3/\` where it last existed as a regular file). Identical content used in all 3 locations since they were intended to share the same rule.

## Affected paths

- \`Releases/v5.0.0/.claude/PAI/PULSE/Observability/.cursor/rules/use-bun-instead-of-node-vite-npm-pnpm.mdc\`
- \`Releases/v5.0.0/.claude/skills/Art/Tools/.cursor/rules/use-bun-instead-of-node-vite-npm-pnpm.mdc\`
- \`Releases/v5.0.0/.claude/skills/Prompting/Templates/Tools/.cursor/rules/use-bun-instead-of-node-vite-npm-pnpm.mdc\`

## Notes

- This PR fixes only the static release directory. The underlying installer logic that converts relative symlinks to broken absolute paths during extraction may still need investigation, but inlining the rule content makes the fix robust regardless.
- A second case-sensitivity bug (PAI/Pulse vs PAI/PULSE, PAI/Tools vs PAI/TOOLS) also breaks the v5.0.0 release on Linux. I'll file that separately as it requires a project-wide naming decision.

## Test plan

- [x] Inspect each replaced file: contains valid Cursor IDE \`.mdc\` rule with frontmatter
- [x] \`file <path>\` reports regular text file, not symlink
- [ ] Run install on fresh Linux VPS, confirm Pulse starts without ELOOP

### DIFF
--- a/Releases/v5.0.0/.claude/PAI/PULSE/Observability/.cursor/rules/use-bun-instead-of-node-vite-npm-pnpm.mdc
+++ b/Releases/v5.0.0/.claude/PAI/PULSE/Observability/.cursor/rules/use-bun-instead-of-node-vite-npm-pnpm.mdc
@@ -1,1 +1,111 @@
-../../CLAUDE.md
+---
+description: Use Bun instead of Node.js, npm, pnpm, or vite.
+globs: "*.ts, *.tsx, *.html, *.css, *.js, *.jsx, package.json"
+alwaysApply: false
+---
+
+Default to using Bun instead of Node.js.
+
+- Use `bun <file>` instead of `node <file>` or `ts-node <file>`
+- Use `bun test` instead of `jest` or `vitest`
+- Use `bun build <file.html|file.ts|file.css>` instead of `webpack` or `esbuild`
+- Use `bun install` instead of `npm install` or `yarn install` or `pnpm install`
+- Use `bun run <script>` instead of `npm run <script>` or `yarn run <script>` or `pnpm run <script>`
+- Bun automatically loads .env, so don't use dotenv.
+
+## APIs
+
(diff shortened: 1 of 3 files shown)


danielmiessler:
Hey @gl0bal01, thank you for the PR and the work that went into it.

PR #1116, opened 2026-05-01
feat(pulse): WebSocket browser audio streaming

Alessandro-Improta:
## Commits

| SHA | Description |
|-----|-------------|
| `4470f21` | feat(pulse): WebSocket browser audio streaming for headless Linux |
| `b988d75` | fix(pulse): unlock AudioContext on first page interaction, not toggle-only |

---

## What

**WebSocket browser audio streaming** — Pulse streams voice notifications to connected browser tabs via a `/ws/audio` WebSocket. On headless Linux where `afplay` doesn't exist and no native audio device is present, the browser becomes the speaker. macOS `afplay` path is fully preserved.

## Why

On headless Linux, ElevenLabs TTS generates audio bytes but has nowhere to send them — no `afplay`, no ALSA device in a typical SSH-only server setup. Users accessing Pulse from a browser over LAN or SSH tunnel had no way to hear voice notifications. The fix: stream the raw MP3 bytes to all connected browser tabs via WebSocket; the browser decodes and plays with the Web Audio API. The toggle in the Pulse header lets users opt in once; localStorage persists the preference across reloads.

### `PAI/PULSE/pulse.ts` — WebSocket audio endpoint

- `Bun.serve()` `websocket:` handler with `open`/`close`/`message` callbacks; connected clients tracked in a module-level `Set`.
- `/ws/audio` route: responds to `Upgrade: websocket`; all other requests fall through to the existing HTTP router.
- `setAudioBroadcaster(fn)` injection point: `pulse.ts` wires in the broadcaster after `startVoice()` returns, avoiding a circular `voice ↔ pulse` import dependency.
- Broadcaster sends raw MP3 `ArrayBuffer` as a binary WS frame to every connected client; fire-and-forget (a slow/disconnected client cannot block `sendNotification()`).
- `/voice/health` response includes `ws_audio_clients` count.

### `PAI/PULSE/VoiceServer/voice.ts` — broadcaster + Linux audio fallback

- `setAudioBroadcaster(fn)` / `setWsAudioClientCount(n)` exports for injection from `pulse.ts`.
- `playAudio()` branches on `process.platform`:
  - `darwin` → `afplay` (unchanged)
  - `linux` → tries `paplay` → `aplay` → `mpg123 -q` in order; logs `debug` if none found, never throws.
- After the native-playback block, always calls `broadcastAudio(buffer)` regardless of native result.
- 402/`payment_required` ElevenLabs errors downgraded from `error` → `warn`.

[...]

PR #1148, opened 2026-05-02
feat(pulse): add Linux systemd support to init scripts

m8ryx:
## Problem

  `manage.sh install` fails on Linux with `launchctl: command not found`.
  The PULSE init scripts were written for macOS only — they use `launchctl`,
  launchd plist XML, and `~/Library/LaunchAgents/`. `start-pulse.sh` also
  unconditionally calls `security unlock-keychain`, another macOS-only binary.
  Pulse cannot be installed or auto-started on any Linux machine.

## Changes

**`manage.sh`** — detects OS via `uname -s` and branches per sub-command:
  - `install`: writes a systemd user-scope unit, calls `daemon-reload` /
    `enable` / `start`, then verifies port 31337 binds (same 10s probe as macOS)
  - `start` / `stop` / `status` / `uninstall`: use `systemctl --user` on Linux,
    unchanged launchctl path on macOS
  - Adds `loginctl enable-linger` so the service survives logout and auto-starts
    at boot on Linux
  - All macOS launchctl/plist code is preserved unchanged in `else` branches

**`com.pai.pulse.service`** — new systemd user-scope unit template using the
  same `__HOME__` / `__BUN_PATH__` placeholder pattern as the existing plist.
  Logs to `PULSE/logs/` to match macOS behavior. `TimeoutStopSec=5` prevents
  `systemctl stop` from hanging while waiting for bun to respond to SIGTERM.

**`start-pulse.sh`** — guards the keychain unlock behind a `uname -s` Darwin
  check so it no-ops on Linux.

**`package.json`** — PULSE had no `package.json`, so `smol-toml`, `grammy`,
  `jose`, and `minisearch` were unresolvable at runtime on a fresh install.

## Testing

  - Tested on Pop!_OS 24.04 LTS (systemd 255, bun 1.3.3)
  - `manage.sh install` → service enabled, port 31337 verified, `/healthz` → 200
  - macOS launchctl path verified by code inspection; not exercised on Darwin

### DIFF
--- a/Releases/v5.0.0/.claude/PAI/PULSE/com.pai.pulse.service
+++ b/Releases/v5.0.0/.claude/PAI/PULSE/com.pai.pulse.service
@@ -1,0 +1,18 @@
+[Unit]
+Description=PAI Pulse — unified daemon (cron, voice, observability, hooks)
+After=network.target
+
+[Service]
+Type=simple
+ExecStart=__BUN_PATH__ run pulse.ts
+WorkingDirectory=__HOME__/.claude/PAI/PULSE
+Restart=on-failure
+RestartSec=30
+TimeoutStopSec=5
+Environment=HOME=__HOME__
+Environment=PATH=__HOME__/.bun/bin:/usr/local/bin:/usr/bin:/bin
+StandardOutput=append:__HOME__/.claude/PAI/PULSE/logs/pulse-stdout.log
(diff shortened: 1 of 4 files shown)


alessandro-estate-guru:
We're duplicating work. Fixed here:

https://github.com/danielmiessler/Personal_AI_Infrastructure/pull/1116

danielmiessler:

Specifically applied:
- PULSE/package.json (new)

PR #1151, opened 2026-05-02
fix(service): resolve bash absolutely in spawnScript for Linux service envs

m8ryx:
When Pulse runs under systemd, the inherited PATH is minimal and "bash" is not resolvable by name — Bun.spawn(["bash", ...]) throws ENOENT, silently killing every cron job.

Use Bun.which("bash") with a /bin/bash fallback so the absolute path is resolved at module load time against whatever PATH is available, then cached for all subsequent spawnScript calls.

Ref: PR #1126 (fix/linux-support by dmcgoldrd)

This set of PRs attempts to capture the fixes for linux in those and add into simpler PRs for easier review/test/merge.


danielmiessler:

Specifically applied:
- PULSE/package.json (new)

PR #1152, opened 2026-05-02
fix(installer): cross-platform service check and multi-shell alias scan

m8ryx:
validate.ts: replace macOS-only launchd plist check with platform-branched check (launchd plist on Darwin, systemd user unit on Linux). Replace zsh-only .zshrc alias scan with a loop over .zshrc/.bashrc/.profile so the check passes regardless of the user's default shell.

actions.ts: reloadPulse() now checks the correct platform service file before attempting manage.sh restart — previously the macOS-only plist guard caused the reload to silently no-op on Linux.

References: https://github.com/danielmiessler/Personal_AI_Infrastructure/pull/1116

This set of PRs attempts to capture the fixes for linux in those and add into simpler PRs for easier review/test/merge.


danielmiessler:

Specifically applied:
- PULSE/package.json (new)

PR #1147, opened 2026-05-02
fix: wire Pulse TELOS dashboard to real user data

Alessandro-Improta:
## Overview
This PR wires the Pulse TELOS dashboard to live user data by fetching from the `/api/telos/overview` endpoint and implements a series of parser improvements to handle real-world Markdown variations in the TELOS primitive files (Missions, Goals, Challenges, Strategies, and Problems).

### 🔌 Exhaustive Data Hydration (`use-telos-data.ts`)
- **Runtime Fetching:** Transitioned the dashboard from static fallbacks to live API fetching with hydration via `useEffect`.
- **Robust Merging Logic:** Implemented an exhaustive `mergeTelosData` function that performs field-by-field validation and type-safe merging for the entire TELOS schema. This ensures the UI stays resilient by falling back to baseline data only when specific API fields are missing or malformed.
- **Coverage:** Added merging support for Dimensions, Snapshots, Problems, Missions, Goals, Metrics, Challenges, Strategies, Projects, Team, Budget, Recommendations, and Narrative Seeds.

### 🔍 Robust Markdown Parsing (`observability.ts`)
- **Preamble Support:** `parseSections` now scans file preambles for ID-bullets (e.g., `- **P2:**`) that appear before the first `##` heading, ensuring no data is lost in files like `PROBLEMS.md`.
- **Format Flexibility:** 
  - Added support for `Goal N: text` and `Challenge N: text` formats alongside standard ID-prefixes.
  - Updated regex to handle bold colons (e.g., `**P2:**`) frequently used in manual edits.
- **Prose Extraction:** Added a fallback for `MISSION.md` that extracts bold-labeled sections as missions when explicit ID headings are absent.
- **Heading Normalization:** Improved heading parsing to normalize word-prefixes into standard IDs (e.g., "Challenge 1" -> `C1`).

### 🛠️ Schema & Type Cleanup
- Enriched the API response with missing fields for `strategies` (`overcomes`) and `challenges` (`note`, `blocks`).
- Cleaned up internal types and introduced `Dict` helpers for cleaner data transformation.

## Verification
- Dashboard successfully hydrates with real data from local TELOS files.
- Verified parser handles both "Goal 1:" and "G1:" formats.
- Confirmed preamble items in `PROBLEMS.md` are correctly captured.

PR #1154, opened 2026-05-03
fix(telos): crash guards for one-mission users in Life Dashboard

brycemagera:
## Problem

Any user with fewer than 2 missions defined gets an immediate `TypeError` on the Telos page:

```
TypeError: Cannot read properties of undefined (reading 'horizon')
```

Three spots used `missions[1]` as a fallback:
- `sections.tsx` line 89 — fallback mission for rendering
- `horizon.tsx` line 15 — same
- `app.tsx` — initial mission state hardcoded to `"M1"`; when live data loads with only `M0`, no valid mission is selected

## Fix

`sections.tsx` + `horizon.tsx`: `missions[1]` → `missions[0]` + explicit null guard so the component can return null cleanly if no missions exist at all.

`app.tsx`: initial state `"M1"` → `"M0"`; added `useEffect` to reset to the first available mission ID when live data arrives with different IDs than the initial state assumed.

## How This Fits With the Other Telos PRs

Three open PRs each fix a different layer of the telos stack — all complementary, no conflicts:

| PR | Layer | Files |
|----|-------|-------|
| **#1127** (jmmarkiewicz) | CLI summary generator | `GenerateTelosSummary.ts` — regex, strategy bullets, goal section classifier |
| **#1147** (Alessandro-Improta) | Live data pipeline | `observability.ts` parser robustness + `use-telos-data.ts` fetch wired to `/api/telos/overview` |
| **#1154** (this PR) | Crash prevention | `sections.tsx`, `horizon.tsx`, `app.tsx` — guards for < 2 missions |

**Recommended merge order:** #1127 (standalone) → #1147 (data plumbing) → #1154 (crash guards surface once data flows).

## Revision Note

Originally this PR also introduced a new `/telos` endpoint in `pulse.ts` + `modules/telos.ts`. After reviewing #1147, that approach was dropped — #1147's use of the existing `/api/telos/overview` endpoint is architecturally correct and avoids a parallel endpoint. This PR is now focused exclusively on crash prevention.


danielmiessler:
Thanks a lot for this, @brycemagera — really appreciate you flagging it and taking the time to write the fix.

When I went back through the current source, this turns out to already be handled: one-mission/bullet-format crash guards for the Life Dashboard are already handled by the current parseIdEntries parser. The system has moved a fair bit since you opened this (the PAI → LifeOS rename and a few subsystem rewrites), so the gap you spotted has since been closed independently.

PR #1160, opened 2026-05-03
feat: configurable install locations via CLAUDE_CONFIG_DIR and PAI_DIR

neilinger:
## Why

`CLAUDE_CONFIG_DIR` and `PAI_DIR` are documented in `PAI/DOCUMENTATION/PAISystemArchitecture.md` ("Directory Structure") as the two env vars that control where PAI installs and runs. Until now, they didn't actually do anything — paths were hardcoded across the codebase, so setting either var had no effect.

Successor to #778 (v4.0.3 — `fix/pai-dir-hardcoded-paths`, narrower scope: 26 runtime files). This PR is the v5.0.0 successor: same goal, but with the centralized two-domain `lib/paths.{ts,sh}` libraries, full installer support, and ~150 migrated call sites — informed by additional PAI experience since #778.

The concrete user scenario this unblocks: keep Claude Code's tool config (`~/.claude` — settings, hooks, skills, agents) where Claude Code expects it, while putting the personal PAI data tree (`MEMORY`, `USER`, `ALGORITHM`) somewhere managed independently — a synced volume, a separate filesystem, an encrypted home, etc. The two domains are *related but separable*: PAI defaults to nesting inside Claude's config, but the install can pull them apart when `PAI_DIR` points to an absolute path outside `CLAUDE_CONFIG_DIR`.

A single `PAI_ROOT` would conflate "Anthropic's tool config" with "my data" and lose that flexibility. Two vars is the minimum that respects the seam.

## Architecture

```
CLAUDE_CONFIG_DIR  →  hooks, skills, agents, settings.json
PAI_DIR            →  MEMORY, ALGORITHM, TOOLS, USER, DOCUMENTATION
```

The two domains are orthogonal once both are resolved. `PAI_DIR` may live outside `CLAUDE_CONFIG_DIR`. Cross-domain access goes through absolute paths from env vars — never relative filesystem imports — so the layout works whether `PAI_DIR` nests inside `CLAUDE_CONFIG_DIR` (default) or sits on a different filesystem entirely.

## What changed

- **`hooks/lib/paths.ts` + `PAI/lib/paths.ts`** — mirror libraries, each self-contained so consumers in one domain never reach across the filesystem to import from the other. Resolve env vars with absolute-path validation, whitespace handling, and `$HOME`/`~` expansion. Byte-identical bodies below the docblock; a unit test guards drift.
- **`PAI/lib/paths.sh`** — shell-script helper providing the same contract.
- **~150 consumer files migrated** from hardcoded paths or ad-hoc `process.env.CLAUDE_CONFIG_DIR ?? ...` patterns to the centralized helpers.
- **Installer rework** — honors `CLAUDE_CONFIG_DIR` / `PAI_DIR` env vars, adds `--claude-config-dir` / `--pai-dir` CLI flags, rewrites `settings.json` placeholders, fixes plist substitution, and exports the resolved paths in the shell-rc snippet.

## Behavioral note: `settings.json` portability

The installer now writes **resolved absolute paths** into `settings.json` (e.g. `/Users/alice/.claude/hooks/...`) instead of `$HOME/.claude/...` literals.

**Why this is required.** Claude Code expands `$HOME` at runtime but does **not** expand `$CLAUDE_CONFIG_DIR`. If `CLAUDE_CONFIG_DIR != ~/.claude` and the literals stayed in the file, hooks would resolve to the wrong tree. Resolving at install time is the only way to make custom install roots work without changes inside Claude Code itself.

**Single-machine impact: none.** Hooks, permissions, and credential paths resolve to the same physical files under default config.

**Multi-machine impact: a regression for dotfile sync.** The generated `settings.json` is no longer portable across machines without re-running the installer. Users who version-control `~/.claude/settings.json` will see the file pin to the install machine's username and home directory. Accepted as the cost of supporting custom roots.

## Test evidence

- **41 bun unit tests** (`hooks/lib/paths.test.ts`, `PAI/PAI-Install/engine/paths.test.ts`) — GREEN
- **12 bats integration tests** (`PAI/lib/paths.bats`) — GREEN
- **Evals regression suite** (`skills/Evals/Suites/Regression/paths-resolution-regression.yaml`) — GREEN
- **Mirror-drift guard** — byte-identity check between the two `lib/paths.ts` files; if they ever diverge, the test fails

## Default-config compatibility

For users with no env vars set, runtime behavior is identical to pre-branch. `CLAUDE_CONFIG_DIR` defaults to `~/.claude`, `PAI_DIR` defaults to `${CLAUDE_CONFIG_DIR}/PAI`. No migration step required. (See "Behavioral note" above for the on-disk caveat.)

## Drive-by fix

Includes one small fix that's strictly speaking unrelated to install paths: `PAI/PULSE/Observability/observability.ts` already imports `yaml` on `main`, but `PAI/PULSE/Observability/package.json` doesn't declare it. Adding `yaml: ^2.8.3` to the dependency list. Surfaced while testing the install-paths feature — kept inline rather than in a separate PR because it's a one-line bug fix for a latent issue.

## Out of scope

- Mirror lib is deliberate (each domain self-contained, no cross-domain relative imports) — not proposed for elimination here.
- v4.0.3 / earlier release backports — preserved on side branches, not part of this PR.
- Pre-existing template substitution bugs (e.g. `{{DA_NAME}}`) — separate PR.
- Architecture redesign — this PR implements the existing spec, doesn't redesign it.


danielmiessler:
Hey @neilinger, thanks for raising this, and sorry it sat for a while.

PR #1165, opened 2026-05-03
Test/telos pulse contracts

m8ryx:
This is a much larger PR - addresses 3 related bugs with Pulse not pulling in the correct data. I think there are still some that may need fixed. In addition a suite of 113 tests to help in validation (these will not catch the case-sensitivity issue). Documentation on architecture and contracts.

Only tested on Ubuntu-family Linux.

 Three bug fixes:
  - Spark graphs in "Corners of Life" rendered flat — SPARK_SEEDS keys in subtabs.tsx didn't match
  the API-returned subtab IDs (business/finances/work/life vs. money/freedom/creative/relationships)
  - Preferences panel showed generic placeholders — /api/telos/overview hardcoded preferences: null;
   added buildPreferences() to read real TELOS markdown files
  - Music section absent entirely — music field missing from the PreferenceContext interface,
  FALLBACK data, and the render block in sections.tsx

  One supporting change:
  - Exported 6 pure parse functions from observability.ts to make them testable without running the
  server

  Test suite (113 tests, 4 files):
  - observability-parsers.test.ts — parse functions against synthetic fixtures
  - telos-api.test.ts — live API shape contract (skips gracefully when server is down)
  - telos-data-normalization.test.ts — merge/normalize logic as a pure function
  - telos-component-contracts.test.ts — FALLBACK shape and SPARK_SEEDS alignment

  Documentation:
  - README.md — architecture overview, three-layer data flow, coupling points, dev workflow
  - tests/CONTRACTS.md — detailed contract spec per layer, documented bugs, fixture format notes


danielmiessler:
Thanks a lot for this, @m8ryx — really appreciate you flagging it and taking the time to write the fix.

When I went back through the current source, this turns out to already be handled: thanks for the Linux/systemd Pulse support and the Telos/Pulse parsing hardening — all of it already lives in current source (the install engine has the same launchd/systemd check, PULSE casing and heading parsing are in place), and the PR targeted a since-removed v5.0.0 release snapshot, so we're closing it as already-handled. The system has moved a fair bit since you opened this (the PAI → LifeOS rename and a few subsystem rewrites), so the gap you spotted has since been closed independently.

PR #1177, opened 2026-05-04
Fix/migrate approve telos path

m8ryx:
fixed path in TELOS migration


danielmiessler:
Hey @m8ryx, thank you for the PR and the work that went into it.

PR #1188, opened 2026-05-05
 Fix Daniel Miessler blog URLs in PAISystemArchitecture.md

tzioup:
example.com -> danielmiessler.com


j0rg1 (approved):

alessandro-estate-guru (approved):

danielmiessler:
Closing — `example.com` is the intentional public-shipped placeholder in those docs. Not a leak, not a typo, not in scope to "fix." Different documents (e.g. `CLAUDE.md`) reference real URLs when they're routing-table entries the user needs to navigate to; the system-prompt/architecture docs use `example.com` deliberately.

Thanks for the catch — closing as not-a-bug.

Issue #1133, opened 2026-05-02
v5.0.0: CLAUDE.md @-imports USER/PROJECTS/PROJECTS.md but the file is not shipped


wojteksbt:
## Summary

`Releases/v5.0.0/.claude/CLAUDE.md` line 8 declares `@PAI/USER/PROJECTS/PROJECTS.md`, but the v5.0.0 tree ships no `PAI/USER/PROJECTS/` directory. Claude Code silently skips unresolvable `@`-imports, so every fresh install loses the project context the doc claims is loaded.

Line 74 doubles down on the promise:
> Startup context is `@`-imported above (PRINCIPAL_IDENTITY, DA_IDENTITY, **PROJECTS**, PRINCIPAL_TELOS) — always available.

The other four imports on line 8 resolve. Only `PROJECTS/PROJECTS.md` is missing.

## Reproduce

```
git clone https://github.com/danielmiessler/Personal_AI_Infrastructure
cd Personal_AI_Infrastructure/Releases/v5.0.0/.claude
ls PAI/USER/PROJECTS/   # → No such file or directory
```

## Suggested fix

**Option A (preferred) — ship a placeholder** at `PAI/USER/PROJECTS/PROJECTS.md` matching the style of `PRINCIPAL_IDENTITY.md`. The line-74 prose treats it as always-available, so a scaffold is the consistent answer.

**Option B — drop the import.** Remove line 8's entry and the "PROJECTS" mention on line 74; let `/interview` create it.

## Environment

PAI v5.0.0, `main` branch.

brycemagera:
+1 on this. While digging into a related schema-v1 path issue, I found the dead `USER/PROJECTS/PROJECTS.md` reference isn't isolated to `CLAUDE.md` — it propagates through several other files in `Releases/v5.0.0/.claude/`:

| File | Line | Reference |
|---|---|---|
| `CLAUDE.md` | 8 | `@PAI/USER/PROJECTS/PROJECTS.md` (this issue) |
| `settings.json` | 1211 | `postCompactRestore.fullFiles` entry |
| `hooks/RestoreContext.hook.ts` | 111 | `defaultFullFiles` array |
| `PAI/ALGORITHM/v5.7.0.md` | 643 | algorithm spec table |
| `PAI/ALGORITHM/v6.0.0.md` | 522 | algorithm spec table |
| `PAI/ALGORITHM/v6.1.0.md` | 527 | algorithm spec table |
[...]

PR #1212, opened 2026-05-08
fix(v5): restore identity context after compaction

tzioup:
## What

Adds `USER/DA_IDENTITY.md` and `USER/PRINCIPAL_IDENTITY.md` to `postCompactRestore.fullFiles` in the v5.0.0 release settings.

## Why

These two files define the core of the DA–Principal relationship: who the DA is, who it's working for, how they collaborate. They're loaded at session start via `@`-imports in CLAUDE.md:

```
@PAI/USER/PRINCIPAL_IDENTITY.md
@PAI/USER/DA_IDENTITY.md
```

But `@`-imports only fire at session start — they do **not** re-fire after mid-session context compaction. Long sessions (deep Algorithm runs, multi-step builds, extensive research) routinely hit the compaction threshold. When they do, these files are lost from context, and the DA loses awareness of who it is and who it's working for.

This is not a user preference — it's a gap in the default config that undermines PAI's core design.

### Philosophy reference

From the [Life OS Thesis](Releases/v5.0.0/.claude/PAI/DOCUMENTATION/LifeOs/LifeOsThesis.md):

> *"PAI is the Life Operating System... it manages the resources, processes, **identity**, memory, and interfaces that let you live and work."*

> *"Every real assistant does one job: **understand your current state, understand your ideal state**, and hill-climb you from one to the other."*

> *"Everyone running PAI names their own DA. The DA is personal — **a name, a voice, a personality, an identity**."*

From the [System Prompt](Releases/v5.0.0/.claude/PAI/PAI_SYSTEM_PROMPT.md):

> *"Its **primary directive is understanding the Principal** so that it can help them move from their current state to their IDEAL STATE."*

Identity is the foundation of the Current State → Ideal State loop. If the DA doesn't know who the Principal is — their goals, challenges, work patterns, communication preferences — it can't hill-climb toward anything. `postCompactRestore` exists precisely for files that must survive compaction. Identity files are the textbook case.

`PROJECTS.md` is already in `postCompactRestore` (correctly). The identity files belong alongside it — they're at least as load-bearing.

## Scope

- 1 file changed: `Releases/v5.0.0/.claude/settings.json`
- Net: +3 lines, -1 line
- Adds two entries to an existing array

### DIFF
--- a/Releases/v5.0.0/.claude/settings.json
+++ b/Releases/v5.0.0/.claude/settings.json
@@ -1208,7 +1208,9 @@
   "postCompactRestore": {
     "_docs": "System default files re-injected after compaction. User additions via EXTRA_COMPACT_RESTORE in PAI_CONFIG.yaml.",
     "fullFiles": [
-      "USER/PROJECTS/PROJECTS.md"
+      "USER/PROJECTS/PROJECTS.md",
+      "USER/DA_IDENTITY.md",
+      "USER/PRINCIPAL_IDENTITY.md"
     ]
   },
   "contextFiles": [],


brycemagera:
Big +1 on the intent — `postCompactRestore` is exactly the right place for identity files, and the reasoning in the description is on the money.

One issue worth flagging before merge: `USER/DA_IDENTITY.md` is the **legacy schema path**. Schema-v1 (introduced for the Pulse DA subsystem) moved the canonical DA identity to `USER/DA/<primary>/DA_IDENTITY.md`, with the active DA selected via `USER/DA/_registry.yaml`. The flat `USER/DA_IDENTITY.md` is now a deprecated placeholder; on systems that have run `/interview`, it's a `deprecated: true` stub redirecting to the new path.

Reference: [`PAI/DOCUMENTATION/Pulse/DaSubsystem.md`](../blob/main/Releases/v5.0.0/.claude/PAI/DOCUMENTATION/Pulse/DaSubsystem.md):

- Registry schema (~line 66): `_registry.yaml` with `primary: <da-name>`
- Migration table (~line 262): `PAI/USER/DA_IDENTITY.md` → `PAI/USER/DA/your-da/DA_IDENTITY.md`

So as written, this PR re-injects a deprecated stub into context post-compact instead of the live identity file. `PRINCIPAL_IDENTITY.md` is fine — that path is current.

### Suggested change

The cleanest fix is to point at the schema-v1 path. Two paths forward depending on how prescriptive you want to be:

**Option A — assume the registry exists** (matches what `/interview` produces today; users without `/interview` won't have a populated DA anyway):

```diff
 "fullFiles": [
-  "USER/PROJECTS/PROJECTS.md"
+  "USER/PROJECTS/PROJECTS.md",
+  "USER/PRINCIPAL_IDENTITY.md",
+  "USER/DA/<primary>/DA_IDENTITY.md"
 ]
```

…which requires the loader to expand `<primary>` from the registry. That's a hook-side change.

**Option B — list the path that would resolve via the existing hook**, once the hook itself is registry-aware. I just opened #1216 to fix the registry path resolution in `RestoreContext.hook.ts`'s Tier 2 block (which has the same stale path bug). If that lands, a static path here works fine because `getDAName()` is already in `hooks/lib/identity.ts` and could be wired up alongside.

Either way, swapping `USER/DA_IDENTITY.md` → `USER/DA/<primary>/DA_IDENTITY.md` (or whatever templating syntax the loader supports — possibly an `_expansion` field?) gets the intent of this PR landing correctly post-migration. Happy to help spin a follow-up if useful.

brycemagera:
One more, surfaced while looking at this: `USER/PROJECTS/PROJECTS.md` (the existing entry on the line above) is itself a dead path upstream — `Releases/v5.0.0/.claude/PAI/USER/PROJECTS/` doesn't ship in the release tree. That's tracked in #1133, which I just commented on with the broader catalog (it's also referenced from `RestoreContext.hook.ts:111`'s `defaultFullFiles` and 5 Algorithm version docs).

So this PR sits at the intersection of two half-finished migrations:
- DA: `USER/DA_IDENTITY.md` → `USER/DA/<primary>/DA_IDENTITY.md` (#1216 fixes the hook side)
- PROJECTS: dead `USER/PROJECTS/PROJECTS.md` reference → likely `USER/TELOS/PROJECTS.md` (per #1133)

Whatever direction lands for both, a single sweep of `postCompactRestore.fullFiles` would let this PR cover all three identity/state files cleanly:

```diff
 "fullFiles": [
-  "USER/PROJECTS/PROJECTS.md"
+  "USER/TELOS/PROJECTS.md",            // or wherever #1133 lands
+  "USER/PRINCIPAL_IDENTITY.md",
+  "USER/DA/<primary>/DA_IDENTITY.md"   // pending substitution model
 ]
```

Apologies for the double comment — wanted to keep the two migration concerns separately addressable.

PR #1215, opened 2026-05-08
feat(DAInterview): add --non-interactive mode for installers and CI

brycemagera:
## What

Adds a `--non-interactive` mode to `DAInterview.ts` so installers, CI harnesses, and team-template forks can populate a Pulse-compatible DA identity without TTY prompts.

## Why

Today `DAInterview.ts` is interactive-only — every input goes through Bun's `prompt()`, which requires a TTY. This blocks installers, automated tests, and team-template forks from setting up DA identity without human prompting.

Concretely, Pulse's DA subsystem (`PAI/PULSE/Assistant/module.ts`, `Assistant/checks/heartbeat.ts`, `Assistant/checks/growth.ts`) requires:

- `~/.claude/PAI/USER/DA/_registry.yaml` with `primary: <name>`
- `~/.claude/PAI/USER/DA/<name>/DA_IDENTITY.yaml`

If either is missing, the dashboard shows the "DA Identity is empty" empty-state and the heartbeat warns `"No primary DA in registry. Run: bun PAI/TOOLS/DAInterview.ts"`. The only path to those files today is the interactive wizard.

`install.sh` already has a `PAI_TEST_AUTOMATED=1` hook for headless installs, but the DA setup step still blocks on TTY without this PR.

## What ships

`--non-interactive` mode on `DAInterview.ts`:

```bash
bun PAI/TOOLS/DAInterview.ts --non-interactive \
  --name aria --principal Alex \
  [--display-name ARIA] [--preset friendly] [--dynamic peers] \
  [--main-voice-id <elevenlabs-id>] [--algorithm-voice-id <elevenlabs-id>] \
  [--force]
```

Builds the `InterviewData` shape from CLI flags + preset defaults, skipping all `prompt()` calls. Writes the same outputs as the interactive flow:

- `PAI/USER/DA/<name>/DA_IDENTITY.yaml`
- `PAI/USER/DA/<name>/DA_IDENTITY.md`
- `PAI/USER/DA/<name>/growth.jsonl`
- `PAI/USER/DA/<name>/opinions.yaml`
- `PAI/USER/DA/<name>/diary.jsonl`
- Updates `PAI/USER/DA/_registry.yaml`

### Required flags

- `--name <da-name>` — the DA's name (becomes slug after sanitize)
- `--principal <principal-name>` — the principal's name

### Optional flags

- `--display-name <DISPLAY>` — defaults to upper(name)
- `--preset <key>` — must match a key in `_presets.yaml`; default `"friendly"`
- `--dynamic <peers|commander|mentor>` — default `"peers"`
- `--main-voice-id <id>` — ElevenLabs voice for the DA's main voice
- `--algorithm-voice-id <id>` — ElevenLabs voice for ALGORITHM mode TTS (renders the `voice.algorithm` block)
- `--force` — overwrite an existing DA at the same slug

## Safety guards

- Required-flag validation with clear stderr + exit code 2
- `--preset` and `--dynamic` validated against the allowed sets
- Value-taking flags reject `--`-prefixed tokens (so `--name --principal X` errors instead of silently consuming `--principal` as the name)
- Empty-slug detection (e.g. `--name "★★★"` no longer corrupts `DA_DIR/` itself)
- Refuses to overwrite an existing DA without `--force`
- `--non-interactive` + `--update` combination explicitly rejected (update support can land in a follow-up)

Output is one parseable line on stdout — easy for wrapping scripts to capture:

```
DAInterview: created DA "<name>" (slug=<slug>) at <daDir>
```

## Out of scope

- Modifying `PAI-Install/engine/actions.ts` to call the new mode. That's a separate concern; this PR just adds the primitive.
- Backporting newer DA schema fields. The generator already treats them as optional.
- Updating `PAI/USER/DA/_example/identity.yaml` (currently uses an older flat schema) — cosmetic, no code path consumes it.

## Testing

Verified end-to-end against a temp `HOME` so the local `<da_name>/DA_IDENTITY.yaml` was untouched (mtime confirmed unchanged after the test run):

| Scenario | Expected | Actual |
|---|---|---|
| Missing `--name` | exit 2, stderr names `--name` | ✓ |
| Missing `--principal` | exit 2, stderr names `--principal` | ✓ |
| `--name --principal X` (flag-eats-flag) | exit 2, "requires a value (got: --principal)" | ✓ |
| Bad `--preset` | exit 2, lists available presets | ✓ |
| Bad `--dynamic` | exit 2, lists `peers, commander, mentor` | ✓ |
| `--name "★★★"` (empty slug) | exit 2, slug-emptiness error | ✓ |
| Happy path (required only) | exit 0, all 5 files + registry written | ✓ |
| Happy path with all flags | exit 0, voice block correctly populated, YAML parses cleanly | ✓ |
| Re-run without `--force` | exit 2, refuses to overwrite | ✓ |
| Re-run with `--force` | exit 0, overwrites cleanly | ✓ |
| `--non-interactive --update` | exit 2, explicit rejection message | ✓ |
| Generated YAML round-trip via `yaml` package | parses successfully | ✓ |

Interactive mode is structurally untouched — the new path branches at the top of `main()` based on `nonInteractive`, leaving the existing `printBanner` → phases → summary flow exactly as it was.

## Follow-ups (not in this PR)

- Wire `PAI-Install/engine/actions.ts` to call this after collecting basic identity info, so fresh installs end up with a Pulse-compatible DA on first session instead of the "DA Identity is empty" empty-state.
- Add `--update --non-interactive` support (currently rejected) for scripted DA updates.
- Refresh `PAI/USER/DA/_example/identity.yaml` to match the schema-v1 structure.

### DIFF
--- a/Releases/v5.0.0/.claude/PAI/TOOLS/DAInterview.ts
+++ b/Releases/v5.0.0/.claude/PAI/TOOLS/DAInterview.ts
@@ -1,14 +1,32 @@
 #!/usr/bin/env bun
 
 /**
- * DAInterview.ts -- Interactive CLI wizard for creating DA identities
+ * DAInterview.ts -- CLI wizard for creating DA identities
  *
- * Usage:
+ * INTERACTIVE MODE (default — TTY required):
  *   bun PAI/TOOLS/DAInterview.ts                    # Quick mode (default)
  *   bun PAI/TOOLS/DAInterview.ts --depth standard   # Quick + Standard
  *   bun PAI/TOOLS/DAInterview.ts --depth deep       # All phases
  *   bun PAI/TOOLS/DAInterview.ts --update           # Update existing primary DA
  *   bun PAI/TOOLS/DAInterview.ts --update --da devi # Update specific DA
+ *
+ * NON-INTERACTIVE MODE (no TTY needed — for installers, CI, scripts):
+ *   bun PAI/TOOLS/DAInterview.ts --non-interactive \
+ *     --name aria --principal Alex \
+ *     [--display-name ARIA] [--preset friendly] [--dynamic peers] \
+ *     [--main-voice-id <elevenlabs-id>] [--algorithm-voice-id <elevenlabs-id>]
+ *
+ *   Required flags: --name, --principal
+ *   Optional flags: --display-name (defaults to upper(name)),
+ *                   --preset (must match a key in _presets.yaml; default "friendly"),
+ *                   --dynamic (peers|commander|mentor; default "peers"),
+ *                   --main-voice-id, --algorithm-voice-id (both default empty),
+ *                   --force (overwrite an existing DA at the same slug).
+ *
+ *   By default --non-interactive refuses to overwrite an existing
+ *   DA_IDENTITY.yaml — pass --force or run --update interactively instead.
+ *   --non-interactive does not currently combine with --update; use one or the
+ *   other. Update support can be added in a follow-up.
  *
  * Creates:
  *   PAI/USER/DA/{name}/DA_IDENTITY.yaml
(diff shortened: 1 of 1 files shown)


danielmiessler:
Hey @brycemagera, thanks for raising this, and sorry it sat for a while.

PR #1253, opened 2026-05-13
feat(Research): add provider-specific execution guides for researcher agents

Kharsys:
## Summary

Three of the four Researcher agents (Gemini, Grok, Perplexity) currently have no provider-specific call code anywhere in the Research skill or in the agent files. With no API/CLI wire-format defined, each non-Claude agent collapses to the same backend (built-in `WebSearch` under a different persona name), so "multi-model research" via Standard / Extensive mode does not actually exercise multiple models. This PR adds the missing execution layer.

## What's added

A new directory `Releases/v5.0.0/.claude/skills/Research/Workflows/Execution/` with four guides, one per researcher:

- `ClaudeExecution.md` documents the `WebSearch` path (included for parity, no behavior change)
- `GeminiExecution.md` documents the `gemini` CLI invocation with Google Search grounding
- `GrokExecution.md` documents the xAI Chat Completions API call with live search enabled
- `PerplexityExecution.md` documents the Perplexity Sonar Chat Completions API call

Each guide owns the provider call shape (CLI or `curl` with `jq -n --arg` for safe user-input escaping), the model selection table mapped to research modes, citation extraction rules, URL verification via `curl`, the file-based output JSON schema, and a fallback policy that retries once at a lower model tier then writes a stub error file. Never silently swaps providers.

## Agent changes

Each of `ClaudeResearcher.md`, `GeminiResearcher.md`, `GrokResearcher.md`, `PerplexityResearcher.md` gains step 3 in its Mandatory Startup Sequence: read the sibling execution guide. Existing persona, voice, output format, Research Philosophy and Research Methodology sections are unchanged. Nothing was removed.

## SKILL.md change

A new `### Agent Execution Guides` subsection under `## Workflow Routing` points to the four guides and states the no-silent-fallback policy at the skill level.

## File-based output schema

Each Execution guide documents a shared JSON shape for `${research_dir}/<agent-slug>.json`:

```json
{
  "agent": "claude|gemini|grok|perplexity",
  "query": "...",
  "summary": ["..."],
  "sources": [{ "title": "...", "url": "...", "verified": true }],
  "confidence": 0.0,
  "full_findings": "..."
}
```

The schema is shared across all four researchers so a downstream synthesis step can cross-reference findings agnostic of which provider produced them. Current upstream Research workflows synthesize inline as text rather than from files; this schema is therefore forward-looking with respect to the merged tree, standardizing the contract for any future file-based synthesizer. If `research_dir` is not provided by the caller, each researcher returns the same payload inline and skips file writing.

## Correctness property preserved

A researcher whose provider call fails writes a stub file (`summary: []`, error reason in `full_findings`) rather than substituting `WebSearch`. The orchestrator's decision stays explicit: it can surface the gap to the user or substitute a different researcher. Silent substitution would defeat the purpose of multi-model research.

## Why no OpenAI / GPT researcher

GPT plus web search via the OpenAI Responses API is architecturally similar to the Claude plus `WebSearch` path (LLM-with-search, citation-after-generation). It does not add an orthogonal coverage angle that the existing four researchers do not already cover. Grok contributes live X data and STEM strength; Perplexity contributes retrieval-first architecture with citations bundled into the response; Gemini contributes Google Search grounding. Each is structurally different. If a fifth researcher pointed at GPT-4o/5 is desired later, it can be added with the same Execution-guide pattern.

## Smoke test recipe

Local setup expects `~/.claude/.env` with `XAI_API_KEY`, `PERPLEXITY_API_KEY`, and `GEMINI_API_KEY` populated, plus the `gemini` CLI installed (validated against v0.39.1).

```bash
# Gemini
gemini -m gemini-2.5-flash -p "what is Bun JavaScript runtime" --grounding google_search

# xAI Grok
source ~/.claude/.env
BODY=$(jq -n --arg q "what is Bun JavaScript runtime" \
  '{model:"grok-3", messages:[{role:"system",content:"Research assistant. Cite sources."},{role:"user",content:$q}], search_parameters:{mode:"on"}, max_tokens:512}')
curl -sS -H "Authorization: Bearer $XAI_API_KEY" -H "Content-Type: application/json" \
  -d "$BODY" https://api.x.ai/v1/chat/completions | jq '.choices[0].message.content, .citations'

# Perplexity Sonar
BODY=$(jq -n --arg q "what is Bun JavaScript runtime" \
  '{model:"sonar", messages:[{role:"system",content:"Research assistant. Cite sources."},{role:"user",content:$q}], max_tokens:512, return_citations:true}')
curl -sS -H "Authorization: Bearer $PERPLEXITY_API_KEY" -H "Content-Type: application/json" \
  -d "$BODY" https://api.perplexity.ai/chat/completions | jq '.choices[0].message.content, .citations'
```

Expected: response content plus at least one citation URL. Missing-key scenarios should produce auth errors that each agent maps to the stub error file documented in the corresponding execution guide.

## Diff stat

```
9 files changed, 583 insertions(+), 4 deletions(-)
```

### DIFF
--- a/Releases/v5.0.0/.claude/agents/ClaudeResearcher.md
+++ b/Releases/v5.0.0/.claude/agents/ClaudeResearcher.md
@@ -83,7 +83,13 @@ curl -X POST http://localhost:31337/notify \
    - This loads all necessary Skills, standards, and domain knowledge
    - DO NOT proceed until you've read this file
 
-3. **Then proceed with your task**
+3. **Load your provider execution guide:**
+   - Read: `~/.claude/skills/Research/Workflows/Execution/ClaudeExecution.md`
+   - This file holds the WebSearch protocol used by this researcher, multi-mode depth strategy, citation handling, URL verification protocol, and file-based output schema
+   - Unlike the Gemini, Grok, and Perplexity researchers, this researcher does not call an external provider API; the execution guide documents the WebSearch path for parity with the other three
+   - DO NOT proceed until you've read this file
+
+4. **Then proceed with your task**
 
 **This is NON-NEGOTIABLE. Load your context first.**
 
--- a/Releases/v5.0.0/.claude/agents/GeminiResearcher.md
+++ b/Releases/v5.0.0/.claude/agents/GeminiResearcher.md
@@ -83,7 +83,13 @@ curl -X POST http://localhost:31337/notify \
    - This loads all necessary Skills, standards, and domain knowledge
    - DO NOT proceed until you've read this file
 
-3. **Then proceed with your task**
+3. **Load your provider execution guide:**
+   - Read: `~/.claude/skills/Research/Workflows/Execution/GeminiExecution.md`
+   - This file holds the gemini CLI invocation shape, model selection table by research mode, fallback chain on rate-limit or timeout, citation extraction rules, URL verification protocol, and file-based output schema
+   - Execute according to that file. Never silently fall back to Claude WebSearch for primary research; if the gemini CLI is unavailable or fails after one retry at a lower model tier, return the structured error documented in the execution guide
+   - DO NOT proceed until you've read this file
+
+4. **Then proceed with your task**
 
 **This is NON-NEGOTIABLE. Load your context first.**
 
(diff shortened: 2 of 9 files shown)


danielmiessler:
Thanks for building this, @Kharsys — it's a thoughtful contribution and clearly real work went into it.

After looking at it against where the system is now, I'm going to pass on merging it: thanks for the multi-provider researcher execution-guide proposal; the researchers are currently Claude WebSearch personas, not real provider API callers, so wiring true per-provider execution is a design call rather than a drop-in port. It's more of a product/direction call than a bug fix, and it isn't the direction I'm taking that part of the system right now.

PR #1262, opened 2026-05-14
Pulse: add Assistant module + cron stubs (fix empty DA Identity banner)

atabisz:
## Summary

The Pulse Life Dashboard ships an `Assistant` page (`Releases/v5.0.0/.claude/PAI/PULSE/Observability/src/app/assistant/page.tsx`) and `pulse.ts` imports `./Assistant/module` at startup, but the backend module — and the cron scripts referenced by `PULSE.toml` — were never included in the v5.0.0 release.

Effect: the `[da].enabled = true` config plus a populated `USER/DA_IDENTITY.md` is not enough to make the Assistant tab work. `pulse.ts:117` swallows the missing import in a try/catch, all `/assistant/*` requests fall through to `404`, and the page renders the empty-state banner ("DA Identity is empty — let's fill it in") forever, regardless of identity content.

This PR fills that gap.

## What's added

`Releases/v5.0.0/.claude/PAI/PULSE/Assistant/`

- `module.ts` — exports `startAssistant`, `stopAssistant`, `assistantHealth`, `handleAssistantRequest`. Implements 10 endpoints matching the frontend contract:
  - `GET /assistant/identity` — name, full name, role, color, origin story, principal, uptime
  - `GET /assistant/health` — `identity_loaded` flag, primary DA, scheduled task count, opinions count
  - `GET /assistant/personality` — base description, traits (defaults + overrides), preferences, autonomy, writing, voice
  - `GET/POST /assistant/tasks` — list (DA + Pulse cron) and create
  - `DELETE /assistant/tasks/:id` — cancel
  - `PATCH /assistant/personality/traits` — update trait values
  - `GET /assistant/diary` — diary entries
  - `GET /assistant/opinions` — raw opinions YAML + count
  - `GET /assistant/avatar` — image (png/jpg/webp) if present
- `checks/heartbeat.ts` — writes `MEMORY/DA/heartbeat.json` (cron job `assistant-heartbeat`).
- `checks/tasks.ts` — fires due one-time tasks via `/notify` (cron job `assistant-tasks`).
- `checks/diary.ts` — daily diary stub (cron job `assistant-diary`).
- `checks/growth.ts` — weekly growth report placeholder (cron job `assistant-growth`).

## Data sources

- Identity → `~/.claude/PAI/USER/DA_IDENTITY.md` (parsed in-process, no extra deps)
- Principal → `~/.claude/PAI/USER/PRINCIPAL_IDENTITY.md`
- Tasks / diary / opinions / personality overrides → `~/.claude/PAI/MEMORY/DA/`
- Avatar → `~/.claude/PAI/USER/DA/avatar.{png,jpg,webp}`

## Verification

Tested on **Ubuntu 24.04.4 LTS** (kernel 6.17). Restarted Pulse, then:

```
$ curl -s http://localhost:31337/assistant/health
{"status":"ok","primary_da":"Clive","identity_loaded":true,...}
```

The `assistant` page now renders the populated DA card (name, role, origin story, principal, uptime, opinions count) and the cron jobs section, instead of the empty-state banner. Verified end-to-end via headless Chrome screenshot.

**No automated tests added** — these endpoints are thin readers over identity / cron / opinions files; the meaningful failure mode (missing module not loading) is what the manual verification above exercises directly. Following PLATFORM.md guideline 5 ("at minimum, manual verification steps"), the manual verification recipe is:

1. Restart Pulse.
2. `curl -s http://localhost:31337/assistant/health` → expect `identity_loaded:true` and a non-null `primary_da`.
3. `curl -s http://localhost:31337/assistant/identity | jq .name` → matches `**Name:**` in `~/.claude/PAI/USER/DA_IDENTITY.md`.
4. Open `/assistant` in the browser — expect the populated DA card, not the empty-state banner.

## Backward compatibility

- **No regression possible on existing platforms.** Before this PR, `pulse.ts:117` already wrapped the `./Assistant/module` import in `try/catch` and silently swallowed the failure, so every prior install was running without the module. Adding the module supplies the missing implementation; it cannot break a code path that was already a no-op.
- **No new dependencies.** Pure `fs`/`path` plus Bun's built-in `Response.json` / `crypto.randomUUID` — nothing to install.
- **No frontend changes.** The page contract was already correct.
- **No `PULSE.toml` change.** Cron job names match what the existing config already expects.

## Platform coverage

- ✅ **Linux (Ubuntu 24.04.4 LTS)** — runtime-tested, all 10 endpoints + 4 cron jobs verified.
- ⚠️ **macOS** — not re-verified on macOS by this contributor. The module uses only cross-platform Node/Bun APIs (`fs`, `path`, `Response`, `crypto.randomUUID`) and reads paths via `process.env.HOME`, with no platform-specific syscalls or shell invocations, so macOS behavior should match — but per PLATFORM.md guideline 2 ("Document what you can't"), this is documented as untested rather than claimed.
- ❌ **Windows** — out of scope (consistent with PLATFORM.md's existing Windows status).

## Notes / scope

- Diary/opinions content pipelines are out of scope: stubs land empty arrays, and the UI handles that gracefully (existing empty-list rendering).
- Cron stubs are intentionally minimal (`heartbeat`, `tasks`, `diary`, `growth`) — they satisfy the `PULSE.toml` references and unblock the Assistant page; richer behavior belongs in follow-up PRs.

### DIFF
--- a/Releases/v5.0.0/.claude/PAI/PULSE/Assistant/checks/diary.ts
+++ b/Releases/v5.0.0/.claude/PAI/PULSE/Assistant/checks/diary.ts
@@ -1,0 +1,26 @@
+/**
+ * Assistant daily diary — placeholder that writes a diary entry stub for
+ * today if one doesn't already exist. Real summarization will replace this
+ * once the diary content pipeline is built. Output sentinel: NO_ACTION.
+ */
+import { join } from "path"
+import { existsSync, writeFileSync, mkdirSync } from "fs"
+
+const HOME = process.env.HOME ?? ""
+const DIR = join(HOME, ".claude", "PAI", "MEMORY", "DA", "diary")
+mkdirSync(DIR, { recursive: true })
+
+const today = new Date().toISOString().slice(0, 10)
+const path = join(DIR, `${today}.json`)
+if (!existsSync(path)) {
+  writeFileSync(path, JSON.stringify({
+    date: today,
+    interaction_count: 0,
+    topics: [],
+    mood: "neutral",
+    avg_rating: 0,
+    notable_moments: [],
+    learning: null,
+  }, null, 2))
+}
+console.log("NO_ACTION")
--- a/Releases/v5.0.0/.claude/PAI/PULSE/Assistant/checks/growth.ts
+++ b/Releases/v5.0.0/.claude/PAI/PULSE/Assistant/checks/growth.ts
@@ -1,0 +1,6 @@
+/**
+ * Assistant weekly growth report — placeholder. Future: synthesize weekly
+ * patterns from diary + opinions + heartbeats and emit a summary. Output
+ * sentinel: NO_ACTION.
+ */
+console.log("NO_ACTION")
--- a/Releases/v5.0.0/.claude/PAI/PULSE/Assistant/checks/heartbeat.ts
+++ b/Releases/v5.0.0/.claude/PAI/PULSE/Assistant/checks/heartbeat.ts
@@ -1,0 +1,12 @@
+/**
+ * Assistant heartbeat — writes ts to MEMORY/DA/heartbeat.json so the
+ * dashboard "last_heartbeat" field is non-null. Output sentinel: NO_ACTION.
+ */
+import { join } from "path"
+import { mkdirSync, writeFileSync } from "fs"
+
+const HOME = process.env.HOME ?? ""
+const DIR = join(HOME, ".claude", "PAI", "MEMORY", "DA")
+mkdirSync(DIR, { recursive: true })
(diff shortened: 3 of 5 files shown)


danielmiessler:
Thanks for this — built independently for 5.1.0 before this PR landed. The 5.1.0 shipped module has the same 10-endpoint contract and the four cron stubs, with the same `module.ts` + `checks/` layout. The directory was first introduced in our private tree on April 21 (commit `130315c10`), predating this PR by ~3 weeks, so we converged on the same design rather than diverging.

Closing as superseded by the 5.1.0 work. Your verification notes (`curl /assistant/health`, `jq .name` check, headless Chrome screenshot) are exactly the bar future Pulse PRs should hit — thank you.

PR #1267, opened 2026-05-15
Fix tools directory for pai alias

maxolasersquad:
The current code adds the alias `alias pai='bun <home-dir>/.claude/PAI/Tools/pai.ts'`. However the `Tools` directory is actually `TOOLS`. This creates a broken alias on Linux.


Seikilos:
Did you verify everything else works afterwards?

```
❯ grep -R "Tools" 2>/dev/null | wc -l
533
❯ grep -R "TOOLS" 2>/dev/null | wc -l
309

```
There are a lot of casing issues right now.
Wouldn't it make more sense to sweep all files and replace TOOLS with Tools?

I really just think of symlinking Tools to TOOLS 😅

PR #1268, opened 2026-05-15
Add optional --install-mode option to choose installer

maxolasersquad:
This adds the ability to run the install with --install-mode or -m to specify either the gui or cli installer. Useful when wanting to force the cli installer, even on desktop environments.

This commit also cleans up how the shell profile gets bun added to the path by writing both lines to the file in a single transaction, and it removes an unused variable.


danielmiessler:
Hey @maxolasersquad, thanks for raising this, and sorry it sat for a while.

PR #1274, opened 2026-05-17
fix(pulse): cross-platform Windows + case-sensitive filesystem compatibility

alexgordon25:
@
## Summary

Two minor cross-platform fixes surfaced while running PAI v5.0.0 on Windows 11 (Pulse as a Windows service under NSSM). Both also help case-sensitive Linux / case-sensitive macOS filesystems.

## Fix 1 — `Pulse` → `PULSE` directory casing (11 occurrences, 8 files)

Directory on disk is `PAI/PULSE/` (uppercase), but several modules reference it as `PAI/Pulse/` (lowercase). Works on case-insensitive Windows NTFS / macOS HFS+ default, fails silently on case-sensitive Linux and APFS-case-sensitive Mac. In particular `getDashboardDir()` in `observability.ts` resolved to a missing path, producing 404 at the dashboard root.

Files touched:
- `Observability/observability.ts` (×3)
- `Performance/cost-aggregator.ts`
- `checks/github-work.ts`
- `checks/notification-governor.ts`
- `checks/poller-meta-monitor.ts` (×2)
- `lib.ts`
- `modules/imessage.ts` (×2)
- `modules/user-index.ts`

## Fix 2 — `path.isAbsolute()` for Windows path detection

`getDashboardDir()` in `Observability/observability.ts` used `dir.startsWith("/")` to detect absolute paths. POSIX-only — Windows absolute paths begin with a drive letter (`C:\…`). Replaced with `path.isAbsolute()` which handles both.

## Testing

Both patches running in production on a Windows 11 desktop install for ~6 days under NSSM (`PaiPulse` service). Dashboard at `localhost:31337/` serves all routes; cron jobs run clean; cloud-sync events continue posting.

## Scope

PR intentionally excludes identical occurrences in standalone scripts not imported from the live daemon (`pulse-old.ts`, `pulse-unified.ts`, `run-job.ts`, `setup.ts`). Happy to extend if you would like the sweep to include them.

## Diff stats

8 files changed, +17 / −15. No new dependencies. No behavior change on case-insensitive filesystems.
@

### DIFF
--- a/Releases/v5.0.0/.claude/PAI/PULSE/Observability/observability.ts
+++ b/Releases/v5.0.0/.claude/PAI/PULSE/Observability/observability.ts
@@ -24,7 +24,7 @@
  *   GET  /, /work, /telos, /health, etc. — Static Next.js pages (fallback handler)
  */
 
-import { join, extname } from "path"
+import { join, extname, isAbsolute } from "path"
 import { readFileSync, readdirSync, existsSync, realpathSync } from "fs"
 import YAML from "yaml"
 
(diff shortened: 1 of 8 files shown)


danielmiessler:
Hey @alexgordon25, thanks for raising this, and sorry it sat for a while.

PR #1284, opened 2026-05-18
fix(pulse/telegram): drop SDK session resume to prevent empty responses

### LINKED ISSUE #1143: PAI/PULSE/modules/telegram.ts: short messages produce empty replies (numTurns=0); literal {{DA_NAME}}/{{PRINCIPAL_NAME}} reach SDK and chat-log
## Summary

Two related runtime bugs in `PAI/PULSE/modules/telegram.ts` that bite together: short user replies ("Agreed", "OK", "thanks") return the fallback `"Sorry, I wasn't able to generate a response. Try again?"` instead of a real response. And the bot's runtime systemPrompt has literal `{{DA_NAME}}` / `{{PRINCIPAL_NAME}}` placeholder strings — the substitution gap #1135 calls out for `PAI_SYSTEM_PROMPT.md` is not applied to TypeScript source either.

---

## Bug 1 — SDK session resume + manual history-prepend produces `numTurns: 0` on short messages

`telegram.ts:182-188` builds a prompt that prepends the last 10 messages of conversation history as plain text:

```ts
prompt = `Previous conversation:\n${historyText}\n\nPrincipal's new message: ${sanitized}`
```

`telegram.ts:218-221` ALSO sets `sdkOptions.resume = lastSessionId`, resuming the prior agent session that already contains the same history:

```ts
if (lastSessionId) {
  sdkOptions.resume = lastSessionId
}
```

So the SDK gets the conversation history twice. For long content-rich messages it muddles through. For short acknowledgments after a complex multi-turn previous response, the resumed agent loop short-circuits with `numTurns: 0, cost: 0` after ~3s — the agent sees the new "Agreed" as redundant against its existing transcript and never issues a model call. The SDK returns no streamed text, `fullText` stays empty at `telegram.ts:293`, and `telegram.ts:294` substitutes the `"Sorry, I wasn't able..."` fallback string.

### Receipts

Six consecutive turns from one session, all sharing the same `sessionId`:

| User message | numTurns | cost ($) | Outcome |
|---|---:|---:|---|
| `(37 chars, work)` | 1 | 0.27 | OK |
| `(38 chars)` | 1 | 0.28 | OK |
| `(130 chars)` | 13 | 0.86 | OK |
| `(998 chars)` | 1 | 0.40 | OK |
| **`Agreed` (6 chars)** | **0** | **0** | **fallback** |
| **`I agreed to what you proposed` (29 chars)** | **0** | **0** | **fallback** |

Pulse log lines for the failing turn:

```
"Message received" textLength=6
"Session initialized" sessionId=d0eb813d-...
"SDK session complete" durationMs=3088 numTurns=0 cost=0 sessionId=d0eb813d-...
"Empty response from SDK"
"Response sent" responseLength=55
```

### Suggested fix

Pick one of the two history paths, not both:

- **Option A** — drop `resume`. Each turn starts fresh; the manually-prepended history is the only context. Simpler and avoids the SDK's "this session already produced an answer" short-circuit.
- **Option B** — keep `resume`, drop the manual prepend. Trust the SDK's session memory to carry context.

I'd lean toward Option A for predictability — the resumed-session code path also accumulates max-turns budget across resumes, which is a separate latent issue.

---

## Bug 2 — Literal `{{DA_NAME}}` / `{{PRINCIPAL_NAME}}` strings reach the SDK and the chat log

`telegram.ts:201-214` constructs the systemPrompt body with hardcoded placeholder strings that are never substituted at runtime:

```ts
append: `\n\n## TELEGRAM MODE OVERRIDE (highest priority — overrides CLAUDE.md format rules)

You are {{DA_NAME}}, responding via Telegram. {{PRINCIPAL_NAME}} is messaging you from his phone.

CRITICAL RULES FOR TELEGRAM MODE:
- ...
- Speak as {{DA_NAME}} — first person, natural, conversational, like talking to a friend
...
```

So the bot's runtime system prompt literally contains the strings `{{DA_NAME}}` and `{{PRINCIPAL_NAME}}`. Same issue in the chat-log writer at `telegram.ts:71-79`:

```ts
const entry = `\n### ${ts}\n**{{PRINCIPAL_NAME}}:** ${userMsg}\n\n**{{DA_NAME}}:** ${botMsg}\n\n---\n`
```

Result in `~/.claude/PAI/PULSE/logs/telegram/chat-log.md`:

```
**{{PRINCIPAL_NAME}}:** Agreed
**{{DA_NAME}}:** Sorry, I wasn't able to generate a response. Try again?
```

### Reproduce

```bash
grep -nE '\{\{[A-Z_]+\}\}' ~/.claude/PAI/PULSE/modules/telegram.ts
# lines 77 (chat log entry), 201, 203, 209, 211 (systemPrompt body)

tail -20 ~/.claude/PAI/PULSE/logs/telegram/chat-log.md
# shows literal {{DA_NAME}} / {{PRINCIPAL_NAME}} in entries
```

### Suggested fix

Either substitute the placeholders at module init (read identity from `PAI/USER/DA_IDENTITY.md` / `PRINCIPAL_IDENTITY.md` like the installer does for other files), or replace the placeholder strings with abstract role labels — `"the DA"` / `"the principal"` — since the system-prompt context is functional, not display.

Note: #1135's installer-side fix won't reach this code path because the strings are in TypeScript source, not in `PAI_SYSTEM_PROMPT.md`. This needs source-code substitution (or a build-time codegen step), not just a fix in `engine/actions.ts`.

---

## Environment

PAI v5.0.0, Bun 1.3.9, macOS, `claude-agent-sdk` 0.1.x.

catchingknives:
Fixes #1143.

## Problem

After a handful of messages on a long-lived Pulse process, the Telegram bot stops replying and ships the canned fallback `"Sorry, I wasn't able to generate a response. Try again?"`. Affects any PAI user with the Telegram module enabled long enough to hit it. Most reliably reproduced (per #1143) by sending a short acknowledgment like "Agreed" or "OK" after a content-rich previous turn.

## Root cause

`startTelegram` carries conversation state on two redundant channels:

1. Last 10 exchanges spliced into the prompt as `Previous conversation: …` (lines 180–188).
2. `resume: lastSessionId` passed to the SDK's `query()` (the removed block).

The SDK gets the same history twice. For long messages it muddles through; for short acknowledgments the resumed session short-circuits in 2–3 s with `numTurns: 0`, `cost: 0`, empty `result` — and `fullText` stays empty, triggering the fallback. #1143 has the same diagnosis and explicitly recommends "Option A — drop `resume`" as the predictable fix. This PR is that fix.

Evidence from a real session, all turns on the same `sessionId`:

| Time (UTC) | numTurns | cost  | result |
|------------|---------:|------:|--------|
| 14:43      | 1        | $0.42 | OK     |
| 15:42      | 1        | $0.31 | OK     |
| 15:58      | 0        | $0    | empty  |
| 18:04      | 0        | $0    | empty  |
| 18:07      | 0        | $0    | empty  |

(Independent receipts in #1143 from a different user — same pattern, different session IDs.)

## Fix

Drop the `resume` pass. The in-prompt history already carries conversational context (the last 10 exchanges — what users perceive as "memory"), so removing SDK-side resume is a no-op for user experience and removes the saturation failure mode. `lastSessionId` capture is kept for log correlation.

## Compatibility

No config changes. Behavior identical from the user's perspective except that responses no longer go empty.

## Relationship to #1281

#1281 (`v2 — per-thread sessions, live tool reactions, empty-response recovery`) addresses the same symptom but bundles it with a thread-aware rewrite (+575/-143). It keeps `resume` and wraps an empty response in a retry-with-fresh-session recovery layer. That works, but treats the symptom rather than the cause — and ships a much larger surface to review.

This PR is offered as a complementary minimal-diff path: 12 lines, addresses #1143's specific recommendation, doesn't preclude #1281's broader threading/observability work. If #1281 lands first, this PR's change is subsumed (the v2 runner no longer passes `resume` either, modulo its own design). If this PR lands first, #1281 rebases cleanly on top since the removed lines are not part of its retained code path.

### DIFF
--- a/Releases/v5.0.0/.claude/PAI/PULSE/modules/telegram.ts
+++ b/Releases/v5.0.0/.claude/PAI/PULSE/modules/telegram.ts
@@ -215,11 +215,13 @@ CRITICAL RULES FOR TELEGRAM MODE:
         },
       }
 
-      // Resume previous session for context continuity
-      if (lastSessionId) {
-        sdkOptions.resume = lastSessionId
-      }
-
+      // Intentionally NOT passing `resume: lastSessionId`. The in-prompt
+      // history block above already carries the last-10 exchanges of context;
+      // passing `resume` on top of that double-books conversation state and
+      // saturates the SDK session after a few turns — at which point query()
+      // short-circuits with numTurns:0 / cost:0 / empty result, producing the
+      // user-visible "Sorry, I wasn't able to generate a response" fallback.
+      // `lastSessionId` is still tracked below for log correlation only.
       const conversation = query({ prompt, options: sdkOptions as any })
 
       // Collect response with timeout


danielmiessler:
Thanks @catchingknives, the drop-SDK-resume fix is already in the current source (the Telegram module was reworked for dynamic voice and no longer passes resume). The repo is a generated artifact, so we port and close rather than merge. Please pull the latest. Thanks for the sharp diagnosis!

PR #1285, opened 2026-05-18
feat(pulse/telegram): render markdown in messages via HTML parse mode

catchingknives:
## Why

The Telegram module sends `ctx.reply(text)` with no `parse_mode`, so the DA's `**bold**`, `*italic*`, `` `code` ``, code fences, and links arrive as literal characters on the phone. The existing TELEGRAM MODE OVERRIDE prompt worked around this by steering the DA toward plain prose — a render-side problem masquerading as a content-side restriction.

## Change

- New `mdToHtml` helper in `modules/markdown-html.ts` converts a useful subset of markdown to Telegram HTML: **bold**, *italic*, ~~strikethrough~~, `inline code`, fenced code, [links](url), ATX headers (→ bold lines), bullets (→ `•`).
- Telegram module sends with `parse_mode: "HTML"` by default; plain-text fallback on any send error via `.catch()`.
- New `markdown_mode: "html" | "plain"` field in `TelegramConfig`, defaults to `"html"` — opt-out preserved for anyone who wants raw text.
- TELEGRAM MODE OVERRIDE prompt flipped from "no formatting" to an inline cheatsheet of what renders, so the DA writes naturally-formatted prose.

## Robustness choices

- **Code blocks are stash-and-restored** before other transforms — bold/italic cannot leak into them.
- **Italic boundaries** admit punctuation, whitespace, and HTML tag chars so `**a _b_**` nests correctly while `snake_case` and `Array<T>` survive untouched.
- **Unbalanced markers** degrade to literal characters, so chunk-splitting at the 4096-char boundary cannot produce broken HTML.

## Tests

`Releases/v5.0.0/.claude/PAI/PULSE/modules/telegram.test.ts` — 21 cases covering snake_case safety, nested formatting, HTML escaping, code-block isolation, unbalanced markers, URLs with underscores, multiple inline spans.

```
$ bun test ./Releases/v5.0.0/.claude/PAI/PULSE/modules/telegram.test.ts
21 pass, 0 fail
```

## End-to-end validation

Showcase message exercising every transformation sent through a real Telegram bot; renders correctly on iOS.

## Relationship to other open PRs

- **#1284** (the empty-response fix) — independent of this PR. Different code path (SDK invocation vs send path). No conflict.
- **#1281** (`v2 — per-thread sessions, live tool reactions, empty-response recovery`) — no functional overlap on markdown rendering. #1281 sets `parse_mode: "Markdown"` on its `/help`, `/status`, and `/threads` command-output text only, not on the DA's actual replies, and uses the legacy Telegram Markdown mode (no nesting, no underscores-in-identifiers safety). This PR adds proper HTML rendering of the DA's prose with a regex-tested converter and a plain-text fallback. The two PRs **do** share `modules/telegram.ts` and would file-conflict on merge — whichever lands second will need a rebase, but the conceptual scopes are orthogonal.

### DIFF
--- a/Releases/v5.0.0/.claude/PAI/PULSE/modules/markdown-html.ts
+++ b/Releases/v5.0.0/.claude/PAI/PULSE/modules/markdown-html.ts
@@ -1,0 +1,60 @@
+/**
+ * Markdown → Telegram HTML converter.
+ *
+ * Telegram's parse_mode "HTML" supports a small tag set: b, i, u, s, code,
+ * pre, a, tg-spoiler. It does not render headers, lists, or tables. This
+ * helper converts a useful subset of common markdown emitted by the DA into
+ * that tag set so prose renders cleanly on phones.
+ *
+ * HTML special chars are escaped first so model-emitted '<', '>', '&' survive
+ * verbatim inside the message body.
+ *
+ * Design choices:
+ *  - Code blocks and inline code are stashed to placeholders BEFORE other
+ *    transforms run, so bold/italic/etc. cannot leak into code content.
+ *  - Italic markers (* and _) require punctuation/whitespace/tag-boundary on
+ *    both sides so snake_case identifiers and arithmetic stars are not
+ *    mangled, while still allowing italic to nest inside bold tags.
+ *  - Unbalanced markers render as literal characters — no broken HTML can be
+ *    emitted, so chunk-splitting before conversion is safe.
+ *  - Headers and bullets collapse to <b> lines and "• " glyphs respectively,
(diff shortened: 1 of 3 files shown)


danielmiessler:
Thanks @catchingknives, nicely built. The Telegram module was reworked since v5.0.0 and now deliberately steers to plain conversational prose for phone reading (with an egress sanitizer), so an HTML render layer would fight the current prompt contract. Closing as superseded rather than merging (the repo is generated from our source tree). If we revisit richer rendering, your escape-first mdToHtml is the reference. Thank you!

PR #1309, opened 2026-05-29
feat(statusline): honor user-configured location over IP geolocation (#843)

### LINKED ISSUE #843: feat(statusline): Support user-configured location for weather instead of IP geolocation only
First off — huge thanks to Daniel and all the contributors for PAI. It's been a great experience, and the work you all have put into this is really appreciated.

## Problem

The statusline determines location via IP geolocation (`ip-api.com`), which returns the wrong city for users behind a VPN, proxy, or corporate network. When the exit node is in a different city than the user's actual location, the statusline shows the wrong city and wrong weather.

`settings.json` already has a `location` object with `city`, `regionName`, `lat`, and `lon` fields, but the statusline ignores it — both the location display and the weather fetch rely entirely on IP geolocation, falling back to hardcoded San Francisco coordinates (37.7749, -122.4194) when the geolocation call fails.

## Feature Request

Allow users to set their location in `settings.json` and have the statusline prefer that over IP geolocation. Specifically:

1. **Location display** — If `settings.json` has `.location.city` and `.location.regionName`, use those instead of calling `ip-api.com`
2. **Weather fetch** — If `settings.json` has `.location.lat` and `.location.lon`, use those coordinates for the `open-meteo.com` API call instead of coordinates from the IP geolocation cache

IP geolocation would remain the fallback for users who don't configure a location.

## Current Workaround

I've patched my local `statusline-command.sh` in two places to read from `settings.json` first:

**Location display** (falls back to IP geolocation if no settings.json location):
```bash
if [ -f "$SETTINGS_FILE" ]; then
    settings_loc=$(jq -r 'if .location.city then .location | {city, regionName, country: "United States", lat, lon} else empty end' "$SETTINGS_FILE" 2>/dev/null)
fi
if [ -n "$settings_loc" ]; then
    echo "$settings_loc" > "$LOCATION_CACHE"
else
    loc_data=$(curl -s --max-time 2 "http://ip-api.com/json/?fields=city,regionName,country,lat,lon" 2>/dev/null)
    # ... existing IP geolocation logic
fi
```

**Weather fetch** (reads lat/lon from settings.json when location cache is empty — also avoids a race condition where location and weather parallel fetches mean weather can run before the location cache is written):
```bash
lat="" lon=""
if [ -f "$LOCATION_CACHE" ]; then
    lat=$(jq -r '.lat // empty' "$LOCATION_CACHE" 2>/dev/null)
    lon=$(jq -r '.lon // empty' "$LOCATION_CACHE" 2>/dev/null)
fi
# Fallback: read from settings.json
if [ -z "$lat" ] || [ -z "$lon" ]; then
    if [ -f "$SETTINGS_FILE" ]; then
        lat=$(jq -r '.location.lat // empty' "$SETTINGS_FILE" 2>/dev/null)
        lon=$(jq -r '.location.lon // empty' "$SETTINGS_FILE" 2>/dev/null)
    fi
fi
lat="${lat:-37.7749}"
lon="${lon:-122.4194}"
```

This works, but gets overwritten on every PAI upgrade since `statusline-command.sh` is a system file.

## Environment

- PAI v4.0.1
- macOS, VPN active

rpriven:
## Problem

The statusline determines location and weather purely via `ip-api.com` IP geolocation, which returns the wrong city for users behind a VPN, proxy, or corporate network whose exit node is in a different city — then falls back to hardcoded San Francisco coordinates. `settings.json` already has a `.location` object (`city`, `regionName`, `lat`, `lon`) but the statusline ignores it for both display and weather.

## Fix

Make `settings.json` `.location` authoritative when present:

- **Location display** uses the configured `city` / `regionName`
- **Weather** uses the configured `lat` / `lon`
- Optional `.location.countryCode` drives the flag emoji (falls back to 🌐 when unset)
- **`ip-api.com` is skipped entirely** when location is configured — correct for VPN/proxy users, *and* one fewer third-party request exposing the user's IP
- IP geolocation remains the fallback for unconfigured users (no behavior change for them)

The `.location.*` reads are `?`-guarded so a malformed user-edited `.location` (e.g. a string instead of an object) can't break the rest of the settings cache.

## Testing

Tested end-to-end on **Debian 13 / PAI v5.0.0**:

- ✅ Configured location **with** `countryCode` → shows configured city + flag, skips ip-api
- ✅ Configured location **without** `countryCode` → shows city + 🌐 fallback, skips ip-api
- ✅ Unconfigured → ip-api fallback works exactly as before (verified `countryCode` populated from ip-api)
- ✅ Partial (`city` only, no coords) → city displays, weather falls back gracefully (no crash)
- ✅ Malformed `.location` → guarded; `TEMP_UNIT` / counts / rest of settings cache intact

Closes #843. Thanks to @DAESA24 for the original analysis and workaround.

### DIFF
--- a/Releases/v5.0.0/.claude/PAI/statusline-command.sh
+++ b/Releases/v5.0.0/.claude/PAI/statusline-command.sh
@@ -43,7 +43,12 @@ else
       "work_count=" + (.counts.work // 0 | tostring) + "\n" +
       "sessions_count=" + (.counts.sessions // 0 | tostring) + "\n" +
       "research_count=" + (.counts.research // 0 | tostring) + "\n" +
-      "ratings_count=" + (.counts.ratings // 0 | tostring)
+      "ratings_count=" + (.counts.ratings // 0 | tostring) + "\n" +
+      "LOC_CITY=" + (.location.city? // "" | @sh) + "\n" +
+      "LOC_REGION=" + (.location.regionName? // "" | @sh) + "\n" +
+      "LOC_LAT=" + (.location.lat? // "" | tostring | @sh) + "\n" +
+      "LOC_LON=" + (.location.lon? // "" | tostring | @sh) + "\n" +
+      "LOC_CC=" + (.location.countryCode? // "" | @sh)
     ' "$SETTINGS_FILE" 2>/dev/null > "$_SETTINGS_CACHE"
     # shellcheck disable=SC1090
     source "$_SETTINGS_CACHE"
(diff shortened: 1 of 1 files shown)


danielmiessler:
Thanks @rpriven, your fix is already live in source and credited inline (upstream PR #1309): user-configured settings.json location is authoritative and ip-api is only the fallback. The repo is generated from our source tree, so we port and close rather than merge. Please pull the latest. Appreciated!

rpriven:
Thanks Daniel — appreciate the credit, and the port-and-close model makes sense now that the repo's generated.

One heads-up from checking the shipped payload after your note: as of today, main's `LifeOS/install/LIFEOS/LIFEOS_StatusLine.sh` doesn't seem to carry this fix yet — location still comes only from ip-api (there's no settings-location priority chain like the one timezone has at the top of the file), and I couldn't find the inline credit either. While verifying, I hit the same gap on two other ported fixes: the #1409 memory markers (templates on main still ship without `<!-- BEGIN ENTRIES -->`, and `parseFile()` returns empty when markers are missing — same fix independently re-submitted and closed in #1443), and the #1404 `$HOME` literals (`install/settings.system.json` lines 5/6/9, confirmed by another user on macOS yesterday).

Easy to reproduce the state: the GitHub blob SHAs for all five relevant files (`MemoryWriter.ts`, both memory templates, `settings.system.json`, `LIFEOS_StatusLine.sh`) are **identical between the v6.0.5 tag and main** — so none of these files has changed in the public tree since the 7/5 release, i.e. before these ports were closed as handled.

All four look like one mechanism rather than four separate misses: whatever generates the public payload from your source tree may be dropping some ported fixes. Figured one pointer to the pattern was more useful than four bug reports. Happy to re-verify any of these against the next release cut.

PR #1314, opened 2026-05-29
Security: stop leaking API credentials in Recon tools (URL token + process argv)

klausagnoletti:
## Summary

Two credential-exposure issues in `Packs/Security/src/Recon/Tools/`, found during a security audit of a PAI install.

### 1. `IpinfoClient.ts` - API token in URL query string
Both `lookup()` and `batchLookup()` built the request as `https://ipinfo.io/...?token=${this.apiKey}`. URL query strings are routinely captured in proxy logs, upstream/CDN access logs, and client history, so the ipinfo token leaks anywhere along the path. Fixed by sending it as an `Authorization: Bearer <token>` header (ipinfo.io supports this), which is not logged the same way.

### 2. `SubdomainEnum.ts` - API key in process arguments
`runChaos()` ran ``$`chaos -key ${key} -d ${domain} -silent` ``, putting `PDCP_API_KEY` in the process argument list where any local user/process can read it via `ps`/`/proc`. `chaos` already reads `PDCP_API_KEY` from the environment, so the key is now passed via `.env()` and dropped from argv.

## Impact
Credential disclosure (low-to-moderate): leaked API keys for ipinfo.io and ProjectDiscovery. No functional change to results.

## Testing
Behavior unchanged (same endpoints, same auth, same output). Header auth verified against ipinfo.io's documented `Authorization: Bearer` support; chaos verified to read `PDCP_API_KEY` from env.

## Note
A related `subprocess(shell=True)` in `WebAssessment/WebappScripts/with_server.py` was reviewed and left as-is: it executes the operator's own `--server` command at their existing shell privilege (needs `cd`/`&&`), so it is by-design rather than a vulnerability.

### DIFF
--- a/Packs/Security/src/Recon/Tools/IpinfoClient.ts
+++ b/Packs/Security/src/Recon/Tools/IpinfoClient.ts
@@ -83,10 +83,14 @@ export class IPInfoClient {
     // Rate limiting
     await this.rateLimit();
 
-    const url = `${this.baseUrl}/${ip}/json?token=${this.apiKey}`;
+    // Send the token via Authorization header, not the URL query string
+    // (URL params leak into proxy/server logs).
+    const url = `${this.baseUrl}/${ip}/json`;
 
     try {
-      const response = await fetch(url);
+      const response = await fetch(url, {
+        headers: { Authorization: `Bearer ${this.apiKey}` },
+      });
 
       if (!response.ok) {
         if (response.status === 429) {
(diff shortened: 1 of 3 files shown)


danielmiessler:
Thanks @klausagnoletti — good security eyes. The safe patterns here are already in the current source: IpinfoClient sends the token via the Authorization header rather than the URL query string, and SubdomainEnum passes the key to chaos through the environment rather than argv, so tokens do not land in proxy logs or the process list. The with_server.py shell=True is an operator-supplied local command at the same trust level as the shell invoking it, so there is no privilege boundary to cross there. Nothing left to change. This repo is generated from our source tree, so we close rather than merge. Please pull the latest. Appreciated!

PR #1315, opened 2026-05-29
feat(voice): add native macOS `say` TTS engine option

camerondgray:
### Summary

Adds an opt-in `voice_engine` setting to the Pulse voice module so PAI can speak through the native macOS `say` binary instead of the ElevenLabs cloud API. Setting `voice_engine = "say"` makes all voice notifications free, fully offline, and key-free. Default behavior is unchanged.

### 🎯 Motivation and Context

**Problem:**
- ElevenLabs TTS is metered — every notification costs credits, and hitting the quota cap silently breaks voice (a 401 with no fallback).
- Privacy: every notification's text is POSTed to `api.elevenlabs.io`. For a *personal* infrastructure tool, sending all assistant speech to a third party is undesirable for many users.
- It's currently the only engine — there is no offline/local option.

**Solution:**
Introduce a `voice_engine` setting (`"elevenlabs"` | `"say"`). The `"say"` engine routes TTS through `/usr/bin/say` — no API key, no network, nothing leaves the machine. ElevenLabs stays the default, so existing installs are unaffected.

### 📋 Changes

Single file: `Releases/v5.0.0/.claude/PAI/PULSE/VoiceServer/voice.ts`

- `VoiceConfig` gains `voice_engine?: "elevenlabs" | "say"` and `default_say_voice?`.
- `VoiceEntry` gains `sayVoice?` so each `daidentity.voices.<name>` can map to a macOS voice.
- New `speakWithSay()` + `sayRateFromSpeed()`; `playAudio()` refactored to share a `playAudioFile()` helper so the `say` path keeps the existing per-voice volume control.
- `sendNotification()` branches on the engine. The ElevenLabs path and the `/notify` HTTP contract are unchanged, so existing callers need no edits.
- `startVoice()` resolves the engine (config → `PAI_VOICE_ENGINE` env → default `elevenlabs`), warns on an unrecognized value, and only requires an API key for the ElevenLabs engine.
- `voiceHealth()` reports the active engine.

### ⚙️ Usage

```toml
# PULSE.toml
[voice]
enabled = true
voice_engine = "say"            # "elevenlabs" (default) | "say"
default_say_voice = "Samantha"  # optional; omit to use the system voice
```

Or via environment: `PAI_VOICE_ENGINE=say` (and optionally `PAI_SAY_VOICE=Samantha`).
Optional per-voice mapping in `settings.json`:

```json
"daidentity": { "voices": { "main": { "voiceId": "…", "sayVoice": "Samantha" } } }
```

Run `say -v '?'` to list installed voices; higher-quality voices can be added in **System Settings → Accessibility → Spoken Content**.

### ✅ Benefits

- **Free** — no ElevenLabs usage/credits.
- **Private** — notification text never leaves the machine.
- **Resilient** — no API key or quota dependency, and no network round-trip (lower latency).
- **Non-breaking** — default stays ElevenLabs; existing configs and all `/notify` callers are unaffected.

### 🧪 How Has This Been Tested?

- [x] `voice_engine="say"` speaks via `/notify` → `handleVoiceRequest` with **no API key set** (system voice and a named "Samantha" voice both verified audibly)
- [x] Default (unset) still resolves to `elevenlabs` — confirmed via `voiceHealth()`
- [x] `PAI_VOICE_ENGINE=say` environment selection works
- [x] Unrecognized engine value logs a warning and falls back to `elevenlabs`
- [x] Per-voice volume preserved (say → AIFF → `afplay -v`); pronunciation preprocessing applied on both engines

### 📊 Types of Changes

- [x] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

### ✅ Checklist

- [x] My code follows the PAI code style
- [x] I have tested this change thoroughly
- [x] This change is backward compatible
- [x] No shell interpolation — `say` is invoked via `spawn()` with an argument array, and `--` terminates option parsing so message text can never be treated as flags (consistent with the execSync→execFileSync hardening in #1046)

### 🖥️ Platform

macOS only (`say` is a macOS binary). On non-macOS hosts the engine should be left at the `elevenlabs` default; selecting `say` there fails gracefully through the existing voice-error path.

### DIFF
--- a/Releases/v5.0.0/.claude/PAI/PULSE/VoiceServer/voice.ts
+++ b/Releases/v5.0.0/.claude/PAI/PULSE/VoiceServer/voice.ts
@@ -1,10 +1,18 @@
 /**
  * PAI Pulse — Voice Module
  *
- * ElevenLabs TTS, macOS notifications, pronunciation preprocessing.
+ * Text-to-speech (ElevenLabs or native macOS `say`), macOS notifications,
+ * pronunciation preprocessing.
  * Absorbed from VoiceServer/server.ts into a Pulse-embeddable module.
  *
- * Config resolution (3-tier):
+ * TTS engines (select via `voice_engine` in the [voice] section of PULSE.toml,
+ * or the PAI_VOICE_ENGINE env var):
+ *   - "elevenlabs" (default) — cloud TTS, requires ELEVENLABS_API_KEY.
+ *   - "say"                  — native macOS `say` binary. Free, fully offline,
+ *                              no API key, and no notification text ever leaves
+ *                              the machine. macOS only.
+ *
+ * Config resolution (3-tier, applies to both engines):
  *   1. Caller sends voice_settings in request body → use directly (pass-through)
  *   2. Caller sends voice_id → look up in settings.json daidentity.voices → use those settings
  *   3. Neither → use settings.json daidentity.voices.main as default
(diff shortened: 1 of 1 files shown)


danielmiessler:
Thanks for building this, @camerondgray — it's a thoughtful contribution and clearly real work went into it.

After looking at it against where the system is now, I'm going to pass on merging it: appreciate the offline/free native-say TTS option; it's a voice-identity design call for me and needs re-porting onto the refactored Linux-aware VoiceServer. It's more of a product/direction call than a bug fix, and it isn't the direction I'm taking that part of the system right now.

PR #1321, opened 2026-05-30
fix(CostTracker): stop false BYPASS alerts on manifests + fix dead x-api-key detector

### LINKED ISSUE #1319: CostTracker fires a false 'cost alert / BYPASS' for package.json dependency declarations
## What you see

On a normal PAI install, the hourly CostTracker cron pops a macOS notification:

> PAI cost alert: 12 NEW API-risk call site(s) since baseline 1 call site(s)
> classified as BYPASS — review and patch

Because CostTracker exists to catch real Anthropic over-billing, this reads as
"money is leaking right now." It isn't.

## Steps to reproduce

1. A `package.json` under a scan root (`PULSE/`, `skills/`, …) declares
   `@anthropic-ai/*` as a dependency (the default install does).
2. No baseline snapshot exists yet (`bun CostTracker.ts baseline` never run).
3. Run `bun CostTracker.ts alert-check` (or wait for the cron).

## Expected vs actual

- **Expected:** no alert. A dependency *declaration* never executes or bills;
  the real SDK callers are guarded with `delete process.env.ANTHROPIC_API_KEY`.
- **Actual:** `package.json:<line>` (the `@anthropic-ai/...` line) is classified
  **BYPASS** → "review and patch"; and with no baseline, *every* site shows as
  "NEW since baseline."

## Root cause

`classifyCallSite()` special-cases `.md` as non-executable but not dependency
manifests, so the SDK-dependency string in `package.json` falls through to the
SDK→bypass branch. A missing baseline makes everything "NEW."

## Impact

A cost/security guardrail that cries wolf trains the user to ignore it — so the
day a *real* bypass fires, it gets dismissed. A false positive in a security
tool is worse than silence.

## Follow-ups (separate from the core fix)

- First-run UX: don't alarm "NEW since baseline" when no baseline exists —
  prompt the user to run `baseline` instead.
- Notification: multiple alerts render as one run-on line with no separator and
  no file path — hard to parse and not actionable.

Related false-negative (opposite direction): see the companion issue on the
broken `x-api-key` detector.

### LINKED ISSUE #1320: CostTracker never detects raw x-api-key/HTTP Anthropic calls — broken regex + unknown classification
## Summary

CostTracker's raw-HTTP `x-api-key` detector never fires. Two compounding bugs mean a genuine raw `x-api-key` HTTP call to Anthropic — the exact billing-bypass class the tool targets — is silently missed (the tool fails **open**).

## Bug 1 — broken regex (literal pipe, not alternation)

The pattern is:

    "x-api-key.*anthropic\\|x-api-key.*sk-ant"

In the regex handed to `rg`, `\|` is a **literal pipe**, not alternation. So it only matches a line literally containing `anthropic|x-api-key`, which never occurs in real code.

Reproduce:

    # current pattern — no match:
    printf 'headers["x-api-key"]="sk-ant-abc"\n' | rg -e 'x-api-key.*anthropic\|x-api-key.*sk-ant'
    # corrected pattern — matches:
    printf 'headers["x-api-key"]="sk-ant-abc"\n' | rg -e 'x-api-key.*(anthropic|sk-ant)'

## Bug 2 — raw-HTTP hits classify as `unknown`, never `bypass`

Even when a raw-HTTP pattern (`api.anthropic.com/v1/messages`) matches, the classifier routes only `--bare` and SDK reasons to `bypass`; raw-HTTP falls through to `unknown`, so it never raises an alert and goes silent after the first baseline.

## Impact

A **false negative** in a security/cost leak detector — more dangerous than the companion false-positive issue, because the failure is silent. A real raw `x-api-key` call to Anthropic bills the API and is never flagged.

## Fix

- Regex → `x-api-key.*(anthropic|sk-ant)` (proper alternation).
- Raw-HTTP patterns classify `bypass`.
- Regression tests assert the pattern matches `sk-ant`/`anthropic` lines and rejects a literal-pipe-only line.

ugiya:
## What you see (the symptom)

On a normal PAI install, the hourly CostTracker cron pops a macOS cost alert (#1319) for `package.json` dependency declarations, while a broken detector regex silently misses real raw `x-api-key` calls (#1320). This PR fixes both and hardens the classifier.

## What this changes

- **SDK-pattern-gated manifest exemption** — `package.json`/lockfile *dependency declarations* classify `legit`, but `--bare`/raw-HTTP inside a manifest's scripts still classify `bypass` (an earlier naive fix masked them).
- **Fix the `x-api-key` regex** — `\|` (literal pipe) → `(anthropic|sk-ant)` alternation; the detector was matching nothing.
- **Raw-HTTP patterns now classify `bypass`, not `unknown`.**
- **Stable `id`+`kind` per RISK_PATTERN; baseline keyed on `file:line:patternId`** so note/classification changes no longer churn the baseline ("phantom NEW").
- **Exact-basename manifest match** (not `endsWith`); drop dead `bun.lockb`; exclude config-with-scripts (`deno.json`/`composer.json`); `LEGIT_HINTS` win before the manifest check.
- **Add `CostTracker.test.ts`** (`bun test`, 13 cases) incl. the `--bare`-in-`package.json` regression, baseline-key stability, and the regex fix.

## Verification

`bun test` → 13 pass / 0 fail. Local `scan` → bypass 0; `alert-check` → no alerts.

## Not in this PR (follow-ups)

- First-run UX: don't alarm "NEW since baseline" when no baseline exists.
- Notification: multiple alerts render as one run-on line; separate them and include the offending file path.

Fixes #1319
Fixes #1320

### DIFF
--- a/Releases/v5.0.0/.claude/PAI/TOOLS/CostTracker.test.ts
+++ b/Releases/v5.0.0/.claude/PAI/TOOLS/CostTracker.test.ts
@@ -1,0 +1,135 @@
+import { describe, expect, test } from "bun:test";
+import { mkdtempSync, rmSync, writeFileSync } from "fs";
+import { tmpdir } from "os";
+import { join } from "path";
+import {
+  baselineKey,
+  classifyCallSite,
+  isDependencyManifest,
+  RISK_PATTERNS,
+  type CallSite,
+  type RiskPattern,
+} from "./CostTracker";
+
+function risk(id: string): RiskPattern {
+  const pattern = RISK_PATTERNS.find((candidate) => candidate.id === id);
+  if (!pattern) throw new Error(`Missing risk pattern fixture: ${id}`);
+  return pattern;
+}
+
+describe("CostTracker classification", () => {
+  test("SDK dependency in package.json is legit with manifest note", () => {
+    const result = classifyCallSite("PULSE/package.json", risk("claude-agent-sdk"));
+
+    expect(result.classification).toBe("legit");
+    expect(result.note).toContain("dependency manifest");
+  });
+
+  test("--bare in package.json is bypass, not manifest-exempt", () => {
+    const result = classifyCallSite("PULSE/package.json", risk("claude-bare"));
+
+    expect(result.classification).toBe("bypass");
+    expect(result.note).toContain("--bare");
+  });
+
+  test("raw x-api-key in package.json is bypass, not manifest-exempt", () => {
+    const result = classifyCallSite("PULSE/package.json", risk("raw-http-x-api-key"));
+
+    expect(result.classification).toBe("bypass");
+    expect(result.note).toContain("raw Anthropic HTTP");
+  });
+
+  test("Evals package.json SDK hit keeps the specific Evals legit note", () => {
+    const result = classifyCallSite("skills/Evals/package.json", risk("anthropic-sdk"));
+
+    expect(result.classification).toBe("legit");
+    expect(result.note).toContain("Evals");
+    expect(result.note).toContain("EVALS_ALLOW_API_BILLING");
+  });
+
+  test("mypackage.json is not manifest-exempt by suffix", () => {
+    const result = classifyCallSite("PULSE/mypackage.json", risk("anthropic-sdk"));
+
+    expect(isDependencyManifest("PULSE/mypackage.json")).toBe(false);
+    expect(result.classification).toBe("bypass");
+  });
+
+  test("deno.json with --bare remains bypass", () => {
+    const result = classifyCallSite("PULSE/deno.json", risk("claude-bare"));
+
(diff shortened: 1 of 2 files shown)


danielmiessler:
Thank you — you nailed both halves of this. The `x-api-key.*anthropic\|x-api-key.*sk-ant` pattern was being read by ripgrep as a **literal pipe**, so the detector matched nothing; and `package.json` dependency lines were tripping a permanent false BYPASS because a manifest can't carry the env-scrub guard.

Both are now fixed in source: the alternation is `x-api-key.*(anthropic|sk-ant)`, and `.json` manifests are classified as non-executable call sites (verified: `PULSE/package.json` no longer flags as BYPASS). It ships in the next release — the public repo is generated from our source tree, so we port community fixes upstream and let the release carry them rather than merging directly. Really appreciate the precise report. 🙏

PR #1328, opened 2026-06-04
fix: resolve PROJECTS.md to correct USER/PROJECTS/ path in Telos dashboard

rfdiazpr:
## Problem

`PROJECTS.md` lives at `PAI/USER/PROJECTS/PROJECTS.md`, not inside `PAI/USER/TELOS/`. Both `getAllTelosData()` in `telos-data.ts` and the save route in `api/file/save/route.ts` were resolving it through `TELOS_DIR`, so clicking a project card in the Telos dashboard returned an empty/missing file.

## Fix

Add a `resolveTelosFilePath()` helper in both files that routes `PROJECTS.md` requests to `USER/PROJECTS/PROJECTS.md` and all other files through `TELOS_DIR` as before.

Also aligns the `TELOS_DIR` constant in `save/route.ts` with `telos-data.ts` — it was still pointing to the old `life/telos` path.

## Files changed

- `Releases/v2.5/.claude/skills/Telos/DashboardTemplate/Lib/telos-data.ts`
- `Releases/v2.5/.claude/skills/Telos/DashboardTemplate/App/api/file/save/route.ts`

### DIFF
--- a/Releases/v2.5/.claude/skills/Telos/DashboardTemplate/App/api/file/save/route.ts
+++ b/Releases/v2.5/.claude/skills/Telos/DashboardTemplate/App/api/file/save/route.ts
@@ -3,7 +3,13 @@ import fs from 'fs'
 import path from 'path'
 import os from 'os'
 
-const TELOS_DIR = path.join(os.homedir(), '.claude/skills/life/telos')
+const TELOS_DIR = path.join(os.homedir(), '.claude/skills/PAI/USER/TELOS')
+const PROJECTS_FILE = path.join(os.homedir(), '.claude/skills/PAI/USER/PROJECTS/PROJECTS.md')
+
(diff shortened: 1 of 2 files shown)


rfdiazpr (commented):
Daniel,
This was suggested to me by PAI.
The bug (PROJECTS.md served from wrong path) is a PAI infra bug. You could file an issue or PR at danielmiessler/Personal_AI_Infrastructure describing the TELOS_DIR vs USER/PROJECTS/ mismatch so others who build a Pulse dashboard hit the right path
  
The upstream has the same bug PROJECTS listed as a core file but getAllTelosData only reads from TELOS_DIR. The fix applies there too.

danielmiessler:
Hey @rfdiazpr, thank you for the PR and the work that went into it.

PR #1330, opened 2026-06-05
fix(TOOLS): frontmatter parser drops block-list tags/related — KnowledgeGraph edges silently dead

Tokyofloripa:
# fix(TOOLS): frontmatter parser drops YAML block-list `tags:`/`related:` → KnowledgeGraph tag & related edges are silently dead

## TL;DR
On any real knowledge archive, `KnowledgeGraph.ts` reports **`tag: 0` and `related: 0` edges** — only body `[[wikilinks]]` ever connect. Root cause: the hand-rolled frontmatter parser **skips indented lines**, so the YAML *block-list* form of `tags:` and `related:` (which is what the Knowledge skill actually writes) is never parsed. The same defect kills `MemoryRetriever.ts`'s tag-match bonus. Two-function fix per file; weights, BFS, and CLI behavior untouched.

## Reproduction (v5.0.0, before)
```
$ bun .claude/PAI/TOOLS/KnowledgeGraph.ts stats
  Nodes: 120
  Edges: 304 (tag: 0, wikilink: 304, related: 0)   ← tag & related edge classes empty
  Isolated nodes: 16
```
A note whose frontmatter is the standard block-list form:
```yaml
tags:
  - area/healthcare
  - role/physician
related:
  - "[[albert-einstein-hospital]]"
```
shows `Tags: []` and **"No direct connections found"** in `related <slug>` — its weight-5 `related` edges and weight-1 tag edges never form.

## Root cause
1. **`parseFrontmatter()`** (`KnowledgeGraph.ts` ~L88, `MemoryRetriever.ts` ~L98) does `if (line.startsWith("  ") || line.startsWith("\t")) continue;`. The indented `  - item` lines of a block list are skipped, so `tags:`/`related:` resolve to empty → `node.tags = []` → **zero tag co-occurrence edges**, and notes render `type: unknown`.
2. **`extractRelated()`** (`KnowledgeGraph.ts`) only recognizes the *structured* `- slug:` / `type:` form. Real notes use wikilink list items (`- "[[slug]]"`), so **zero `related` edges** form.

Net: of the three documented edge classes (related=5, wikilink=3, tag=1), only wikilink (body-scanned) works. The graph layer is effectively non-functional for the note format the Knowledge skill itself produces.

## The fix (scope-contained — only the two parser functions)
- **`parseFrontmatter()`**: parse YAML block-list values (`key:` followed by indented `- item` lines) into a `string[]`, stripping `- ` and matched quotes; inline `[a, b]` arrays and quoted scalars still parse; no lowercasing (callers handle it).
- **`extractRelated()`**: a `normalizeSlug()` helper extracts the target from all four item forms — `- "[[slug]]"`, `- [[slug]]`, `- "slug"`, `- slug` — unwrapping `[[…]]`, dropping `|alias`, keeping the last `/` segment; the legacy `- slug:`/`type:` structured form still works.
- `buildGraph`, BFS/`traverse`, edge weights (5/3/1), `TAG_GROUP_CAP`, and every CLI verb are **unchanged**.

## Testing evidence (after)
```
$ bun .claude/PAI/TOOLS/KnowledgeGraph.ts stats
  Nodes: 120
  Edges: 10672 (tag: 9936, wikilink: 304, related: 432)   ← both classes live; wikilink unchanged
  Isolated nodes: 2
```
- `related <slug>` now resolves the note's real typed `related:` links and its tags.
- **`MemoryRetriever.ts`**: a query whose term appears *only* in a note's `tags:` (absent from the body) now retrieves it at `score: 5.0` (= `TAG_MATCH_WEIGHT`) — impossible before (tags parsed to `[]`).
- Correctness checks: a note *without* a `tags:` field still shows `Tags: []` (no over-match / contamination); top tag clusters are real facets, not parser artifacts; `git diff` hunks are confined to `parseFrontmatter` + `extractRelated`; all CLI verbs exit 0; node count unchanged (skip-files honored).

## Impact
Affects every PAI user who lets the Knowledge skill write notes (the default block-list frontmatter). Before this, `KnowledgeGraph traverse/related/hubs/find` and `MemoryRetriever`'s tag bonus operate on empty tag/related data; the "edges > nodes" health expectation is unreachable. After, the graph and tag-aware retrieval work as documented.

## Optional follow-up (not in this PR)
`KnowledgeGraph.ts` and `MemoryRetriever.ts` carry **two independent hand-rolled frontmatter parsers** — the divergence is what let this bug live in one place. A future cleanup could extract a single shared `parseFrontmatter` so the next fix touches one implementation.

## Provenance
Found while building a 2,000+ note domain knowledge cluster, where the dead tag/related edges made graph traversal useless — the parser fix was the precondition for the graph layer to function at all. Patches: `KnowledgeGraph.patch`, `MemoryRetriever.patch` (apply cleanly to `Releases/v5.0.0/.claude/PAI/TOOLS/`, verified via `git apply --check`).

### DIFF
--- a/Releases/v5.0.0/.claude/PAI/TOOLS/KnowledgeGraph.ts
+++ b/Releases/v5.0.0/.claude/PAI/TOOLS/KnowledgeGraph.ts
@@ -80,20 +80,44 @@ function parseFrontmatter(content: string): Record<string, any> {
   const match = content.match(/^---\n([\s\S]*?)\n---/);
   if (!match) return {};
   const result: Record<string, any> = {};
-  for (const line of match[1].split("\n")) {
+  const lines = match[1].split("\n");
+  for (let i = 0; i < lines.length; i++) {
+    const line = lines[i];
+    if (line.startsWith("  ") || line.startsWith("\t")) continue;
     const colonIdx = line.indexOf(":");
     if (colonIdx > 0) {
       const key = line.substring(0, colonIdx).trim();
-      // Skip indented lines (YAML nested content handled separately)
-      if (line.startsWith("  ") || line.startsWith("\t")) continue;
       let value: any = line.substring(colonIdx + 1).trim();
-      if (value.startsWith("[") && value.endsWith("]")) {
+      if (value.length === 0) {
+        const items: string[] = [];
+        let j = i + 1;
+        while (j < lines.length && (lines[j].startsWith("  ") || lines[j].startsWith("\t"))) {
+          const itemLine = lines[j].trim();
+          if (itemLine.startsWith("-")) {
+            let item = itemLine.replace(/^-\s*/, "");
+            const quotedItem = item.match(/^\s*(["'])([\s\S]*)\1\s*$/);
+            if (quotedItem) {
+              item = quotedItem[2];
+            }
+            item = item.trim();
+            if (item.length > 0) {
+              items.push(item);
+            }
+          }
+          j++;
+        }
+        if (items.length > 0) {
+          value = items;
+          i = j - 1;
+        }
+      }
+      if (typeof value === "string" && value.startsWith("[") && value.endsWith("]")) {
         value = value
           .slice(1, -1)
           .split(",")
           .map((s: string) => s.trim().replace(/['"]/g, ""))
           .filter((s: string) => s.length > 0);
-      } else if (value.startsWith('"') && value.endsWith('"')) {
+      } else if (typeof value === "string" && value.startsWith('"') && value.endsWith('"')) {
         value = value.slice(1, -1);
       }
       result[key] = value;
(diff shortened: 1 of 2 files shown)


danielmiessler:
Thank you — solid find. The frontmatter parser wasn't collecting YAML **block-list** items, so `related:` / `tags:` written as an indented `- item` block parsed to an empty string, and the related-slug retrieval boost never fired on kb-v3 notes.

Fixed in source: the parser now collects block-list entries and extracts the slug from typed `- slug: x` items (skipping the `type:` continuation lines so no spurious keys leak). Verified against a kb-v3-shaped note — `related` and `tags` both populate correctly. Ships next release. Appreciate it. 🙏

Issue #1324, opened 2026-06-03
pai CLI shim references PAI/Tools/ but install ships PAI/TOOLS/ — breaks on Linux


NorthwoodsSentinel:
# `pai` CLI shim references `PAI/Tools/` but install ships `PAI/TOOLS/` — breaks on Linux

## Summary

The `pai` command-line shim at `/usr/local/bin/pai` and the bashrc alias both reference `${PAI_ROOT}/PAI/Tools/pai.ts` (mixed case). The installed PAI tree ships the tools directory as `PAI/TOOLS/` (all uppercase). Linux is case-sensitive — the shim doesn't resolve. macOS users don't see this because HFS+ and APFS default to case-insensitive.

## Affects

PAI v5.0.0 on any case-sensitive filesystem (Linux distros, including WSL2 on Windows). All Linux installs are at minimum suspect; reproduced on Lares running PAI 5.0.0 (WSL2 Ubuntu).

## Reproduction

```bash
ls /root/.claude/PAI/TOOLS/pai.ts    # exists
ls /root/.claude/PAI/Tools/pai.ts    # ENOENT — no such file or directory

cat /usr/local/bin/pai
# Output: #!/bin/bash
#         exec bun "/root/.claude/PAI/Tools/pai.ts" "$@"
# (or similar — references the lowercase path)

grep 'pai' ~/.bashrc | grep -i alias
# Output: alias pai='bun /root/.claude/PAI/Tools/pai.ts'

pai
# bun: error opening /root/.claude/PAI/Tools/pai.ts: No such file or directory
# (or similar — depends on which shim/alias was tried first)
```

## Local workaround

```bash
ln -s TOOLS /root/.claude/PAI/Tools
```

A same-name symlink at the lowercase path makes both casings resolve. Tested working on Lares (Linux, WSL2 Ubuntu) on 2026-05-21.

## Root cause

The PAI source/release tooling produces a `TOOLS/` directory but the installer-templated shim and bashrc alias hardcode `Tools/`. Both layers exist in the public release; the inconsistency was probably introduced when the source tree directory was renamed (or the shim was written) without updating both places.

Filesystem semantics:
- macOS HFS+/APFS default: case-insensitive (preserves case in metadata, lookups are case-insensitive). `Tools/` and `TOOLS/` resolve to the same entry.
- Linux ext4/btrfs/zfs/xfs: case-sensitive. `Tools/` and `TOOLS/` are distinct.
[...]

PR #1336, opened 2026-06-08
Align Packs/ source PAI/Tools → PAI/TOOLS to match shipped v5 runtime dir (case-sensitive Linux)

Oliver-Liu99:
## Summary

Rewrites the **10 remaining `Packs/` source files** that still reference the runtime tools directory with the legacy lowercase `PAI/Tools/` so they match the canonical all-caps `PAI/TOOLS/` used everywhere else in v5. This is the **source-side half of #1324**.

## Root cause (corrects the issue framing)

I investigated #1324 and the shipped release is **not** actually broken on Linux today:

- The installed artifact is **not** a `/usr/local/bin/pai` shim — it's a single `alias pai='bun .../PAI/TOOLS/pai.ts'` line, and the v5 installer already emits the correct **`TOOLS`** casing (and strips stale aliases on reinstall).
- The entire `Releases/v5.0.0/` tree is self-consistent on `PAI/TOOLS/` (zero lowercase `PAI/Tools/` references inside it). The shipped graders import `../../../../PAI/TOOLS/Inference.ts` and the Telos chat route spawns `~/.claude/PAI/TOOLS/Inference.ts`.

The real defect is a **source/release divergence**: the casing convention changed v4 (`PAI/Tools`) → v5 (`PAI/TOOLS`), the release builder rewrites `Tools`→`TOOLS` on packaging (so shipped output is correct), but **10 upstream `Packs/` source files were never back-ported**. On case-insensitive macOS/APFS every path resolves to the one physical dir, masking it; on case-sensitive Linux/ext4 (incl. WSL2) the lowercase references fail to resolve when the source is run directly.

## Changes (12 references, 10 files — all `PAI/Tools/` → `PAI/TOOLS/`)

- **Executable** (would break at runtime on Linux): `Packs/Telos/.../api/chat/route.ts` (spawns `Inference.ts`); 3 Evals model-graders import `PAI/TOOLS/Inference` (`LLMRubric`, `NaturalLanguageAssert`, `PairwiseComparison`).
- **Copy-paste commands** (break on Linux paste): `PAIUpgrade/Workflows/{Upgrade,AlgorithmUpgrade}.md`, `Media/.../Art/Workflows/{Essay,Mermaid,Visualize}.md`.
- **Doc/changelog consistency**: `Utilities/src/Fabric/SKILL.md`.

## Scope / safety

- Only the literal **`PAI/Tools/`** runtime-tools token is rewritten. The three legitimately capital-T `Tools/` families — skill-level `<Skill>/Tools/`, `PAI/DOCUMENTATION/Tools/`, and the repo-root `Tools/` — are **intentionally untouched**.
- **No directory rename / `git mv`** needed: the physical dir is already `TOOLS`. Verified case-exact via `git ls-files` against `Releases/v5.0.0/` (immune to the macOS case-insensitivity that hides this bug). After the change, `grep -rn 'PAI/Tools/' Packs/` is empty.
- Result resolves correctly on **both** case-insensitive macOS and case-sensitive Linux, since canonical `TOOLS` already matches the physical dir.

## Suggested follow-ups (intentionally not bundled here, to keep this PR focused)

1. **CI grep-guard**: fail the build if `grep -rn 'PAI/Tools/' Packs/ Releases/$LATEST/` is non-empty, to stop the source/release divergence from silently reappearing.
2. **Durable upstream fix**: the private release-builder (`skills/_PAI/Tools/ShadowRelease.ts`, not in this public tree) currently relies on a build-time `Tools`→`TOOLS` rewrite; back-porting these `Packs/` edits makes the source canonical so the build no longer has to paper over it.
3. `.pai-protected.json` could add a `Releases/*/.claude/PAI/TOOLS/*.ts` glob (the v5 runtime tools dir is currently unprotected by name).

Refs #1324

### DIFF
--- a/Packs/Media/src/Art/Workflows/Essay.md
+++ b/Packs/Media/src/Art/Workflows/Essay.md
@@ -513,10 +513,10 @@ For non-blog images that only need transparency, or to remove backgrounds after 
 
 ```bash
 # Use the Images Skill for background removal
-bun ~/.claude/PAI/Tools/RemoveBg.ts /path/to/output.png
+bun ~/.claude/PAI/TOOLS/RemoveBg.ts /path/to/output.png
 
 # Or batch process multiple images
-bun ~/.claude/PAI/Tools/RemoveBg.ts image1.png image2.png image3.png
+bun ~/.claude/PAI/TOOLS/RemoveBg.ts image1.png image2.png image3.png
 ```
 
 **See:** `~/.claude/skills/Images/Workflows/BackgroundRemoval.md` for full documentation.
--- a/Packs/Media/src/Art/Workflows/Mermaid.md
+++ b/Packs/Media/src/Art/Workflows/Mermaid.md
@@ -669,7 +669,7 @@ GOING INTO BLOG/WEBSITE: Remove background for transparency
 **For blog/website use** — use the **Images skill** for background removal:
 
 ```bash
-bun ~/.claude/PAI/Tools/RemoveBg.ts /path/to/mermaid-diagram.png
+bun ~/.claude/PAI/TOOLS/RemoveBg.ts /path/to/mermaid-diagram.png
 ```
 
 **See:** `~/.claude/skills/Images/Workflows/BackgroundRemoval.md` for full documentation.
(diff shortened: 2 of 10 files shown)


danielmiessler:
Hey @Oliver-Liu99, thank you for the PR and the work that went into it.

PR #1342, opened 2026-06-10
feat: Add Claude Fable 5 model pricing and display name to Performance dashboard

### LINKED ISSUE #1341: Performance dashboard: Claude Fable 5 sessions costed at Sonnet pricing and shown as raw model id
The Performance dashboard doesn't know about Claude Fable 5 yet. Two problems:

- **Costs are wrong.** `MODEL_PRICING` has no `claude-fable-5` entry, so `getPricing()` falls through to the Sonnet default ($3/$15). Fable is $10/$50 — Fable-heavy sessions get under-costed about 3.3x.
- **Display is broken.** `shortModel()` only knows opus/haiku/sonnet, so Fable sessions show up as a truncated raw id like `claude-fable-5[1m]` instead of "Fable".

One detail that matters: transcripts carry suffixed ids (`claude-fable-5[1m]`), so a table entry alone won't fix it — the fuzzy match needs a `fable` branch too.

The fix is small (PR incoming):

- Add `"claude-fable-5": { input: 10, output: 50, cacheWrite: 12.50, cacheRead: 1.00 }` to `MODEL_PRICING`
- Add a `fable` branch to the fuzzy match
- Add the "Fable" display name to `shortModel()`

Tested in a sandbox with a synthetic Fable transcript — costs came out exact ($10.00 for 1M input, $5.00 for 100K output, and so on down the line).

Two things I hit while testing, separate from this fix — I can file them if you want them tracked:

- First run on a completely fresh system crashes: `saveState()` writes into `PAI/PULSE/Performance/`, but nothing ever creates that directory
- The Opus rows still carry $15/$75 pricing; current Opus 4.x is $5/$25

asdf8675309:
Fixes #1341

Fable 5 sessions were getting costed at Sonnet rates and showing up as a raw model id. Three small changes:

- `MODEL_PRICING`: add `claude-fable-5` ($10/$50; cache write 12.50, cache read 1.00 — same ratios as the rest of the table)
- `getPricing()`: add a `fable` fuzzy-match branch so suffixed ids like `claude-fable-5[1m]` price right instead of falling through to Sonnet
- `shortModel()`: show "Fable" on the dashboard

Testing: ran it on macOS (Bun) against a sandboxed fresh HOME with a synthetic Fable transcript. Exit 0, no errors, and the costs match hand math exactly — 1M input → $10.00, 100K output → $5.00, 200K cache write → $2.50, 4M cache read → $4.00. The dashboard maps the id to "Fable".

6 added lines across two files in the v5.0.0 bundle. Nothing platform-specific.

### DIFF
--- a/Releases/v5.0.0/.claude/PAI/PULSE/Observability/src/app/performance/page.tsx
+++ b/Releases/v5.0.0/.claude/PAI/PULSE/Observability/src/app/performance/page.tsx
@@ -102,6 +102,7 @@ function formatTokens(n: number): string {
 }
 
 function shortModel(m: string): string {
+  if (m.includes("fable")) return "Fable";
   if (m.includes("opus")) return "Opus";
   if (m.includes("haiku")) return "Haiku";
   if (m.includes("sonnet")) return "Sonnet";
--- a/Releases/v5.0.0/.claude/PAI/PULSE/Performance/cost-aggregator.ts
+++ b/Releases/v5.0.0/.claude/PAI/PULSE/Performance/cost-aggregator.ts
@@ -21,6 +21,8 @@ const STATE_FILE = join(PAI_DIR, "PULSE", "Performance", "aggregator-state.json"
 
 // Model pricing per million tokens (as of 2026-04)
 const MODEL_PRICING: Record<string, { input: number; output: number; cacheWrite: number; cacheRead: number }> = {
+  // Fable 5
+  "claude-fable-5": { input: 10, output: 50, cacheWrite: 12.50, cacheRead: 1.00 },
   // Opus 4 / 4.6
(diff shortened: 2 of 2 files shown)


danielmiessler:
Thank you — and this turned out to be bigger than just adding Fable. The pricing table was stale enough that `opus` was fuzzy-matching to the old **$15/$75** Opus-4 rate (Opus 4.8 is $5/$25), and Fable fell through to Sonnet pricing entirely.

Refreshed the whole `MODEL_PRICING` table to current models — **Fable 5 $10/$50, Opus 4.8 $5/$25, Sonnet 5 $3/$15, Haiku 4.5 $1/$5** — with the fuzzy fallbacks now pointing at the current tier in each family (and Fable/Mythos added). Verified pricing for all four. Fixed in source, ships next release. Appreciated. 🙏

PR #1350, opened 2026-06-14
ISA: add Deliverable Framing guidance for outward-artifact ISAs

Resistor52:
## What

Adds a short **Deliverable Framing** subsection to `Packs/ISA/src/SKILL.md` for the case where an ISA's ideal state is an outward-facing artifact (document, email, post, deck, report) rather than code or infrastructure.

## Why

One Principle keeps paying off for artifact-type ISAs: author the durable substance (facts, analysis) separately from the framing layer (voice, positioning, calls to action) aimed at a specific audience. When the two are entangled in the first draft, a late correction to the audience or purpose forces a rewrite; when they are kept separate, it is a light revision of the wrapper. The subsection also names the three inputs (recipient and role, prior visibility, artifact type) that, when assumed rather than settled in OBSERVE, cause the entanglement.

## Scope

Purely additive: a new subsection placed after the Three-Guardrail Taxonomy. It does not modify the locked twelve-section table or any existing section. Placement and wording are easy to change; happy to move it, trim it, or reword if you'd prefer it lighter or in a different spot.

### DIFF
--- a/Packs/ISA/src/SKILL.md
+++ b/Packs/ISA/src/SKILL.md
@@ -67,6 +67,22 @@ The first three are author-stated (declarative). Anti-criteria are derived — t
 
 ---
 
+## Deliverable Framing (outward-artifact ISAs)
+
+When an ISA's ideal state is an outward-facing artifact (a document, email, post, deck, report) rather than code or infrastructure, one Principle pays for itself repeatedly: author the durable substance (the facts and analysis) separately from the framing layer (the voice, positioning, and calls to action) aimed at a specific audience.
+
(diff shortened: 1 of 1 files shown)


danielmiessler:
Thanks @Resistor52 — this was a thoughtful proposal and the substance-vs-framing separation for outward artifacts is a real insight. We went a different way with ISA doctrine, so we are not adopting it as a core section, but it was a good idea and exactly the kind of engagement we want. This repo is generated from our private source tree each release, so we close rather than merge here. Appreciate you digging into the ISA design at this depth.

PR #1370, opened 2026-06-20
fix(SatisfactionCapture): don't fabricate a neutral 5 when there's no real rating

lewta:
## Problem

`SatisfactionCapture.hook.ts` writes a fabricated `rating: 5` whenever it lacks a
real satisfaction signal, in three places:

1. Inference call **returns failure** → wrote `{rating:5,"Inference failed…"}`.
2. Inference call **throws** → wrote `{rating:5,"Inference error…"}`.
3. Inference **succeeds but the model omits/garbles the rating** →
   `const rating = (… valid …) ? r.rating : 5;` quietly substituted 5.

A fabricated 5 is "we don't know" masquerading as "neutral." These rows
accumulate and flatten every today/week/month average — the aggregator
(`PAI/statusline-command.sh` ~L1422-1438) flat-means `.rating` and never reads
`.confidence`, so low-confidence placeholders still count at full weight. In one
real install, **238 of 881 signals (27%) were `"Inference failed"` placeholders.**

## Fix

Introduce `isValidRating(x)` as the single source of truth (a real number in
1-10). Persist a rating only when it's valid; in every other case log for
observability and **write nothing**. The success path no longer coerces a
missing rating to 5. No behavior change when the model returns a valid rating.

## Alternatives considered (and why skipping wins)

- **Lower the confidence on the placeholder row** — no effect. The aggregator
  flat-means `.rating` and ignores `.confidence`; the failure rows are already
  written at `confidence: 0.3` and still skewed every average.
- **Write a sentinel rating excluded from averages** — would require teaching
  every rating consumer (the statusline jq, `LearningPatternSynthesis.ts`, …) to
  recognize and exclude it: a larger, multi-file change for no added benefit.

Skipping keeps the change small while preserving observability — failures are
still logged to stderr. If failures should be queryable, the right home is a
separate failure stream, not the ratings signal.

## Testing evidence

Two tests under `Releases/v5.0.0/.claude/hooks/tests/`:

- `isValidRating.test.ts` — unit test of the validity guard (accepts 1-10;
  rejects null/undefined/0/11/NaN/Infinity/strings/objects). `bun test` → 2 pass.
- `SatisfactionCapture.failure-path.test.sh` — behavioral: forces a deterministic
  inference failure (removes `claude` from `PATH`), runs the hook against a
  throwaway temp `PAI_DIR` (never touches a real `~/.claude`), asserts no row
  written. Before/after on identical input:

```
[PRE-FIX ] inference failure → wrote 1 row {"rating":5,"sentiment_summary":"Inference failed: …"}
[POST-FIX] inference failure → wrote 0 rows; stderr: "… skipping signal (no rating written)"
→ PASS (exit 0)
```

### DIFF
--- a/Releases/v5.0.0/.claude/hooks/SatisfactionCapture.hook.ts
+++ b/Releases/v5.0.0/.claude/hooks/SatisfactionCapture.hook.ts
@@ -460,30 +460,17 @@ async function main() {
           }
         }
       } else {
-        // Inference failed — default to 5 (neutral)
+        // Inference failed — skip the signal instead of fabricating a neutral 5.
+        // A failed classification is not a real satisfaction signal; writing a
+        // default-5 row pollutes ratings.jsonl and drags every running average
+        // toward the mushy middle. Log for observability, write nothing.
         const errorReason = result.error || 'unknown';
-        console.error(`[SatisfactionCapture] Inference failed: ${errorReason} — defaulting to 5`);
-        writeRating({
-          timestamp: getISOTimestamp(),
-          rating: 5,
-          session_id: sessionId,
-          source: 'implicit',
-          sentiment_summary: `Inference failed: ${errorReason.slice(0, 80)}`,
-          confidence: 0.3,
-        });
+        console.error(`[SatisfactionCapture] Inference failed: ${errorReason} — skipping signal (no rating written)`);
       }
     } catch (err) {
-      // Inference errored — default to 5 (neutral)
+      // Inference errored — skip the signal (see note above). Do not fabricate a 5.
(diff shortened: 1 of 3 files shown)


danielmiessler:
Thanks @lewta, the "absence of a signal is not a neutral 5" fix (the isValidRating guard and skip-instead-of-write paths) is already in the current source and ships in the latest release. This repo is a generated artifact, rebuilt from our private source tree each release, so we port fixes into source and close rather than merge here. Please pull the latest. Closing with thanks!

lewta:
Thank you for the feedback @danielmiessler !

PR #1371, opened 2026-06-20
fix(learning-readback): add a freshness window so the digest can't show stale signals

lewta:
## Problem

`loadLearningDigest()` injects "Recent Learning Signals" into every session's
context. It calls `getRecentLearnings()`, which returns the N most-recent entries
per category (ALGORITHM / SYSTEM) — but with **no date shown and no age cutoff**.

So when a category is quiet, the digest surfaces weeks-old entries as though they
were current, with nothing to signal their age. In a real install this presented
signals from 18+ days prior at session start, indistinguishable from today's.

## Fix

- `parseLearningDate(filename)` — extracts the timestamp from the
  `YYYY-MM-DD-HHMMSS_LEARNING_*.md` name (single source of truth).
- `isFresh(filename, maxAgeDays, now?)` — true only within the window;
  unparseable names are treated as not-fresh (a signal we can't date is not shown
  as current).
- `getRecentLearnings()` now filters by a 21-day freshness window and stamps each
  entry with its date (`[7/10] (2026-06-20) …`), matching the dated convention
  already used by `loadFailurePatterns()`.

No change when entries are recent; a quiet category now correctly shows fewer (or
no) entries rather than stale ones.

## Testing evidence

`Releases/v5.0.0/.claude/hooks/tests/learning-readback.test.ts` — `bun test` → 3 pass:
- `parseLearningDate` extracts the date and rejects malformed names.
- `isFresh` includes recent, excludes old and unparseable.
- `loadLearningDigest` against temp fixtures (a today entry + a 60-day-old entry)
  shows the fresh one and hides the stale one.

### DIFF
--- a/Releases/v5.0.0/.claude/hooks/lib/learning-readback.ts
+++ b/Releases/v5.0.0/.claude/hooks/lib/learning-readback.ts
@@ -26,12 +26,41 @@
 import { readFileSync, existsSync, readdirSync } from 'fs';
 import { join } from 'path';
 
+/** Default freshness window for the learning digest. Entries older than this are
+ *  excluded so a quiet category can't surface weeks-old signals as if current. */
+const DEFAULT_MAX_AGE_DAYS = 21;
+
 /**
- * Read the N most recent learning files from a LEARNING subdirectory.
- * Files are named YYYY-MM-DD-HHMMSS_LEARNING_*.md with YAML frontmatter.
- * Extracts the **Feedback:** line and rating for compact display.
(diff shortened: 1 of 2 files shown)


danielmiessler:
Thanks @lewta, the learning-readback freshness window (21-day default, date-stamped lines) is in the current source, tests included, and ships in the latest release. The repo is generated from our source tree each release, so we port and close rather than merge. Please pull the latest. Thanks for a well-tested PR!

lewta:
Thank you for the feedback @danielmiessler !

PR #1375, opened 2026-06-23
fix: add pai-logo.png to Pulse Observability public/ dir

klausagnoletti:
## What changed
Added `pai-logo.png` to `Releases/v5.0.0/.claude/PAI/PULSE/Observability/public/` and added a `.gitignore` re-include rule so it is tracked.

## Why
The logo was missing from the Next.js static assets folder (`public/`), causing a broken image in the Pulse dashboard header on fresh installs. The file already existed at `PAI-Install/public/assets/pai-logo.png` but was never placed where Next.js expects it at build time.

Without this fix, every fresh PAI v5 install gets a broken logo. The workaround is to manually copy the file and then copy it again to `out/` (since the dashboard is a static export), but neither survives a `bun run build`.

## Root cause
The `.claude/.gitignore` blanket-ignores `*.png` and only re-includes specific paths. `Observability/public/` was not among them.

## Tested on
Windows 11 / WSL2

## Checklist
- [x] No personal data
- [x] Works without user-specific credentials
- [x] .gitignore re-include added so file survives future releases


danielmiessler:
Thanks a lot for this, @klausagnoletti — really appreciate you flagging it and taking the time to write the fix.

When I went back through the current source, this turns out to already be handled: logo already renamed to lifeos-logo.png, committed, and un-ignored in current source — the broken-header fix is live. The system has moved a fair bit since you opened this (the PAI → LifeOS rename and a few subsystem rewrites), so the gap you spotted has since been closed independently.

PR #1379, opened 2026-06-24
Add Codex as a first-class PAI harness

simonasrazm:


danielmiessler:
Thank you for this, @simonasrazm — and sorry it sat open so long.

The codebase changed a lot since you opened it (the PAI → LifeOS rename, path restructures, and a shift to a generated single-skill public artifact), and this one no longer applies to the current tree: codex is already a first-class harness — the install was redesigned to be AI-native so any harness (Codex included) runs it, making the build-time adapter/rewrite approach unnecessary.

PR #1384, opened 2026-06-30
feat(pulse/telegram): bidirectional image support

klausagnoletti:
# Pulse Telegram bridge: bidirectional image support

## Summary

The Pulse Telegram module (`PAI/PULSE/modules/telegram.ts`) is text-only today: a
single `bot.on("message:text")` handler, no media in or out. This adds **bidirectional
image support** so you can send your DA a photo or PNG/JPEG from your phone and have it
actually look at the image, and so the DA can send images back. Implemented and running
in production on my instance.

## Motivation

The most natural thing to do from a phone is send a screenshot or a photo. Today the
bridge silently drops any non-text message (the `message:photo` / `message:document`
update has no handler, so grammY ignores it and the user gets no reply). Outbound, the
DA can generate or fetch an image but has no way to deliver it.

### Inbound (photo or PNG/JPEG document)

A new `bot.on(["message:photo", "message:document"])` handler downloads the file, then
hands the local path to the SDK session, which uses the `Read` tool (vision) to view it.
Captions flow through and are sanitised + injection-scanned exactly like text messages.

Because the saved path is handed to a `bypassPermissions` session, inbound bytes are
validated **before** they ever reach the decoder:

- **PNG and JPEG only**, identified by **magic bytes**, never the client-declared
  `mime_type` (which is attacker-controllable).
- Everything else is rejected: WebP, GIF, SVG, voice, audio, video, stickers, and all
  non-image documents. Refusing to decode those formats is what retires the libwebp
  (CVE-2023-4863) and SVG-script attack classes: a decoder you never invoke can't hurt you.
- **10 MB** byte cap, plus a **40 MP / 10000 px** pre-decode dimension cap. Documents are
  not re-encoded by Telegram, so a tiny file can claim a gigapixel canvas (decompression
  bomb); the dimension cap is read cheaply from the PNG IHDR / JPEG SOF header before any
  full decode or allocation.
- Saved under a dedicated `state/telegram/incoming/` dir with a `randomUUID()` filename
  and the extension derived from the sniffed type (never the remote path). Best-effort
  deleted after the session reads it.

### Outbound (DA to user)

The DA includes a line `[[IMG:/absolute/path]]` (or `[[IMG:https://url]]`) anywhere in
its reply. The bridge extracts the refs, strips the tags from the visible text (including
during the live streaming edits), and sends each as a photo or document. Documented to the
model via two lines added to the existing `TELEGRAM MODE OVERRIDE` system-prompt block.

### Refactor (no behaviour change to text)

To avoid duplicating ~140 lines of SDK/stream/reply logic across the text and image
handlers, the body of the existing `message:text` handler is extracted verbatim into a
shared `processPrompt(ctx, { userLog, newMessage })`. Both handlers call it. The text path
is unchanged: history-building, session-resume, billing key-strip, timeout, chunking, and
persistence all identical.

### Security boundary

No change to the trust boundary. The `allowed_users` middleware still gates every update,
including media. The threat model is a compromised owner account or an owner forwarding a
malicious file, not the open internet.

## Patch

One file changed (`Releases/v5.0.0/.claude/PAI/PULSE/modules/telegram.ts`), +214 / -47,
parses clean (`bun build`). The PNG/JPEG sniff and dimension logic is unit-tested against
real files and a synthetic decompression-bomb header. Tested and running in production on
my instance: photos, captioned photos, PNG/JPEG documents, and outbound generated images
all round-trip.

## Residual risk (stated plainly)

Decoding any image still invokes the host image decoder, so libpng/libjpeg memory-safety
exposure remains on attacker-influenced (compromised-owner) input. It is bounded by the
magic-byte allowlist, the 10 MB / 40 MP / 10000 px pre-decode caps, the sandboxed
Read-only consumer, and the single-user access gate. The libwebp and SVG classes are fully
retired by never decoding those formats. Operators handling untrusted forwards should keep
host image libraries patched.

## Out of scope (deliberately)

- Voice / audio / video inbound. Those need a transcription/frame pipeline (Whisper, ffmpeg)
  the bridge doesn't have; a focused follow-up PR is the right home for them.
- A separate one-line robustness fix I run locally (passing `pathToClaudeCodeExecutable` to
  the SDK so it works on non-native installs) is NOT in this diff; filable separately if useful.

### DIFF
--- a/Releases/v5.0.0/.claude/PAI/PULSE/modules/telegram.ts
+++ b/Releases/v5.0.0/.claude/PAI/PULSE/modules/telegram.ts
@@ -8,12 +8,12 @@
  * Architecture: grammY polling → auth → SDK session → stream → Telegram
  */
 
-import { Bot } from "grammy"
+import { Bot, InputFile } from "grammy"
 import { query } from "@anthropic-ai/claude-agent-sdk"
 import { ConversationStore } from "../lib/conversation"
 import { sanitize, analyzeForInjection } from "../lib/sanitize"
 import { join } from "path"
-import { appendFile, mkdir } from "fs/promises"
+import { appendFile, mkdir, unlink } from "fs/promises"
 
 // BILLING: Strip ANTHROPIC_API_KEY before any SDK query() call. Bun auto-loads
 // ~/.claude/.env into this process; if the key is present, @anthropic-ai/claude-agent-sdk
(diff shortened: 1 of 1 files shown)


danielmiessler:
Thanks @klausagnoletti, this was a genuinely missing capability and a thoughtful design (local-save + Read for inbound, the [[IMG:]] tag protocol with live-stream stripping for outbound). It is ported onto the current dynamic-voice module with a security pass: inbound writes are confined to the state dir with UUID names, a path-traversal guard, magic-byte PNG/JPEG sniffing, and size/pixel caps; captions run through the same injection scanner as text; the Daniel-only sender lock and egress sanitizer are preserved; and outbound paths are validated before send. Build and logic tests pass; it ships in the next release, credited to you. The repo is generated from our private source tree, so we port fixes into source and close rather than merge here. Thank you for the contribution!

PR #1397, opened 2026-07-03
fix(pulse): use LIFEOS (all-caps) runtime dir in observability.ts

### LINKED ISSUE #1391: v6.0.0 Linux: LifeOS vs LIFEOS case mismatch crash-loops Pulse on case-sensitive filesystems (+ stray dir side effect)
### Summary
`PULSE/Observability/observability.ts` hardcodes the runtime dir as `~/.claude/LifeOS` (mixed case) while `DeployCore` deploys the runtime to ALL-CAPS `~/.claude/LIFEOS`. On case-sensitive filesystems (Linux) Pulse crash-loops with `ENOENT` on first start — and as a side effect the crash loop mkdir's a stray REAL `~/.claude/LifeOS/` directory (it captures `PULSE/state/pulse.pid`), which then shadows any later symlink fix.

On macOS (default case-insensitive APFS) both paths resolve to the same dir, which is why this only bites Linux.

### Environment
Ubuntu 24.04, LifeOS v6.0.0, Pulse via `manage.sh install` (systemd user unit).

### Repro
1. Fresh install on Linux, Setup toolchain, `manage.sh install`.
2. Unit enters `activating (auto-restart)`; `pulse-stderr.log` shows repeated `{"msg":"Pulse crashed","error":"ENOENT: No such file or directory"}`.
3. `ls ~/.claude/` now shows BOTH `LIFEOS/` and a stray `LifeOS/` real dir created by the crash loop.

### Workaround (verified)
Stop the unit, remove the stray dir, `ln -s ~/.claude/LIFEOS ~/.claude/LifeOS`, start → Pulse boots clean (all modules load, port 31337 binds).

### Fix suggestion
Use the same constant the deployer uses (`LIFEOS`, or a shared resolve helper — `DeployComponents.resolveLifeosDir` already tolerates both). Happy to send a small PR.

badosanjos:
Fixes #1391.

`DeployCore` deploys the runtime to `~/.claude/LIFEOS` (all-caps), but `observability.ts` hardcodes `~/.claude/LifeOS` in three places (const `LIFEOS_DIR`, dashboard dir helper, and an env-fallback). On case-sensitive filesystems (Linux) Pulse crash-loops with `ENOENT` on first start — and the crash loop mkdir's a stray real `~/.claude/LifeOS/` directory (captures `pulse.pid`) that then shadows any later symlink workaround. macOS is unaffected only because APFS defaults to case-insensitive.

**Fix:** all three occurrences now use `LIFEOS`, matching the deployer.

**Verified on** Ubuntu 24.04: with paths aligned, Pulse boots clean under the shipped systemd user unit (`manage.sh install`) and binds 127.0.0.1:31337.

### DIFF
--- a/LifeOS/install/LifeOS/PULSE/Observability/observability.ts
+++ b/LifeOS/install/LifeOS/PULSE/Observability/observability.ts
@@ -55,7 +55,7 @@ export interface ObservabilityConfig {
 // ── Path Construction ──
 
 const HOME = process.env.HOME ?? ""
-const LIFEOS_DIR = join(HOME, ".claude", "LifeOS")
+const LIFEOS_DIR = join(HOME, ".claude", "LIFEOS")
 const MEMORY_DIR = join(LIFEOS_DIR, "MEMORY")
 
 const WORK_JSON_PATH = join(MEMORY_DIR, "STATE", "work.json")
@@ -142,7 +142,7 @@ function getDashboardDir(): string {
   const dir = config.dashboard_dir ?? DEFAULT_DASHBOARD_DIR
   // Resolve relative paths against Pulse directory
   if (!dir.startsWith("/")) {
-    return join(HOME, ".claude", "LifeOS", "PULSE", dir)
+    return join(HOME, ".claude", "LIFEOS", "PULSE", dir)
   }
   return dir
 }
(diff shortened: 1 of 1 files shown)


danielmiessler:
Thanks a lot for this, @badosanjos — really appreciate you flagging it and taking the time to write the fix.

When I went back through the current source, this turns out to already be handled: all-caps LIFEOS runtime dir is already the case in current source — thanks for catching the casing issue that has since been resolved. The system has moved a fair bit since you opened this (the PAI → LifeOS rename and a few subsystem rewrites), so the gap you spotted has since been closed independently.

PR #1399, opened 2026-07-03
fix(pulse): make effort badge tier-aware (E1–E5) via a shared resolver

krikauskis:
## What

Introduces `src/lib/effort-tier.ts` — a single `resolveEffortTier()` + tier class
maps — and rewires the three Observability renderers (`EffortBadge`,
`PhaseDetailPanel`, `UnifiedWorkDashboard`) to it, removing three copies of an
inline legacy quality→E-level map.

## Why

`work.json`'s `effort` field carries two vocabularies: a session-init placeholder
(`standard`/`starting`) and the real Algorithm tier written by ISASync (`E1`..`E5`).
The renderers only understood the legacy quality words, producing two bugs:

1. A real tier (`E3`/`E4`) has no map key → the badge renders **nothing**
   (`EffortBadge` returns `null`; the ladder lights nothing).
2. The not-yet-classified placeholder `standard` is mapped to `E1` → a session that
   was never classified shows a confident, false **"E1 STANDARD"** badge.

## Change

- New `src/lib/effort-tier.ts`: `resolveEffortTier(raw)` maps `E1..E5`
  (case-insensitive) → that tier; legacy `extended/advanced/deep/comprehensive` →
  their tier; everything else (incl. `standard`/`starting`) → a neutral **pending**
  state — never a false E1. Ships tier class maps (`TIER_BADGE_CLASSES`,
  `TIER_TEXT_CLASSES`, `EFFORT_TIERS`) as the single source of truth.
- `EffortBadge` / `PhaseDetailPanel` / `UnifiedWorkDashboard` consume the resolver;
  their inline maps are deleted (DRY — one source of truth). `EffortBadge` gains a
  `hidePending` opt to render nothing for placeholders instead of a pending chip.

## Testing

- Visual: a session with `effort="E3"` now shows an E3 badge (was blank); a session
  with `effort="standard"` shows a neutral pending chip (was a false "E1 STANDARD").
- `resolveEffortTier` is a pure function covering E1–E5 (any case), the four legacy
  words, and the never-false-E1 placeholder cases — straightforward to unit-test.

The only maintainer-taste call is cosmetic — whether a not-yet-classified session
shows a neutral "Pending" chip or renders nothing; `hidePending` supports both.

### DIFF
--- a/Releases/v5.0.0/.claude/PAI/PULSE/Observability/src/components/activity/EffortBadge.tsx
+++ b/Releases/v5.0.0/.claude/PAI/PULSE/Observability/src/components/activity/EffortBadge.tsx
@@ -1,26 +1,21 @@
-const EFFORT_CONFIG: Record<string, { bg: string; border: string; text: string; eLevel: string }> = {
-  standard: { bg: "bg-amber-500/15", border: "border-amber-500/30", text: "text-amber-400", eLevel: "E1" },
-  extended: { bg: "bg-orange-500/15", border: "border-orange-500/30", text: "text-orange-400", eLevel: "E2" },
-  advanced: { bg: "bg-rose-500/15", border: "border-rose-500/30", text: "text-rose-400", eLevel: "E3" },
-  deep: { bg: "bg-purple-500/15", border: "border-purple-500/30", text: "text-purple-400", eLevel: "E4" },
-  comprehensive: { bg: "bg-red-500/15", border: "border-red-500/30", text: "text-red-400", eLevel: "E5" },
-};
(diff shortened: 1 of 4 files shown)


danielmiessler:
Thanks a lot for this, @krikauskis — really appreciate you flagging it and taking the time to write the fix.

When I went back through the current source, this turns out to already be handled: effort badge is already tier-aware via a shared, backend-synced resolver (lib/effort.ts); the standard-as-placeholder change conflicts with our canonical Standard=E1 model. The system has moved a fair bit since you opened this (the PAI → LifeOS rename and a few subsystem rewrites), so the gap you spotted has since been closed independently.

PR #1403, opened 2026-07-03
fix(install): hand off to /LifeOS setup — /lifeos-setup is not a registered command

rpriven:
**What happens**

The installer's final step execs `claude "/lifeos-setup"`, but the shipped skill registers as **`LifeOS`** (`SKILL.md` → `name: LifeOS`, `argument-hint: [setup|…]`). Client-side command resolution rejects `/lifeos-setup` (and `/lifeos`) before skill routing ever sees it:

```
Unknown command: /lifeos-setup
```

Only **`/LifeOS setup`** works. So every fresh install currently ends with an error instead of onboarding, and a new user who doesn't guess the casing + argument form is stranded at the finish line. The fallback `printf` and the dry-run message send them to the same dead command.

**Validated:** clean Debian 13 container AND a separate Linux host with only the skill present (Claude Code 2.1.198) — deterministic, client-side. The currently-served `ourlifeos.ai/install.sh` has the same handoff, so it needs a redeploy after this merges.

**The fix**

`/lifeos-setup` → `/LifeOS setup` at all 5 occurrences (exec, dry-run message, fallback printf, two header comments), applied to both the repo installer and the nested skill-payload copy (`LifeOS/install/skills/LifeOS/install/install.sh`).

**Alternative** if you'd rather keep the published copy valid: register an explicit `lifeos-setup` alias command in the skill — the SKILL.md routing table already maps `/lifeos-setup` internally, but the client rejects unregistered commands before that routing applies. Happy to rework the PR that way instead.

Found during day-one Linux (Debian) runtime validation of v6.

### DIFF
--- a/LifeOS/install/install.sh
+++ b/LifeOS/install/install.sh
@@ -5,7 +5,7 @@
 #
 #   Unlike a whole-harness install, this does NOT clobber your setup.
 #   It drops the LifeOS skill into your existing harness, then hands off
-#   to the agentic `/lifeos-setup`, which (with your permission) does the
+#   to the agentic `/LifeOS setup`, which (with your permission) does the
 #   conflict detection, the principal conversation, the TELOS interview
 #   (current state + ideal state), pulls in any sources you provide, and
 #   wires hooks — adapting to YOUR OS and harness as it goes.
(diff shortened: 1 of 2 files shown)


rpriven:
Re-verified against v6.0.3 / current main: both install.sh files still call the unregistered `/lifeos-setup` (install.sh:143), and this PR still applies cleanly — GitHub reports MERGEABLE, no rebase needed after the 6.0.2/6.0.3 packaging churn. The new INSTALL.md flow routes agentic installs around the slash command, but `curl … | bash` remains the human entry point and still hands off to the dead command.

danielmiessler:
Thanks a lot for this, @rpriven — really appreciate you flagging it and taking the time to write the fix.

When I went back through the current source, this turns out to already be handled: thanks for catching the unregistered-command handoff — source already ships /LifeOS setup, so this is fixed upstream. The system has moved a fair bit since you opened this (the PAI → LifeOS rename and a few subsystem rewrites), so the gap you spotted has since been closed independently.

PR #1406, opened 2026-07-03
fix(linux): normalize runtime-root dir casing (LifeOS → LIFEOS) — silent memory failure on case-sensitive FS

christauff:
## Context (friendly heads-up: people are already deploying v6 on Linux!)

First — congrats on the 6.0.0 "one skill, one install" release. I picked it up within hours of the tag and installed it on **Linux** (Ubuntu), and it mostly came up beautifully. This PR reports one silent failure I hit during that deploy, with a fix. Sharing it partly because it's the kind of thing that only surfaces off-macOS, and you probably want to know early.

## The bug: `LifeOS` vs `LIFEOS` — a case-sensitivity silent failure on Linux

The runtime is deployed to the **all-caps** `<configRoot>/LIFEOS/` — `DeployCore` does this deliberately (its own comment: *"Targets the config-root runtime at the ALL-CAPS `<configRoot>/LIFEOS/` so it matches the `@LIFEOS/...` imports in CLAUDE.md"*), and `CLAUDE.template.md` imports `@LIFEOS/...`.

But **many runtime tools and hooks resolve the root as mixed-case** `~/.claude/LifeOS` — e.g. `const LIFEOS_DIR = join(HOME, ".claude", "LifeOS")`.

- On **macOS** (case-insensitive FS by default) `LifeOS` and `LIFEOS` are the same inode → everything works, invisible.
- On **Linux** (case-sensitive) they're different paths → the tools read a directory that doesn't exist and **silently return empty**.

### Repro (Linux)
```
$ bun ~/.claude/LIFEOS/TOOLS/MemoryStatus.ts
Corpus:
  knowledge   no dir yet        # ← but KNOWLEDGE/ exists at LIFEOS/MEMORY/KNOWLEDGE
```
It's looking at `~/.claude/LifeOS/MEMORY/KNOWLEDGE` (nonexistent on Linux) instead of `~/.claude/LIFEOS/MEMORY/KNOWLEDGE`. No error is thrown — the corpus just appears empty, which is the dangerous part. After normalizing the path casing, `MemoryStatus` correctly reports the corpus.

## The fix

Normalize the **runtime-root** path segment `"LifeOS"` → `"LIFEOS"` (the canonical `DeployCore`/`CLAUDE.md` already use) across the install-payload tools and hooks — **119 references across 91 files**.

- **No-op on macOS** (case-insensitive), a **fix on Linux** — strictly safe cross-platform.
- **Scoped carefully:** only touches runtime-root references (`.claude/LifeOS`, `CLAUDE_ROOT/"LifeOS"`, etc.). The legitimately mixed-case **`skills/LifeOS`** and **`install/LifeOS`** source-directory references are deliberately left untouched (verified).

## Notes / happy to adjust

- This is the minimal, mechanical normalization. Longer term a single shared root resolver (e.g. one `lifeosDir()` helper the tools import, instead of each re-deriving the path) would prevent the casing from drifting again — glad to follow up with that if you'd prefer it over the flat normalization.
- `PLATFORM.md` lists Linux as "Fully Supported," so this seemed worth fixing rather than working around locally (a `~/.claude/LifeOS → LIFEOS` symlink also masks it, but that's a band-aid).

Thanks for building and open-sourcing this — genuinely a pleasure to deploy.

### DIFF
--- a/LifeOS/install/LifeOS/PULSE/Observability/observability.ts
+++ b/LifeOS/install/LifeOS/PULSE/Observability/observability.ts
@@ -55,7 +55,7 @@ export interface ObservabilityConfig {
 // ── Path Construction ──
 
 const HOME = process.env.HOME ?? ""
-const LIFEOS_DIR = join(HOME, ".claude", "LifeOS")
+const LIFEOS_DIR = join(HOME, ".claude", "LIFEOS")
 const MEMORY_DIR = join(LIFEOS_DIR, "MEMORY")
 
 const WORK_JSON_PATH = join(MEMORY_DIR, "STATE", "work.json")
@@ -142,7 +142,7 @@ function getDashboardDir(): string {
   const dir = config.dashboard_dir ?? DEFAULT_DASHBOARD_DIR
   // Resolve relative paths against Pulse directory
   if (!dir.startsWith("/")) {
-    return join(HOME, ".claude", "LifeOS", "PULSE", dir)
+    return join(HOME, ".claude", "LIFEOS", "PULSE", dir)
   }
   return dir
 }
@@ -1626,7 +1626,7 @@ function readDirMdFiles(dir: string): { name: string, content: string, sections:
 
 function handleUserIndexApi(filter: string | null): Response {
   try {
-    const LIFEOS_DIR = process.env.LIFEOS_DIR || join(process.env.HOME || "", ".claude", "LifeOS")
+    const LIFEOS_DIR = process.env.LIFEOS_DIR || join(process.env.HOME || "", ".claude", "LIFEOS")
     const indexPath = join(LIFEOS_DIR, "PULSE", "state", "user-index.json")
     const raw = Bun.file(indexPath)
     if (!raw.size) {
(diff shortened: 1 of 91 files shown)


christauff:
Note for triage: the diff is a pure `LifeOS`→`LIFEOS` casing normalization (119 lines, all identical shape) — verified locally on Linux that it makes `MemoryStatus`/`MemoryRetriever` find the corpus without a symlink, and it's a no-op on macOS. If the `claude-review` check goes red, that's the usual fork-PR limitation (secrets/OIDC withheld from `pull_request` events off a fork), not the change.

danielmiessler:
Thanks a lot for this, @christauff — really appreciate you flagging it and taking the time to write the fix.

When I went back through the current source, this turns out to already be handled: linux path-casing fix already landed in source — every PULSE runtime-root literal is all-caps LIFEOS, so the next release ships it automatically. The system has moved a fair bit since you opened this (the PAI → LifeOS rename and a few subsystem rewrites), so the gap you spotted has since been closed independently.

Issue #1419, opened 2026-07-04
Pulse Assistant tab has no backend: pulse.ts imports missing ./Assistant/module, plus /da vs /assistant naming drift


BearGare:
## Summary

The Pulse **Assistant** dashboard tab is fully built on the frontend but has **no backend of any kind** in the shipped release. `pulse.ts` dynamically imports a module that does not exist anywhere in the repo, the import throws, the error is swallowed, and every `/assistant/*` request the tab makes goes unhandled. The result is a dead tab whenever `da.enabled` is true.

Investigating the cause surfaced a second, related problem: a **naming drift between the design doc and the shipped frontend** that should be reconciled before the backend is built.

## Root cause

`PULSE/pulse.ts` wires an assistant module that was never shipped:

- `pulse.ts:127` — `assistantModule = await import("./Assistant/module")` (file does not exist; the `catch` logs `"Assistant module not available"` and continues)
- `pulse.ts:356` — `assistantModule.assistantHealth()`
- `pulse.ts:476` — `assistantModule.startAssistant(config.da, config.jobs)`
- `pulse.ts:548-549` — routes `pathname.startsWith("/assistant/")` to `assistantModule.handleAssistantRequest(...)`

Because the import silently fails, `assistantModule` stays `null` and none of the routes below are ever registered.

Meanwhile the frontend (`PULSE/Observability/src/app/assistant/page.tsx`) is complete and actively polls these endpoints:

- `GET /assistant/identity` (page.tsx:218)
- `GET /assistant/health` (page.tsx:219)
- `GET /assistant/personality` (page.tsx:220)
- `GET /assistant/tasks` (page.tsx:221)
- `GET /assistant/diary` (page.tsx:222)
- `GET /assistant/opinions` (page.tsx:223)
- `GET /assistant/cron`, `POST /assistant/cron` (page.tsx:229, 306)
- `PATCH`/`DELETE /assistant/cron/{name}` (page.tsx:235, 244, 264)
- `PATCH /assistant/personality/traits` (page.tsx:318)

All of these currently resolve to nothing.

## Naming drift: `/da/*` vs `/assistant/*`

The DA Subsystem design doc (https://docs.ourlifeos.ai/Pulse__DaSubsystem, "Status: Architecture complete, pending implementation") specifies a **different** integration surface than the one the frontend was built against:

| | Design doc (DaSubsystem) | Shipped frontend + pulse.ts |
|---|---|---|
[...]

PR #1420, opened 2026-07-04
Fix three fresh-install bugs: missing yaml dep, TELOS.md drift, dead Assistant tab

BearGare:
Three independent fixes found during a fresh v6.0.0 install. Each is its own commit and can be reviewed/reverted independently.

## 1. Missing `yaml` dependency for deployed hooks/TOOLS scripts

Several deployed scripts (`hooks/lib/identity.ts`, `LIFEOS/TOOLS/Banner.ts`, `LoadSkillConfig.ts`, `PipelineOrchestrator.ts`, and others) import the `yaml` npm package, but the deploy pipeline never declares or installs it: there is no root `package.json` anywhere in the release tree, and neither `InstallEngine.ts` nor `DeployCore.ts` runs `bun install`. First run after a fresh install throws `Cannot find package 'yaml'`.

**Fix:** add `install/package.json` (with `yaml` pinned to the version already used by the Agents/Prompting skill tools) and a new `DeployCore.ts` step that copies it to `<configRoot>/package.json` and runs `bun install --cwd <configRoot>`, so `yaml` resolves via `node_modules` for both `<configRoot>/hooks/` and `<configRoot>/LIFEOS/`. Applied to both the top-level and the `skills/LifeOS`-vendored copy of `DeployCore.ts` (the repo carries the two byte-identical). Verified with dry-run and `--apply` against a scratch config root, and by importing `yaml` from a deployed `hooks/lib/*.ts` location.

## 2. TELOS.md vs legacy split-file drift

`TELOS.md`'s own banner says "this is the single source of truth... Pulse reads this file", and `GenerateTelosSummary.ts`'s `LEGACY_FILE_TO_SECTION` comment says the unified `TELOS.md` superseded the split files on 2026-05-01. But the shipped fresh-install template still carries **both** representations un-migrated, and the generator prefers the legacy per-topic file whenever it still exists on disk. Net effect: a fresh install silently ignores edits to `TELOS.md` in favor of the legacy files' stale samples, contradicting what `TELOS.md` tells the user.

Two other consumers read the legacy files with **no** `TELOS.md` fallback at all: `PULSE/checks/life-morning-brief.ts` and `skills/Daemon/Tools/DaemonAggregator.ts`. Archiving the legacy files without fixing these would have made them permanently blind to `TELOS.md`-only content.

**Fix:** archive the 8 legacy files to `USER/TELOS/Archive/2026-05-01/` (the fully-migrated state the generator already expects); rewrite `TELOS/README.md` to document `TELOS.md` as the single file to edit; add a `TELOS.md`-section fallback to both `life-morning-brief.ts` and `DaemonAggregator.ts`. Verified end-to-end against a scratch config root for the generator and both consumers.

## 3. Assistant module stub so the Pulse Assistant tab degrades honestly

`pulse.ts:127` dynamically imports `./Assistant/module` whenever `config.da.enabled` is true, and wires its exports to every `/assistant/*` route backing the fully-built Assistant tab (`Observability/src/app/assistant/page.tsx`). **That module never existed in the repo** — the import throws, is caught, logs `"Assistant module not available"`, and the entire tab API silently no-ops.

A full backend (identity/personality/diary persistence, cron CRUD) is a separate feature, tracked in #1419 (which also documents a `/da/*` vs `/assistant/*` naming drift between the design doc and the shipped frontend). **This PR only adds a stub** so the current no-op is deliberate rather than an accident of a caught import error: `GET /assistant/health` returns `identity_loaded: false` (making the frontend's existing `EmptyStateGuide`/`isFreshInstall` path deterministic), read endpoints return correctly-typed empty JSON, and mutating requests return `501` with a clear message instead of silently doing nothing when a user clicks Save/Delete. Verified via a standalone harness exercising every route.

Relates to #1419.


danielmiessler:
Thanks @BearGare, this was an unusually well-diagnosed PR and all three findings were real. Status: the missing yaml dependency and the morning-brief TELOS.md fallback are both in the current source and ship in the latest release, and the Assistant-tab crash was independently fixed in 6.1.2 (the private module is existence-checked before import, and fresh installs render a setup guide instead of erroring). Please pull the latest. One smaller piece remains on our list as a follow-up: archiving the legacy split TELOS files in the release template so TELOS.md is the single source of truth. Since the repo is generated from our source tree on each release we close rather than merge here, but your work directly shaped these fixes. Closing with thanks.

BearGare:
> Thanks @BearGare, this was an unusually well-diagnosed PR and all three findings were real. Status: the missing yaml dependency and the morning-brief TELOS.md fallback are both in the current source and ship in the latest release, and the Assistant-tab crash was independently fixed in 6.1.2 (the private module is existence-checked before import, and fresh installs render a setup guide instead of erroring). Please pull the latest. One smaller piece remains on our list as a follow-up: archiving the legacy split TELOS files in the release template so TELOS.md is the single source of truth. Since the repo is generated from our source tree on each release we close rather than merge here, but your work directly shaped these fixes. Closing with thanks.

Thank you so much, @danielmiessler  ; this genuinely made my week. This was my first public PR anywhere, and I went back and forth a fair bit before opening it: new to contributing, and honestly a little unsure whether I'd correctly diagnosed real bugs or was about to flag three things that would turn out to be my own misread. So hearing that all three findings were real, and getting a reply this thorough really meant a lot.

Good to know the yaml dep and the morning-brief TELOS.md fallback already ship in the latest, and that the Assistant-tab crash is existence-checked as of 6.1.2. I'll pull the latest. 

Really glad the write-ups were helpful in any way. Thanks again for taking the time; it's a great first experience of contributing to a public project.

PR #1421, opened 2026-07-04
Fix: explicit N/10 ratings silently dropped by SatisfactionCapture

rrittich:
## Problem

`parseExplicitRating()` in `LifeOS/install/hooks/SatisfactionCapture.hook.ts` returns `null` for the natural `N/10` rating form — `10/10`, `9/10, thanks`, `8 out of 10`. The reject-on-slash guard treats the `/` right after the leading number as invalid input:

```ts
const afterNumber = trimmed.slice(match[1].length);
if (afterNumber.length > 0 && /^[/.\dA-Za-z]/.test(afterNumber)) return null; // "10/10" hits this
```

So every explicit `N/10` rating silently falls through to the sentiment-inference path and is recorded with `source: "implicit"` — a guess — instead of the score the user actually gave. Bare-number and word ratings (`8`, `ten`) work; the fraction form does not.

## Fix

Add an `N/10` / `N out of 10` branch **before** the generic parser (ahead of the reject-on-slash guard), reusing a hoisted `SENTENCE_STARTERS` guard so a work phrase like `2/10 items done` is still correctly **not** treated as a rating.

## Behavior

| input | before | after |
|-------|--------|-------|
| `10/10, thank you` | `null` (→ implicit guess) | `{rating: 10, comment: "thank you"}` |
| `9/10` | `null` | `{rating: 9}` |
| `8 out of 10` | `null` | `{rating: 8}` |
| `2/10 items done` | `null` | `null` (guarded, not a rating) |
| `8`, `ten thanks` | unchanged | unchanged |

No change to existing bare-number or word-form ratings. Single-file, no new dependencies.

### DIFF
--- a/LifeOS/install/hooks/SatisfactionCapture.hook.ts
+++ b/LifeOS/install/hooks/SatisfactionCapture.hook.ts
@@ -69,6 +69,10 @@ const RATINGS_FILE = join(SIGNALS_DIR, 'ratings.jsonl');
 const LAST_RESPONSE_CACHE = join(BASE_DIR, 'MEMORY', 'STATE', 'last-response.txt');
 const MIN_PROMPT_LENGTH = 3;
 
+// Sentence-starters that mean a leading number is describing work, not rating it
+// (e.g. "2/10 items done", "3 of the files"). Shared by the fraction and generic parsers.
+const SENTENCE_STARTERS = /^(items?|things?|steps?|files?|lines?|bugs?|issues?|errors?|times?|minutes?|hours?|days?|seconds?|percent|%|th\b|st\b|nd\b|rd\b|of\b|in\b|at\b|to\b|the\b|a\b|an\b)/i;
+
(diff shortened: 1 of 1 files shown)


rrittich:
Discovered that when I wanted to give my DA a 10/10 rating, it didn't seem to elevate that to a preference in later chats. So I asked it to investigate and it discovered this issue in the PAI harness' logic. I'm not a programmer, and I have only a faint idea of what it is talking about here, but hopefully it helps and makes sense.

danielmiessler:
Thanks @rrittich — good catch. The fix for explicit N/10 ratings being silently dropped is in the current source now (fraction parse runs before the generic parser, with a sentence-starter guard so "2/10 items done" is not counted as a rating), and it ships in the next release. Since the repo is generated from our source tree we close rather than merge, but it is in and credited to you. Thanks again!

PR #1427, opened 2026-07-05
fix(memory): hot-layer entry markers — scaffold both, heal a missing BEGIN, migrate existing installs

### LINKED ISSUE #1409: v6.0.0: hot-layer memory templates lack <!-- BEGIN ENTRIES --> — reads return 0 entries, writes stack orphan END markers (memory never persists)
### Summary

The install templates for the two hot-layer memory files (`USER/PRINCIPAL/PRINCIPAL_MEMORY.md`, `USER/DIGITAL_ASSISTANT/DA_MEMORY.md` — see `skills/LifeOS/install/USER/...`) ship **without the `<!-- BEGIN ENTRIES -->` / `<!-- END ENTRIES -->` marker pair** that `LIFEOS/TOOLS/MemoryWriter.ts` requires. Result: the autonomic memory loop reports successful writes but **nothing is ever readable** — memory silently never persists on a fresh install.

### Mechanism

- `parseFile()` (MemoryWriter.ts) hits its markers-missing branch → whole body becomes `preEntriesBody`, `entries: []` → **every read returns 0 entries**.
- `serializeFile()`'s defensive branch prepends a lone `END` marker → **every write appends entries plus one more orphan `END`**; the next parse still finds no `BEGIN`, so the cycle repeats.

After a few reviewer runs our `PRINCIPAL_MEMORY.md` had accumulated 7 orphan `<!-- END ENTRIES -->` markers with unreadable entries stacked between them, while `read()` kept returning empty. No error surfaces anywhere — `setEntries` returns `ok` with `accepted=N`.

### Environment

Ubuntu 24.04, LifeOS v6.0.0.

### Repro

Fresh install → `bun LIFEOS/TOOLS/MemoryWriter.ts test` →

```
write 1: accepted=3, dropped_malformed=1, dropped_overlength=1, dropped_duplicates=1, evictions=0
FAIL: expected 3 entries on readback, got 0
```

(`MemorySystem.ts test` fails its "ISC-32: new entry present after curation" case for the same reason.)

### Fix suggestion (two layers, both verified locally)

1. Add the marker pair to both install templates:

```markdown
<!-- BEGIN ENTRIES -->
<!-- END ENTRIES -->
```

2. Self-heal in `serializeFile()` so a marker-less file converges to well-formed instead of degrading:

```ts
if (!pre.includes(BEGIN_MARKER)) {
  pre += BEGIN_MARKER + "\n";
}
```

With both applied, `MemoryWriter test`, `MemorySystem test`, and `MemoryReviewer test` all pass, and live curation persists entries. Happy to send a PR.

### LINKED ISSUE #1426: Hot-layer memory silently never loads: templates ship without entry markers, writer never re-adds a missing BEGIN
## Summary

The autonomic memory reviewer writes durable facts into the hot-layer files
(`USER/PRINCIPAL/PRINCIPAL_MEMORY.md`, `USER/DIGITAL_ASSISTANT/DA_MEMORY.md`),
but **none of them are ever injected back into a session** on a stock install.
The flagship "memory that compounds" loop captures correctly and recalls nothing.

Two bugs compound to cause it:

1. **The shipped templates contain no entry markers at all.** Both
   `install/USER/PRINCIPAL/PRINCIPAL_MEMORY.md` and
   `install/USER/DIGITAL_ASSISTANT/DA_MEMORY.md` ship with only the prose comment
   `<!-- Entries are appended automatically… -->` — **no `<!-- BEGIN ENTRIES -->`
   and no `<!-- END ENTRIES -->`.**
2. **`MemoryWriter.ts` self-heals a missing END marker but never a missing BEGIN.**
   `serializeFile` has `if (!post.startsWith(END_MARKER)) post = END_MARKER + post`
   with no symmetric guard for BEGIN. So every write to a marker-less file appends
   another `END` and never a `BEGIN`.

## Why it's silent and total

`LoadMemory.hook.ts` (the reader that injects memory each prompt) slices content
between the first `<!-- BEGIN ENTRIES -->` and the first `<!-- END ENTRIES -->`.
With no BEGIN, `startIdx === -1` → it returns `{ entries: [], count: 0 }` and
renders `(no entries yet)` — regardless of how many facts the reviewer has written.
`MemoryHealthCheck.ts` parses with the same marker contract, so it reports empty
too; nothing alarms. The three other readers (`PULSE/modules/memory.ts`,
`MemoryRestore.ts`) share the contract and also see nothing.

Meanwhile `MemoryWriter.parseFile`, on a marker-less file, treats the whole body as
`preEntriesBody` and returns `entries: []`, so dedup / the 48-entry cap / eviction
never run either — the file just accretes stray `END` markers on every write.

## Reproduction

Fresh install, then let the reviewer run a few times (or just inspect a stock file):

```bash
# The shipped template has neither marker:
grep -c 'BEGIN ENTRIES' ~/.claude/LIFEOS/USER/PRINCIPAL/PRINCIPAL_MEMORY.md   # → 0
grep -c 'END ENTRIES'   ~/.claude/LIFEOS/USER/PRINCIPAL/PRINCIPAL_MEMORY.md   # → 0

# After the reviewer has written N facts, the reader still injects zero:
echo '{}' | bun ~/.claude/hooks/LoadMemory.hook.ts
# → ## PRINCIPAL MEMORY [0/48 entries] (no entries yet)   ← despite N stored facts
```

On an install that has been running a while, the file shows the accretion signature:
`BEGIN=0`, `END=<number of writes>`, real `PREFERENCE:/RULE:/NAME:/RELATION:` lines
stranded outside any readable block.

## Impact

- **Universal:** every stock v6 install ships the marker-less templates + the
  non-healing writer, so the hot-layer memory feature is non-functional out of the box.
- **Platform-independent:** it's a template + a logic gap in shared TypeScript, not a
  Linux/casing issue.
- **Silent:** no error, health check blind to it — the assistant simply never seems to
  remember anything durable about the principal.

## Fix

PR attached. Three parts: (1) ship both markers in the templates (fresh installs);
(2) make `serializeFile` heal a missing BEGIN symmetrically (can't recur) and make
`parseFile` recover orphaned entries from an already-broken file; (3) a one-time
`MigrateMemoryMarkers` migration so existing installs repair on update. A regression
test covers the previously-untested "write in one session, read in the next" path.

rpriven:
Fixes #1409. Closes #1426 (my duplicate report — @badosanjos reported the same root cause first in #1409).

## The bug (two parts, one failure)

On a stock install the hot-layer memory files never load into a session. The reviewer
writes durable facts, but `LoadMemory.hook.ts` injects `(no entries yet)`. Root cause is
two compounding defects:

1. **Templates ship with no entry markers.** `install/USER/PRINCIPAL/PRINCIPAL_MEMORY.md`
   and `install/USER/DIGITAL_ASSISTANT/DA_MEMORY.md` had neither `<!-- BEGIN ENTRIES -->`
   nor `<!-- END ENTRIES -->`. The reader slices between those markers, finds no BEGIN,
   and returns empty.
2. **`serializeFile` heals a missing END but never a missing BEGIN.** So every write to a
   marker-less file appended another `END` and never a `BEGIN` — the file accreted stray
   END markers, read as empty forever, and dedup/cap/eviction never ran.

## The fix

**Fresh installs** — add both markers to the two shipped templates:

```
<!-- BEGIN ENTRIES -->
<!-- END ENTRIES -->
```

**Can't recur** — `serializeFile` now adds a missing `BEGIN_MARKER`, symmetric to the
existing `END_MARKER` guard.

**Existing data survives** — `parseFile`, on a marker-less/malformed file, now recovers
the real `PREFIX_PATTERN` entries from the body (dropping stray markers) instead of
discarding them, so the next write re-emits them inside a canonical block. The recovery
scanner skips lines inside code fences and blockquotes, so a documented `PREFERENCE: …`
example is never adopted as a real entry.

**Existing installs — self-heal, plus an on-demand accelerator.** Once a user updates and
has the fixed `MemoryWriter.ts`, a broken file **repairs automatically on the next reviewer
write** (the `parseFile` recovery above re-blocks the orphaned entries) — no manual step.
The Update workflow's `ScaffoldUser.ts copyMissing` deliberately never overwrites existing
USER files, so the template fix alone would not touch them; the writer fix is what covers
them. For an **immediate** repair without waiting for the next reviewer cycle, the new
standalone `MigrateMemoryMarkers.ts` fixes both files on demand — same invocation model as
the existing `MigrateContextFreshness` / `MigrateTelosFreshness` tools (run directly; those
are not auto-invoked by Update either). It's idempotent (canonical files skipped), backs up
before repair, supports `--dry-run`, and reuses the tested `read()`→`setEntries()` path.
Maintainers may optionally wire the migrations into the Update workflow as a separate
enhancement.

## Testing

Two runnable tests (no test runner is configured in the repo; run with an isolated `HOME`):

- `MemoryWriter.markers.test.ts` — 12 checks: fresh-template round-trip, the previously
  **untested cross-session path** (write in session A → read in session B), recovery of
  an already-corrupted file through the real add-flow, and the fence/blockquote guard.
- `MigrateMemoryMarkers.test.ts` — 9 checks: dry-run is side-effect-free, corrupted files
  heal (orphaned entries recovered), canonical files are left byte-identical, idempotent
  on re-run.

```bash
HOME=$(mktemp -d) bun LifeOS/install/LIFEOS/TOOLS/MemoryWriter.markers.test.ts
HOME=$(mktemp -d) bun LifeOS/install/LIFEOS/TOOLS/MigrateMemoryMarkers.test.ts
```

Both green. `MemoryWriter.ts` and `MigrateMemoryMarkers.ts` compile clean under
`bun build`. The happy path is provably unaffected: the new recovery branch only runs when
markers are missing/malformed, and the new BEGIN guard is a structural no-op on well-formed
files (their `preEntriesBody` already contains the marker).

## For users already on a broken install

After updating, existing memory auto-repairs on the next reviewer write. To fix immediately
without waiting, run the migration:

```bash
bun ~/.claude/LIFEOS/TOOLS/MigrateMemoryMarkers.ts        # --dry-run to preview
```

Or hand-edit both `~/.claude/LIFEOS/USER/PRINCIPAL/PRINCIPAL_MEMORY.md` and
`.../DIGITAL_ASSISTANT/DA_MEMORY.md`: put one `<!-- BEGIN ENTRIES -->` immediately above the
first entry and a single `<!-- END ENTRIES -->` after the last, removing any stray END
markers in between. No data is lost — the stranded entries are preserved, just re-blocked.

## Out of scope (noted, not fixed here)

- `parseFile` locates markers with `indexOf`, so prose that *mentions* the literal marker
  text before the real block could shadow entries. Pre-existing (reproduces on `main`
  un-patched); worth hardening to whole-line marker matching in a follow-up.
- `allowDrastic` has no live wiring anywhere — a legitimate full-clear is structurally
  unreachable. Orthogonal; separate follow-up.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/MemoryWriter.markers.test.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/MemoryWriter.markers.test.ts
@@ -1,0 +1,131 @@
+#!/usr/bin/env bun
+/**
+ * Regression test for the hot-layer memory marker bug.
+ * Exercises the never-tested path: a fact written in one "session" must be
+ * readable in the next. Run with an isolated HOME so it hits the real public
+ * API (setEntries/read) without touching the live tree:
+ *   HOME=/tmp/mem-test bun test-memory-markers.ts
+ */
+import { mkdirSync, writeFileSync, readFileSync, rmSync } from "node:fs";
+import { homedir } from "node:os";
+import { join } from "node:path";
+
+const WT = new URL(".", import.meta.url).pathname;
+const HOME = homedir();
+const USER = join(HOME, ".claude/LIFEOS/USER/PRINCIPAL");
+const OBS = join(HOME, ".claude/LIFEOS/MEMORY/OBSERVABILITY");
+const FILE = join(USER, "PRINCIPAL_MEMORY.md");
+
+const FIXED_TEMPLATE = `---
+schema_version: 1
+cap_entries: 48
+cap_chars_per_entry: 256
+last_updated: 2026-01-01
+last_updated_by: bootstrap-template
+convention: pai-freshness-v1
+---
+
+# Principal Hot-Layer Memory
+
+> Auto-curated hot-layer memory about you.
+
+<!-- Entries are appended automatically by the memory curation loop. Empty on a fresh install — this is expected. -->
+<!-- BEGIN ENTRIES -->
+<!-- END ENTRIES -->
+`;
+
+// A file corrupted by the OLD bug: no BEGIN, stray ENDs, orphaned entries.
+const CORRUPTED = `---
+schema_version: 1
+cap_entries: 48
+cap_chars_per_entry: 256
+last_updated: 2026-01-01
+last_updated_by: bootstrap-template
+convention: pai-freshness-v1
+---
+
+# Principal Hot-Layer Memory
+
+<!-- Entries are appended automatically by the memory curation loop. Empty on a fresh install — this is expected. -->
+PREFERENCE: existing fact one ~explicit
+<!-- END ENTRIES -->
+PREFERENCE: existing fact two ~inferred
+<!-- END ENTRIES -->
+`;
+
+const BEGIN = "<!-- BEGIN ENTRIES -->";
+const END = "<!-- END ENTRIES -->";
+
+// Mirror of LoadMemory.hook.ts readMemory(): the consumer contract.
+function readerSees(raw: string): string[] {
+  const s = raw.indexOf(BEGIN), e = raw.indexOf(END);
+  if (s === -1 || e === -1 || e < s) return [];
+  return raw.slice(s + BEGIN.length, e).trim().split("\n").map(l => l.trim()).filter(Boolean);
+}
+
+let failures = 0;
+const check = (name: string, cond: boolean, detail = "") => {
+  console.log(`${cond ? "✅" : "❌"} ${name}${cond ? "" : "  — " + detail}`);
+  if (!cond) failures++;
+};
+
+mkdirSync(USER, { recursive: true });
+mkdirSync(OBS, { recursive: true });
+mkdirSync(join(HOME, ".claude/LIFEOS/USER/DIGITAL_ASSISTANT"), { recursive: true });
+
+const { setEntries, read } = await import(join(WT, "MemoryWriter.ts"));
+
+// ── Scenario A: fresh (fixed) template — write in session 1, read in session 2 ──
+writeFileSync(FILE, FIXED_TEMPLATE);
+check("A0: fresh template starts empty to the reader", readerSees(readFileSync(FILE, "utf8")).length === 0);
+setEntries(FILE, ["PREFERENCE: session-1 wrote this ~explicit"]);
+const afterA = readFileSync(FILE, "utf8");
+check("A1: reader (LoadMemory contract) now sees the written fact", readerSees(afterA).includes("PREFERENCE: session-1 wrote this ~explicit"));
+check("A2: read() API returns 1 entry", (read(FILE) as any).count === 1);
+check("A3: exactly one BEGIN + one END", (afterA.match(/BEGIN ENTRIES/g)||[]).length === 1 && (afterA.match(/END ENTRIES/g)||[]).length === 1);
+
+// ── Scenario B: heal a file the OLD bug already corrupted ──
+// The real heal path: read() (parseFile) recovers orphaned entries, then
+// MemorySystem.add does read → append → setEntries (MemorySystem.ts:243).
+writeFileSync(FILE, CORRUPTED);
+check("B0: corrupted file reads as 0 to the consumer (the bug)", readerSees(CORRUPTED).length === 0);
+const recovered = (read(FILE) as any).entries as string[];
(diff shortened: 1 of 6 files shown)


danielmiessler:
Thanks a lot for this, @rpriven — really appreciate you flagging it and taking the time to write the fix.

When I went back through the current source, this turns out to already be handled: hot-layer memory marker heal already implemented independently in current source. The system has moved a fair bit since you opened this (the PAI → LifeOS rename and a few subsystem rewrites), so the gap you spotted has since been closed independently.

Issue #830, opened 2026-02-28
PAI does not work correctly when installed to a non-default CLAUDE_CONFIG_DIR


raychaser:
I use multiple `CLAUDE_CONFIG_DIR` locations to isolate different workflows. This has been working well, but PAI was effectively hardcoded to `~/.claude` — around 30 files used inline `process.env.PAI_DIR || join(homedir(), '.claude')` patterns that didn't resolve correctly when running from a non-default config directory.

I worked on fixing this and got it working. The fix centralizes all path resolution into a single `getPaiDir()` helper in `hooks/lib/paths.ts` that handles `$HOME`, `${HOME}`, and `~` expansion, with `homedir()` as the fallback. The rest of the changes are mechanical: swap the inline pattern for the helper, remove unused imports.

Here's the diff: https://github.com/danielmiessler/Personal_AI_Infrastructure/compare/main...raychaser:Personal_AI_Infrastructure:main

~48 files touched, though the changes are almost entirely mechanical (import swap + pattern replacement). Tested with a non-default `CLAUDE_CONFIG_DIR` — hooks fire correctly, statusline works, sessions track to the right directory.

I'd love to see this upstreamed but wanted to check if there's interest before opening a PR, given the size of the change.

Naturally, I found out only after doing all of this about https://github.com/danielmiessler/Personal_AI_Infrastructure/issues/694, https://github.com/danielmiessler/Personal_AI_Infrastructure/pull/778, and https://github.com/danielmiessler/Personal_AI_Infrastructure/pull/784. Should have done research first. Maybe a triumvirate can get it done? @Spirotot @neilinger My change is based on Release 4.0.1 but now has the 4.0.2 changes as well.

PR now live: #873

raychaser:
Update: opened #873 which ports the `getPaiDir()` migration to v4.0.3 (the previous PR #838 targeted v4.0.1). Same approach, now aligned with the latest release.

danielmiessler:
Hey @raychaser, thanks for raising this, and sorry it sat for a while.

We're changing how LifeOS ships. Instead of cloning a full `~/.claude` directory and running it as a complete system, LifeOS is becoming a skill you install through an agentic installer. The installer hands integration to your own AI, which reads your actual machine (your OS, your paths, your harness) and wires the hooks and system prompt in where they belong.

[...]

PR #1429, opened 2026-07-06
fix: honour CLAUDE_CONFIG_DIR — hooks and always-on tools work on non-default config roots

borfast:
## Problem

Claude Code supports multiple profiles via `CLAUDE_CONFIG_DIR`, and the agentic installer already detects it correctly (`DetectEnv.ts` reports the right `configRoot`) — but the rest of the pipeline ignores what it found. Installing v6.0.5 into a profile at `~/.claude-gc` produced:

- All 60 hook commands merged into `settings.json` pointing at `$HOME/.claude/hooks/...` — a location where this installer never deployed anything, so **every hook event errored** on the next session.
- Hooks and always-on tools that did run resolved their own paths back to `~/.claude`, reading the wrong profile's settings and **writing MEMORY state/observability files into a different profile** than the one that loaded them.

This is the same class of issue as #830/#284, but scoped to the new agentic-installer ship model rather than the old full-clone one — per the direction stated there: *"The installer hands integration to your own AI, which reads your actual machine (your OS, your paths, your harness) and wires the hooks."* This PR makes the shipped tooling deliver that for non-default config roots.

## Changes

1. **Installer substitution** — `InstallEngine.rewriteHooksConfigRoot()`: at merge time, `InstallHooks.ts` retargets the default-root spellings (`$HOME`/`${HOME}`/`~` + `/.claude`) in payload hook commands at the detected `configRoot`. **No-op when the config root is `~/.claude`, so default installs stay byte-identical.**
2. **Path resolution** — `hooks/lib/paths.ts` `getClaudeDir()` now falls back `CLAUDE_PLUGIN_ROOT` → `CLAUDE_CONFIG_DIR` → `~/.claude` (hooks inherit the env var from the harness that loaded them). New `LIFEOS/TOOLS/Paths.ts` mirrors this for runtime tools, and `LifeosConfig.ts` defaults derive from it.
3. **Hardcode sweep** — ~30 hooks/handlers/libs with inline `join(homedir(), '.claude', ...)`-style fallbacks now route through the shared helpers. Five `LIFEOS/TOOLS` files (including the `lifeos.ts` launcher) imported hooks libs via `../../../.claude/hooks/lib/...` — a relative path that climbs out of the config root and back in through a literal `.claude` segment; now the structure-relative `../../hooks/lib/...`, which resolves in any profile.
4. **User-facing command hints** — new `displayPath()` in `hooks/lib/paths.ts`; the always-on hooks' guidance messages (memory-health "Run: bun …", writing-gate PangramScore instructions, Art workflow listing, etc.) are built from the resolved path constants instead of a hardcoded `~/.claude`, so users are never told to run a command against the wrong profile. Default-root installs render byte-identical messages.
5. **`install.sh`** — skills-dir detection probes `CLAUDE_CONFIG_DIR` before the `~/.claude` / `~/.config/claude` defaults.
6. Nested skill-payload mirrors (`install/skills/LifeOS/...`) kept in sync.

## Verification

Performed a real install into a live `CLAUDE_CONFIG_DIR=~/.claude-gc` profile (Linux, Claude Code):

- `InstallHooks.ts --apply`: all 60 commands rewritten to the profile root (0 left on the default), pre-existing foreign hook entries preserved, backup created.
- All 46 wired command hooks executed with representative event payloads: **zero module-resolution or path failures**; state/observability files written only under the active profile; `~/.claude` left untouched.
- Default-root behaviour: the command rewrite is a no-op and every helper resolves to the same paths as before (`bun build` clean on all touched files).

## Out of scope

- The remaining `LIFEOS/TOOLS` files not invoked by hooks (plus Pulse and skill tools) are swept in the follow-up PR #1431, which is stacked on this branch. This PR stays scoped to the runtime-verified always-on surface.

### DIFF
--- a/LifeOS/Tools/InstallEngine.ts
+++ b/LifeOS/Tools/InstallEngine.ts
@@ -548,6 +548,28 @@ function normalizeCommand(cmd: string): string {
     .trim();
 }
 
+/**
+ * Rewrite the default `~/.claude` config-root spellings inside hook commands to
+ * the actually-detected configRoot. The payload's hooks.json ships commands
+ * against the default root (`$HOME/.claude/...`); on a multi-profile setup
+ * (Claude Code's CLAUDE_CONFIG_DIR) those paths point at a different — usually
+ * nonexistent — profile, so every hook errors on startup. No-op when configRoot
+ * IS the default, keeping default installs byte-identical. Pure (no I/O).
+ */
+export function rewriteHooksConfigRoot(hooks: HooksMap, configRoot: string, homeDir: string): HooksMap {
+  if (resolve(configRoot) === resolve(join(homeDir, ".claude"))) return hooks;
+  const defaultRootPattern = /(\$\{HOME\}|\$HOME|~)\/\.claude(?=\/|\s|$)/g;
+  const rewritten: HooksMap = JSON.parse(JSON.stringify(hooks ?? {}));
+  for (const groups of Object.values(rewritten)) {
+    for (const group of groups) {
+      for (const h of group.hooks ?? []) {
+        if (typeof h.command === "string") h.command = h.command.replace(defaultRootPattern, configRoot);
+      }
+    }
+  }
+  return rewritten;
+}
+
 /** Identity key for a hook entry: http → url, else normalized command. */
 function hookKey(h: HookEntry): string {
   if (h.type === "http" && h.url) return `http:${h.url}`;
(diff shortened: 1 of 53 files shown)


danielmiessler:
Thanks @borfast, for this and for the follow-up #1431. The verification behind it (a real install into a CLAUDE_CONFIG_DIR profile with every hook exercised) was above the bar, and it is appreciated. Two notes on how this lands: the repo is a generated artifact, rebuilt from our private source tree on each release, so we close rather than merge here; and the config-dir and path-handling surface has changed substantially in recent releases (install-time CLAUDE_CONFIG_DIR detection is already in current source), so please pull the latest. On runtime state specifically, we route writes only to a fixed set of approved locations by design, so we are not adopting config-dir-relative state paths as-is. If a concrete multi-profile gap remains against the latest release, we would genuinely welcome a fresh issue. Closing with thanks.

PR #1430, opened 2026-07-06
feat: lifeos -r accepts an optional session ID

borfast:
## Problem

The launcher only forwards a bare `--resume` to Claude Code, so `lifeos -r <session-id>` trips the argument parser's unknown-command error — there is no way to resume a *specific* session through the launcher. Falling back to plain `claude -r <session-id>` loses the appended constitutional system prompt, and on multi-profile setups (`CLAUDE_CONFIG_DIR`) it looks in the wrong config root's session store entirely.

## Change

`-r/--resume` now accepts an optional session ID:

- `lifeos -r` — unchanged; opens Claude Code's interactive session picker.
- `lifeos -r <session-id>` — forwards `--resume <session-id>`, resuming that session with the constitution and the correct profile intact.

The ID is only consumed when the next token doesn't start with `-`, so all existing flag combinations behave exactly as before. Help text updated to match.

## Verification

- `lifeos -r` with no ID still launches the picker (bare `--resume` forwarded, unchanged behaviour).
- `lifeos -r <id>` forwards `--resume <id>` and resumes the named session.
- `lifeos -r -m bd` still parses `-m` as a flag, not as a session ID.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/lifeos.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/lifeos.ts
@@ -9,7 +9,7 @@
  *   pai                  Launch Claude (default profile)
  *   pai -m bd            Launch with Bright Data MCP
  *   pai -m bd,ap         Launch with multiple MCPs
- *   pai -r / --resume    Resume last session
+ *   pai -r / --resume    Resume a session (picker, or pass a session ID)
  *   pai --local          Stay in current directory (don't cd to ~/.claude)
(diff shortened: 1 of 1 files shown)


danielmiessler:
Thanks @borfast — clean addition. lifeos -r accepting an optional session ID (with the picker on a bare -r) is already in the current source, matching your approach and credited to you, and it ships in the next release. The repo is generated from our source tree each release, so we close rather than merge here. Thank you — please pull the latest!

PR #1431, opened 2026-07-06
fix: sweep remaining ~/.claude hardcodes across LIFEOS/TOOLS, PULSE, and skills (follow-up to #1429)

borfast:
> **Stacked on #1429** — this branch contains #1429's commits plus one sweep commit on top. Review just the last commit, or merge #1429 first and this rebases to the sweep alone.

## What

#1429 fixed the *always-on* surface (installer, hooks, hook-invoked tools, launcher) for non-default `CLAUDE_CONFIG_DIR` profiles, and its description promised the rest as a follow-up. This is that follow-up: the same treatment extended to the on-demand tools — **152 files across `LIFEOS/TOOLS`, `LIFEOS/PULSE`, and `skills/*/Tools`**.

- **Path resolution**: every inline `~/.claude` fallback (`join`/`resolve` with `homedir()`/`HOME`, `process.env.LIFEOS_DIR || …` patterns, `${HOME}/.claude/…` template strings) now routes through `LIFEOS/TOOLS/Paths.ts` (`getClaudeDir()` / `getLifeosDir()`), matching the hooks' resolver.
- **Projects-dir encoding**: `ActivityParser` and `KnowledgeHarvester` derive Claude Code's munged projects subdir name (`-Users-x--claude`) from the active config dir instead of hardcoding the default profile's encoding — so transcript/memory discovery works on any profile.
- **User-facing hints**: usage blocks, `.env` key instructions, and audit output render the active profile's paths via `displayPath()` (byte-identical output on default installs).
- **Shell scripts**: `${CLAUDE_CONFIG_DIR:-$HOME/.claude}`.

## Deliberately left as-is

- Doc-notation parsers and scanner regexes that match the canonical `~/.claude` spelling inside *documentation text* (`DocCheck`, `ReferenceCheck`, `ArchitectureSummaryGenerator`, the Daemon `SecurityFilter` redaction pattern) — they parse docs, not the filesystem.
- The remote Arbol layout constant (`ARBOL/.claude`).
- macOS-only Interceptor shell help text (heredoc quoting makes blind substitution risky; happy to follow up if wanted).
- `skills/Telos/DashboardTemplate` — a self-contained Next.js app with its own build; excluded from the sweep.

## Verification

- `bun build` clean on all 299 swept TypeScript files. 75 files failed to build **before** the sweep (missing external deps like `yaml`, `smol-toml` — expected outside a deployed install with Bun auto-install) and the exact same 75 fail after: **zero new failures**.
- Post-sweep audit: zero functional `.claude` literals remain outside the justified list above.
- Unlike #1429's hooks surface (runtime-verified end to end on a live `~/.claude-gc` profile), most of these tools need API keys, macOS, or a running Pulse server — this sweep is build-verified and pattern-reviewed, which is why it's separate from #1429 rather than part of it.

### DIFF
--- a/LifeOS/Tools/InstallEngine.ts
+++ b/LifeOS/Tools/InstallEngine.ts
@@ -548,6 +548,28 @@ function normalizeCommand(cmd: string): string {
     .trim();
 }
 
+/**
+ * Rewrite the default `~/.claude` config-root spellings inside hook commands to
+ * the actually-detected configRoot. The payload's hooks.json ships commands
+ * against the default root (`$HOME/.claude/...`); on a multi-profile setup
+ * (Claude Code's CLAUDE_CONFIG_DIR) those paths point at a different — usually
+ * nonexistent — profile, so every hook errors on startup. No-op when configRoot
+ * IS the default, keeping default installs byte-identical. Pure (no I/O).
+ */
+export function rewriteHooksConfigRoot(hooks: HooksMap, configRoot: string, homeDir: string): HooksMap {
+  if (resolve(configRoot) === resolve(join(homeDir, ".claude"))) return hooks;
+  const defaultRootPattern = /(\$\{HOME\}|\$HOME|~)\/\.claude(?=\/|\s|$)/g;
+  const rewritten: HooksMap = JSON.parse(JSON.stringify(hooks ?? {}));
+  for (const groups of Object.values(rewritten)) {
+    for (const group of groups) {
+      for (const h of group.hooks ?? []) {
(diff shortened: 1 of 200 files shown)


danielmiessler:
Thanks @borfast, this follow-up to #1429 was an unusually careful sweep. Same note as on #1429: the repo is generated from our private source tree each release (so we close rather than merge), the path-handling surface has moved a lot in recent releases, and runtime state is deliberately routed only to approved locations, so we are not taking config-dir-relative state paths as-is. Please pull the latest, and if a specific gap remains we would welcome a fresh issue. Closing with thanks for the contribution.

Issue #1404, opened 2026-07-03
settings.system.json env values ship a literal, unexpanded $HOME — breaks USER scaffold on fresh installs (regression of the #1124 / #1270 class)


rpriven:
**What happens**

The shipped `LifeOS/install/settings.system.json` env block sets path variables as literal strings:

```json
"env": {
    "LIFEOS_DIR": "$HOME/.claude/LIFEOS",
    "PROJECTS_DIR": "$HOME/Projects",
    "LIFEOS_CONFIG_DIR": "$HOME/.claude/LIFEOS",
    ...
```

Claude Code injects `settings.json` env values **verbatim — it never shell-expands them**. Verified directly: after setup, a fresh authed session's `printenv LIFEOS_CONFIG_DIR` returns the literal string `$HOME/.claude/LIFEOS`.

On a clean Debian 13 install (Claude Code 2.1.198, bun 1.3.14), during `/LifeOS setup`:

- `ScaffoldUser` wrote the 93-file USER tree into a junk directory literally named `$HOME/` under the skill folder
- `LinkUser` created a broken, self-referential symlink `~/.claude/LIFEOS/USER → $HOME/.claude/LIFEOS/USER`

The setup agent caught it mid-run. From its own transcript:

> "Root cause found: the env var `LIFEOS_CONFIG_DIR=$HOME/.claude/LIFEOS` is set to a **literal, unexpanded** string. The bun tools don't expand `$HOME`, so:
> - ScaffoldUser wrote 93 files into a junk dir literally named `$HOME/` under the skill folder
> - LinkUser created a **broken symlink**: `~/.claude/LIFEOS/USER → $HOME/.claude/LIFEOS/USER` (also self-referential once expanded — it points back into configRoot, breaking the separation contract)"

It then repaired itself (moved USER into place, removed the junk tree) — genuinely impressive agentic recovery, but every fresh install is currently performing self-surgery. And the deeper issue: **anything else that reads `LIFEOS_CONFIG_DIR`/`LIFEOS_DIR` from env at runtime (hooks, Pulse, TOOLS) receives the literal string** and will mis-resolve unless it defensively expands.

**Repro**

Clean Linux box → bootstrap installer → `/LifeOS setup` → observe the `$HOME/` junk dir under the skill folder + broken symlink. Or simply, post-setup: `claude -p 'printenv LIFEOS_CONFIG_DIR'`.

[...]

Issue #1402, opened 2026-07-03
6.0.0: settings.system.json contract mismatch — payload ships it without hooks, MemoryHealthCheck requires hooks registered in it (fresh installs report CRITICAL)


infinitelyloopy-bt:
**Components:** `install/settings.system.json`, Setup workflow step 4, `LIFEOS/TOOLS/MemoryHealthCheck.ts` / `MemoryHealthGate.hook.ts`, LifeOS 6.0.0.

**What happens:** three parts of 6.0.0 disagree about what `settings.system.json` is:
1. The payload ships `install/settings.system.json` deliberately **without** a `hooks` block ('the system half WITHOUT the hooks block; InstallHooks owns hooks' — Setup.md).
2. Setup.md step 4 overlays it **as `settings.json`**, so nothing ever lands at `~/.claude/settings.system.json`.
3. `MemoryHealthCheck.ts` treats `~/.claude/settings.system.json` as a required regression-reference file and reports **CRITICAL** if it's missing — and once you copy the shipped one there, it still reports CRITICAL because the four memory hooks (`MemoryReviewTrigger`, `MemoryReviewFire`, `MemoryHealthGate`, `MemoryDeltaSurface`) aren't registered inside it.

Net effect: every fresh 6.0.0 install surfaces `🩺 MEMORY HEALTH: CRITICAL` via MemoryHealthGate even when the live `settings.json` wiring is complete and correct.

**Repro:** fresh 6.0.0 install per INSTALL.md → `bun LIFEOS/TOOLS/MemoryHealthCheck.ts` → `settings-missing:settings.system.json` (critical); copy the payload file in → four `settings-hook-missing:...` criticals.

**Workaround:** compose the file as payload system-half + the `hooks` object from `install/hooks/hooks.json` and write it to `~/.claude/settings.system.json` — health then passes (warn-only).

**Suggested fix:** pick one contract — either ship `settings.system.json` including the canonical hooks block and have Setup place it at the config root, or drop the settings.system.json checks from MemoryHealthCheck for skill-payload installs.

rpriven:
One more data point on the contract mismatch, from a Linux (Debian) install: the merged SessionStart hook chain runs `MergeSettings.ts` **every session** with `--system ~/.claude/settings.system.json --output ~/.claude/settings.json` — i.e., against the file setup never places.

[...]

PR #1432, opened 2026-07-06
fix(install): write standing settings.system.json snapshot on hooks install (#1402)

sauldataman:
Problem

  Addresses #1402. MemoryHealthCheck.ts CHECK 3 treats ~/.claude/settings.system.json as the standing source of truth for system hook wiring and requires the four memory hooks to be registered in it. But no
  install step ever writes that file:

  - install.sh never touches settings.
  - InstallHooks.ts --apply merges the payload's hooks.json into settings.json, and copies the hook scripts to disk, but never produces the standing settings.system.json.

  So on a fresh install the health check fails against a file that no step is responsible for creating:

  CRITICAL — settings.system.json not found at ~/.claude/settings.system.json

  (This is the "never written" variant of the contract gap in #1402; the "written without hooks" variant is the same missing producer seen from the other side.)

  Fix

  InstallHooks.ts --apply now also writes the standing snapshot to <configRoot>/settings.system.json:

  - base = install/settings.system.json (the canonical system half — same file SettingsBackport.ts treats as the system layer)
  - hooks = incoming (the payload's own wiring), not merged. Freezing a user's foreign hook entries into "system truth" would make MemoryHealthCheck misreport them as required system hooks.

  Write-path (join(configRoot, "settings.system.json")) and the health check's read-path (join(HOME, ".claude", "settings.system.json")) resolve to the same location under a normal install; both use
  process.env/os.homedir(), so this is not affected by the literal-$HOME issue in #1404.

  Scope / semantics

  - Dry-run unchanged — the snapshot is only written under --apply, after the existing trust gate.
  - Idempotent — re-running --apply reproduces byte-identical output.
  - Refreshed on every --apply, so re-installs stay current. (A user who applied the manual workaround from #1402 will have it replaced by the identical generated version.)
  - Does not touch MemoryHealthCheck.ts — the check's source-of-truth semantics are correct; this supplies the missing producer.
  - Applied to both copies of InstallHooks.ts (skill source + install payload mirror) and the two Setup.md step-7 doc lines.

  Verification

  Sandbox --config-root, then the real consumer with HOME overridden:

  $ bun InstallHooks.ts --config-root $SBX --skill-root $SKILL --apply
  $ bun InstallHooks.ts --config-root $SBX --skill-root $SKILL --apply
  $ HOME=$SBX_HOME bun LIFEOS/TOOLS/MemoryHealthCheck.ts
  CHECK 3 findings: none
  ok: settings-hook-present:settings.system.json:MemoryReviewTrigger.hook.ts
  ok: settings-hook-present:settings.system.json:MemoryReviewFire.hook.ts
  ok: settings-hook-present:settings.system.json:MemoryHealthGate.hook.ts
  ok: settings-hook-present:settings.system.json:MemoryDeltaSurface.hook.ts

  Dry-run writes no snapshot; second --apply is byte-identical to the first.


danielmiessler:
Thanks a lot for this, @sauldataman — really appreciate you flagging it and taking the time to write the fix.

When I went back through the current source, this turns out to already be handled: thanks for chasing the #1402 fresh-install false-CRITICAL — that same issue is now resolved in source by redesigning MemoryHealthCheck CHECK 3 to read truth from the effective settings.json, so the snapshot write is no longer needed. The system has moved a fair bit since you opened this (the PAI → LifeOS rename and a few subsystem rewrites), so the gap you spotted has since been closed independently.

sauldataman:
cool ! thanks

PR #1441, opened 2026-07-06
Inference.ts: scrub ANTHROPIC_BASE_URL alongside API key/token in nested claude spawn

### LINKED ISSUE #1440: Inference.ts nested claude subprocess: ANTHROPIC_BASE_URL not scrubbed alongside API key/token, causing silent auth failures + timeouts when a local proxy is configured
## Where

`PAI/TOOLS/Inference.ts` (same code present across releases, e.g. `Releases/v5.0.0/.claude/PAI/TOOLS/Inference.ts`, `Releases/v4.0.x/.claude/PAI/Tools/Inference.ts`, `skills/PAI/Tools/Inference.ts`).

## Bug

`inference()` spawns a nested `claude --print` subprocess to force subscription billing instead of API-key billing. To do that, it deletes credentials from the child process's environment:

```ts
delete env.ANTHROPIC_API_KEY;
delete env.ANTHROPIC_AUTH_TOKEN;
```

It does **not** also clear `ANTHROPIC_BASE_URL`. If a user has `ANTHROPIC_BASE_URL` set in their shell environment (a common pattern for routing Claude Code traffic through a local proxy — e.g. LiteLLM — for cost tracking/observability), every nested `inference()` call — including the mode/tier classifier invoked by `PromptProcessing.hook.ts` on every prompt, and the Advisor calls in the Algorithm's VERIFY phase — still targets that local proxy, but now with zero valid Anthropic credentials in env (both keys were just stripped). The proxy correctly rejects the request with `401 Unauthorized`. The subprocess has no way to succeed, retries, and eventually burns the full timeout (25s for the classifier) before giving up.

**Observed impact**: with `ANTHROPIC_BASE_URL` set, the mode classifier fails on *every single message*, forcing the fail-safe path (`SOURCE: fail-safe`, defaulting to Algorithm E3) instead of real classification — and the local proxy's logs get flooded with repeated 401s from the retries.

## Fix

Scrub `ANTHROPIC_BASE_URL` alongside the existing two deletes, so nested subscription-billed calls always go straight to Anthropic's real API regardless of what the parent shell has configured for interactive use:

```ts
delete env.ANTHROPIC_API_KEY;
delete env.ANTHROPIC_AUTH_TOKEN;
delete env.ANTHROPIC_BASE_URL;
```

Happy to open a PR with this change (plus the matching one-liner in the corresponding file under `LifeOS/install/`) if that's preferred over a maintainer applying it directly.

jshingler:
Fixes #1440.

## Bug

`Inference.ts`'s nested `claude --print` subprocess spawn deletes `ANTHROPIC_API_KEY` and `ANTHROPIC_AUTH_TOKEN` from the child environment to force subscription billing (per the existing comment on that block), but never cleared `ANTHROPIC_BASE_URL`.

If a user has `ANTHROPIC_BASE_URL` set to a local proxy (e.g. a LiteLLM gateway used for cost tracking/observability — a documented pattern), every nested `inference()` call — including the mode/tier classifier `PromptProcessing.hook.ts` invokes on every prompt, and the Advisor calls in the Algorithm's VERIFY phase — kept targeting that proxy, but now with zero valid Anthropic credentials in env (both keys just stripped). The proxy correctly rejects with `401 Unauthorized`, the subprocess retries, and the call burns its full timeout before giving up.

**Observed impact**: with `ANTHROPIC_BASE_URL` set, the mode classifier failed on *every single message* (`Timeout after 25000ms`), forcing the fail-safe path (`SOURCE: fail-safe`, default tier) instead of real classification — plus a steady stream of 401s in the proxy's logs from the retries.

## Fix

One-line addition scrubbing `ANTHROPIC_BASE_URL` alongside the existing two deletes, so nested subscription-billed calls always go straight to Anthropic's real API regardless of what the parent shell has configured for interactive proxy routing.

## Testing

Verified against a live setup with `ANTHROPIC_BASE_URL` pointed at a local LiteLLM proxy: before the fix, every classifier invocation timed out with repeated 401s in the proxy logs; the fix routes nested calls around the proxy as intended.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/Inference.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/Inference.ts
@@ -162,6 +162,12 @@ async function inferenceAttempt(options: InferenceOptions, modelOverride?: strin
     // either path leaks subscription work onto API-key billing. Scrub both.
     delete env.ANTHROPIC_API_KEY;
     delete env.ANTHROPIC_AUTH_TOKEN;
+    // Also scrub ANTHROPIC_BASE_URL: if it points at a local proxy (e.g. a
+    // LiteLLM gateway used for cost tracking/observability), the child process
+    // would still target that proxy after losing its credentials above,
+    // causing every nested call to fail auth (401) and retry until timeout.
+    // Force nested subscription calls straight to Anthropic's real API.
+    delete env.ANTHROPIC_BASE_URL;
 
     const hasImages = options.imagePaths && options.imagePaths.length > 0;
     const args = [


danielmiessler:
Thanks @jshingler — appreciated, and you were right about the gap. We have since overhauled how the inference / model-routing layer works (Fable vs Opus routing, subprocess execution, env scrubbing), and scrubbing ANTHROPIC_BASE_URL alongside the API key and auth token is already in the current source. This repo is a generated artifact — regenerated from our private source tree on each release — so we close rather than merge PRs here, but the fix is in. Please pull the latest. Closing with thanks!

PR #1340, opened 2026-06-10
feat(PAIUpgrade): mandatory cross-recommendation interference analysis (Step 6b)

82ATW:
## Problem

PAIUpgrade's synthesis gates every recommendation against **prior work** (Thread 0 → Prior Status), but nothing checks recommendations against **each other**. On large reports, that lets four classes of problems ship undetected until implementation time: measurement confounds (one rec changes the thing another rec measures), same-file collisions, input dependencies (one rec evicts what another reads), and doctrine/version bundling (several recs editing one versioned artifact).

## What this adds

Four files under `Releases/v5.0.0/.claude/skills/PAIUpgrade/`:

- **`Workflows/Upgrade.md`** — new **Step 6b**: a four-class interference matrix plus a mandatory adversarial verification pass by a second agent (different model family when available; fresh-context same-model fallback otherwise — disputes resolve by probing ground truth, never by model preference). Gated at **>10 recommendations** (~45 pairs — roughly where a reliable mental pairwise sweep ends); at ≤10 the section reduces to a single line so quick modes stay fast. Every identified conflict must carry a guard or the report is blocked.
- **`References/OutputFormat.md`** — `⚠️ Interactions & Sequencing` joins the canonical section order (position 4), with the section spec and hard rule 11: *Prior-Status gates against the past; this gates against each other.*
- **`SKILL.md`** — extraction sub-rule under rule 4, updated section-order summary, and a Gotchas entry.
- **`References/ExampleReport.md`** — the worked example now demonstrates the ≤10 one-liner form, staying consistent with the spec it illustrates.

## Origin

Found in practice: a 44-recommendation report shipped fully Prior-Status-gated with zero pairwise analysis; the retrofit interference pass found two high-severity measurement confounds (a fallback setting and a safety-routing behavior each silently swapping the model under an A/B eval) plus six same-file collisions. The patch itself was hardened through the same adversarial-review loop it mandates.

## Scope notes

- Targets the `Releases/v5.0.0` tree (the shipped runtime). Happy to also port to `Packs/Utilities/src/PAIUpgrade` if that's the preferred source of truth — that copy currently reflects an earlier generation of the workflow.
- `Workflows/AlgorithmUpgrade.md` (a sibling recommendation-emitting workflow) is intentionally not gated here to keep the change minimal; happy to follow up if useful.

PR #1442, opened 2026-07-07
feat(Upgrade): mandatory cross-recommendation interference analysis (Step 6b)

82ATW:
## What

Adds a mandatory **cross-recommendation interference analysis** (Step 6b) to the Upgrade skill. Prior-Status gating checks each recommendation against the PAST; Step 6b checks them against EACH OTHER.

When a report carries >10 recommendations, the skill must:

1. **Build a pairwise interference matrix** across four classes: measurement confounds, same-file/surface collisions, resource/input dependencies, doctrine/version bundling.
2. **Adversarially verify it** with a second agent (different model family when available); disputes resolve by probing ground truth.
3. **Guard every conflict** — a conflict pair without a named resolution blocks the report.
4. **Emit an `⚠️ Interactions & Sequencing` section** — conflict table + wave-ordered execution sequence + synergies.

At ≤10 recommendations the section reduces to a single line, so quick modes stay fast.

## Why

A 44-recommendation report shipped with every row Prior-Status-gated but zero pairwise analysis; a retrofit found two high-severity measurement confounds (a fallback setting and a safety-routing behavior each silently swapping the model under an A/B eval) plus six same-file collisions. A large recommendation set is itself a system — this gates the set against itself.

## Files

- `LifeOS/install/skills/Upgrade/Workflows/Upgrade.md` — new Step 6b (gate, matrix classes, adversarial pass, guards) + Step 7 section order
- `LifeOS/install/skills/Upgrade/References/OutputFormat.md` — new Interactions & Sequencing section spec + hard rule 11
- `LifeOS/install/skills/Upgrade/SKILL.md` — Step 6b gate in extraction rules, section order, Gotcha
- `LifeOS/install/skills/Upgrade/References/ExampleReport.md` — example of the ≤10 single-line form

## Notes

Supersedes #1340, which targeted the frozen `Releases/v5.0.0` snapshot before the v6.0.0 restructure moved the live skill to `LifeOS/install/skills/Upgrade/`. Same feature, retargeted to the live files with current naming.

### DIFF
--- a/LifeOS/install/skills/Upgrade/References/ExampleReport.md
+++ b/LifeOS/install/skills/Upgrade/References/ExampleReport.md
@@ -84,6 +84,12 @@ echo "Session: ${CLAUDE_SESSION_ID}"
 
 ---
 
+## ⚠️ Interactions & Sequencing
+
+interference: none found across 2 recommendations (≤10 — full matrix not required; see Workflows/Upgrade.md Step 6b)
+
+---
+
 ## 📊 Summary
 
 | # | Technique | Source | Priority | LifeOS Component | Effort |
--- a/LifeOS/install/skills/Upgrade/References/OutputFormat.md
+++ b/LifeOS/install/skills/Upgrade/References/OutputFormat.md
@@ -7,10 +7,11 @@
 1. **✨ Discoveries** — what was found, ranked by interestingness
 2. **🔥 Recommendations** — what to do, ranked by priority tier
 3. **🎯 Technique Details** — full extraction with code/quotes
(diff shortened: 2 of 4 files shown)


danielmiessler:
Thanks @82ATW — genuinely sharp observation that Prior-Status gates each recommendation against the past but nothing gated them against each other. We went a different way and are not adopting it as a mandatory step, but it was a good idea, well designed (the interference taxonomy and the >10-rec gate especially). This repo is generated from our source tree each release, so we close rather than merge here. Thanks for engaging with the Upgrade skill this concretely.

PR #1458, opened 2026-07-10
fix(telegram): add explicit config switch for voice summaries

anikinsasha:
`sendVoiceSummary` was gated only by the presence of `ELEVENLABS_API_KEY`. Because the Telegram module and the CLI `/notify` VoiceServer share that one env var, an install that wants ElevenLabs for the CLI channel but *not* spoken Telegram notes had no way to turn the latter off short of unsetting the key (which kills both).

This adds a `[telegram] voice_summaries` flag, threaded through the existing `loadPulseConfig → startTelegram → activeConfig` path — no new plumbing (`loadPulseConfig` already passes the whole `[telegram]` table through). Defaults to enabled (undefined ⇒ on) so existing installs are unchanged; set `voice_summaries = false` to keep text replies but suppress the voice note. `telegramHealth()` reflects the switch.

Design note: a dedicated `[telegram]` flag was chosen over reusing the top-level `[voice] enabled` because `[voice]` gates the separate CLI VoiceServer/`/notify` channel — conflating the two surfaces would be wrong.

### DIFF
--- a/LifeOS/install/LIFEOS/PULSE/PULSE.toml
+++ b/LifeOS/install/LIFEOS/PULSE/PULSE.toml
@@ -15,6 +15,9 @@ enabled = true
 
 [telegram]
 enabled = true
+# Send a spoken summary of each text reply as a Telegram voice note. Requires
+# ELEVENLABS_API_KEY. Set false to keep text replies but suppress voice notes.
+voice_summaries = true
 
 [imessage]
 enabled = false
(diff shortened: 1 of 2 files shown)


danielmiessler:
Thank you @anikin-xyz — a clean opt-out with `undefined ⇒ on` preserving existing behavior. Ported into source for **v7.0.0**, credited.

PR #1457, opened 2026-07-10
fix: de-hardcode the author's DA name across TOOLS, PULSE, and hooks

anikinsasha:
Following the invitation in #1334 to help scrub the author's personal identifiers out of the shipped payload, this de-hardcodes the DA name (`"kai"`) so a fresh install reflects the name the principal chose during setup instead of the author's.

One root cause — the DA name literal baked into ~15 sites — across six clusters:

1. **TOOLS attribution defaults** — `TelosFreshness.ts` (`by = "kai"` on `bumpTelosTimestamp` / `bumpContextTimestamp`).
2. **PULSE.toml `[da].primary`** — personal literal → neutral placeholder + pointer to the canonical config.
3. **DA-subsystem fallbacks** — `DAGrowth.ts`, `DASchedule.ts` (`?? "kai"`, help text).
4. **CLI banner text** — `MemoryStatus.ts` (`kai status`), `MemoryInsights.ts` (`kai insights`), `BannerRetro.ts` (`KAI` block-letter ASCII).
5. **Work labels** — `WorkSweep.ts` (×4), `PULSE/modules/work.ts`, `hooks/ReminderRouter.hook.ts` (`Agent:kai` → `Agent:<da-name>`).
6. **Telegram + memory graph** — `telegram.ts` speaker-label regex and voice-note filename, `MemoryGraph.ts` stopword list (drops the personal names `kai`, `daniel`).

Mechanism: TOOLS and hooks read the configured name via `getDAName()` (`hooks/lib/identity.ts`); PULSE modules resolve it through the config they already load. The three `Agent:<da>` label producers share one source so their labels agree. `BannerRetro` renders the DA name (`stats.name`, default `"LifeOS"`) instead of hardcoded art. Surgical — no behavior change beyond the name value.

Deferred as a follow-up (template-semantic, riskier than a surgical rename): the structural `kai` keys in `skills/Prompting/Templates/Data/Agents.yaml` / `VoicePresets.yaml` — the displayed names there already use `{{DA_NAME}}`; only the lookup keys are literal.

PR #1459, opened 2026-07-10
fix(pulse): defer [da] identity to the user config (LIFEOS_CONFIG.toml)

anikinsasha:
PULSE's `[da].primary` was read only from the shipped `PULSE.toml`, so the DA name was unoverridable from the user tier — it lived in two files (`PULSE.toml` and `LIFEOS/USER/CONFIG/LIFEOS_CONFIG.toml [da].name`) that could silently disagree.

`loadPulseConfig` now defers `[da].primary` to `LIFEOS_CONFIG.toml [da].name` when present, via the existing `LifeosConfig` loader — converging the two identity configs on one source of truth rather than adding a third layer. Fresh installs without the user config keep the `PULSE.toml` value (`try/catch` fallback, functionally tested both ways).

Composes cleanly with #1457 (which neutralizes the shipped placeholder): with both, the shipped value becomes a pure fresh-install fallback and the user config wins everywhere else.

### DIFF
--- a/LifeOS/install/LIFEOS/PULSE/pulse.ts
+++ b/LifeOS/install/LIFEOS/PULSE/pulse.ts
@@ -17,6 +17,7 @@
 import { join } from "path"
 import { readFileSync, existsSync } from "fs"
 import { parse } from "smol-toml"
+import { loadLifeosConfig } from "../TOOLS/LifeosConfig"
 
 // ── Load .env before anything else ──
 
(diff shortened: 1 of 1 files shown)


danielmiessler:
Thank you @anikin-xyz — converging the DA identity on `LIFEOS_CONFIG.toml` is the right call. Ported into source for **v7.0.0**, credited.

Issue #1466, opened 2026-07-11
Session-resolution guessing: AlgoPhase.ts writes phase to the wrong session's work.json row; Inference.ts --auto-state attaches the wrong ISA


jbmml:
## Summary

Two tools resolve "which session am I acting on" non-deterministically and then act on the guess:

1. `LifeOS/install/LIFEOS/TOOLS/AlgoPhase.ts` — when called without `--slug`/`--uuid` (and `CLAUDE_SESSION_ID` unset, which is always today — Claude Code does not export it to tool subprocesses), `resolveSlug()` falls back to the most-recent algorithm-mode row in `MEMORY/STATE/work.json`. It prints a WARN (the hazard is even called out in the code comment at the fallback site) **but still writes the phase to that row**. With two Algorithm sessions alive, a bare `AlgoPhase think` from one session flips the OTHER session's registry row — cross-session state corruption that propagates to the Pulse dashboard and anything else reading work.json.

2. `LifeOS/install/LIFEOS/TOOLS/Inference.ts` (`synthesizeAdvisorState()`, used by `--mode advisor --auto-state`) — reads `work.json` expecting `active` / `current` / `activeSession` top-level keys. The registry's actual shape is `{ "sessions": { "<slug>": {...} } }` and has no such keys, so resolution **always** falls through to "most recently modified directory in `MEMORY/WORK/`". That fallback is usually right by accident (the current session's ISA is often the newest), but it carries no session identity at all — under concurrency or right after another session completes, it attaches a completed, unrelated session's ISA as the advisor's state blob, so the reviewer confidently reviews the wrong work.

## Repro (AlgoPhase)

With a `work.json` containing two live algorithm-mode rows (easy to hit with two concurrent Claude Code tabs):

```json
{
  "sessions": {
    "20260101-000001_session-a": { "mode": "interactive", "currentMode": "algorithm", "phase": "build",
      "sessionUUID": "aaaaaaaa-0000-0000-0000-000000000001", "updatedAt": "2026-01-01T18:00:00.000Z" },
    "20260101-000002_session-b": { "mode": "interactive", "currentMode": "algorithm", "phase": "build",
      "sessionUUID": "aaaaaaaa-0000-0000-0000-000000000002", "updatedAt": "2026-01-01T18:30:00.000Z" }
  }
}
```

```
$ bun ~/.claude/LIFEOS/TOOLS/AlgoPhase.ts think        # called from session-a's context
[...]

PR #1467, opened 2026-07-11
fix: deterministic session resolution for Inference --auto-state

jbmml:
Part 1 of 2 for #1466 (the `Inference.ts` half; the `AlgoPhase.ts` fail-closed change follows in a separate PR).

## What

`synthesizeAdvisorState()` (the `--mode advisor --auto-state` path) read `work.json` expecting `active`/`current`/`activeSession` top-level keys. The registry's actual shape is `{sessions: {slug: row}}` — those keys never exist — so resolution always fell through to "most recently modified directory in `MEMORY/WORK/`". That fallback is usually right by accident, but it carries no session identity: under concurrency or right after another session completes, it attaches an unrelated session's ISA as the advisor state, and the reviewer reviews the wrong work.

## How

Deterministic resolution, documented in the function's doc comment:

1. Row matching `CLAUDE_SESSION_ID`, any phase — a `phase: complete` row of the SAME session is still this session, and the verification doctrine fires the advisor at the completion boundary itself. Set-but-unmatched returns no state rather than falling through to a guess. (Honesty note: Claude Code does not export `CLAUDE_SESSION_ID` to tool subprocesses today, so this path is forward-looking defense-in-depth; the operative resolution today is steps 2–4.)
2. No env: a `complete` row updated within the last **5 minutes** is treated as the just-finished caller making its mandatory completion-boundary advisor call — the most recent such row wins, UNLESS any fresh live row also exists (then we cannot tell which session is calling → ambiguous). Without this window, fixing the mtime guess would have broken the doctrine's advisor-after-complete flow for every single-session user.
3. No env, no recent-complete: exactly ONE live (non-complete, non-native) row updated within the last **60 minutes**. Older non-complete rows are crash leftovers — the registry stale sweep only prunes them after ~7 days — and are never guessable.
4. Anything else returns an `ADVISOR STATE UNAVAILABLE` marker and warns loudly on stderr. Wrong state is worse than no state for a reviewer, so the mtime-based directory fallback is removed rather than fenced.

Behavior change disclosure: previously `--auto-state` always attached SOME ISA. Now, when the caller's session cannot be determined (multiple live sessions, stale-only registry, unmatched env), the advisor runs with an explicit no-state marker instead of a possibly-wrong blob.

No CLI contract change: flags, positional args, output framing, and exit codes are untouched.

## Verification

Self-contained bun suite added alongside the tool (`Inference.autostate.test.ts`, runs in-process against a throwaway `$HOME`): env-match incl. complete row; env unmatched → no state; single fresh live → its ISA; two live → no state; recent-complete alone → used; two recent-completes → most recent; recent-complete + live → no state; 1h-old complete → no state; 2h-idle live → no state; native placeholder rows excluded. 10/10 pass; `tsc --noEmit --strict` clean.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/Inference.autostate.test.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/Inference.autostate.test.ts
@@ -1,0 +1,133 @@
+/**
+ * Inference.autostate.test.ts — session-resolution contract for
+ * synthesizeAdvisorState() (the --mode advisor --auto-state path).
+ *
+ * Runs in-process: the function resolves everything from $HOME, so each test
+ * points HOME at a throwaway tree. No inference call is ever made — only the
+ * state-synthesis step is exercised. Run with: bun test Inference.autostate.test.ts
+ */
+
+import { test, expect, beforeEach, afterEach } from "bun:test";
+import { mkdtempSync, mkdirSync, writeFileSync } from "fs";
+import { tmpdir } from "os";
+import { join } from "path";
+import { synthesizeAdvisorState } from "./Inference";
+
+const MIN = 60 * 1000;
+const iso = (agoMs: number) => new Date(Date.now() - agoMs).toISOString();
+
+const UUID_A = "aaaaaaaa-0000-0000-0000-000000000001";
+const UUID_B = "aaaaaaaa-0000-0000-0000-000000000002";
+
+let savedHome: string | undefined;
+let savedUserProfile: string | undefined;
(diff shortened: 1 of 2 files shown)


danielmiessler:
Thank you @jbmml — the deterministic session-resolution intent is sound. This specific path (`Inference.ts --mode advisor --auto-state` / `synthesizeAdvisorState`) was retired in the v8 modes/tiers consolidation, so the target function no longer exists in current source and there's nothing to port. Closing as superseded by that redesign, not as a rejection of the idea.

PR #1469, opened 2026-07-12
Fix learning loop: inference-as-truth and per-message lesson minting corrupt the standing corpus

justinkatz94-glitch:
## Problem

The rating/learning loop treats inferred signal as truth and crystallizes too eagerly:

- `SatisfactionCapture` auto-scores every message (mostly model-inferred), and a low score immediately mints a standing learning file plus a full-transcript failure capture, per message.
- When sentiment inference fails, a fabricated neutral 5 is written as a real `implicit` rating (confidence 0.3), so error placeholders pollute every downstream average.
- Session-start readback injects raw learning digests, failure patterns, and aggregate rating averages as if they were vetted guidance, but at session start there is no active plan to apply them to.
- `LearningPatternSynthesis` emits standing recommendations from single occurrences and analyzes fabricated neutrals as real data.

Net effect in a real install: inferred guesses harden into durable lessons, `FAILURES/` grew to 237MB from per-message full-transcript copies, and corrupted "lessons" feed back into context every session.

## Fix (four commits, one concern each)

1. **learning-readback + LoadContext: readback becomes an authored allowlist.** Session start surfaces only crystallized Wisdom Frames (>= 85% confidence) and authored constructive-correction guidance whose relevance is decided by recent corpus signals. Raw frustration text and aggregate averages are never injected.
2. **SatisfactionCapture: provenance gate.** Corrections are classified `situational` (mood/pace, momentary) vs `durable` (articulated repeatable rule), defaulting to situational. Explicit ratings get a deterministic classifier (no extra LLM call); implicit ratings get the classification from the existing inference call. Only durable lows may mint learnings or failure captures. Inference failures are labeled `source: inference-failed`, `confidence: 0` instead of masquerading as real implicit ratings.
3. **FailureCapture: bounded and gated.** The full-transcript `copyFileSync` becomes a last-200-line `transcript.tail.jsonl` (MemorySystem.md doc row updated to match). A new source gate skips non-explicit lows: an inferred low is a guess, not a real failure. Both SatisfactionCapture call sites now pass their `source` so the gate owns the policy in one place.
4. **LearningPatternSynthesis: provenance-aware.** Fabricated neutrals are excluded from averages, pattern detection, and hypothesis generation; situational lows are excluded from recommendation-shaping and hypothesis seeding (reported as a count); a standing recommendation now requires a repeatable pattern (>= 2 occurrences).

## Why this is the right shape

Per-message capture = raw signal only; standing lessons = distilled deliberately from aggregated signal, explicit-first. Inferred signal never hardens into durable lessons, opinions, or metrics. De-minting is loss-free: the synthesis batch's input was always `ratings.jsonl`, never the minted per-message artifacts, and durable lows keep `sentiment_summary` + `detailed_context` in the raw row.

Verified in a live install running this design since late June: ratings flow, situational redirects are counted but mint nothing, failure captures stay bounded, and the synthesis batch consumes the new row format directly.

## Testing

Sandboxed runtime probes against this branch (isolated LIFEOS_DIR / HOME, real hook and tool invocations):

- Explicit low with a mood comment ("2 too slow"): rating row written with `provenance: situational`; zero learning files, zero failure captures.
- Explicit low with an articulated rule ("2 from now on always use bun not npm"): `provenance: durable`; learning file minted, failure capture created with `transcript.tail.jsonl` at exactly 200 lines from a 300-line transcript.
- Readback loaders: a 90% frame surfaces, a 40% frame does not; a corpus feedback line containing a verification complaint surfaces only the authored guidance string, never the raw text.
- `LearningPatternSynthesis --dry-run` over a mixed fixture (3 fabricated neutrals, 1 situational low, durable lows): average computed from analyzable rows only, situational row produced no pattern, recommendation emitted only for the >= 2-occurrence pattern.
- All five files pass `bun build --target bun` at every commit in the series (bisect-safe).

## Notes for reviewers

- `ratings.jsonl` rows gain optional `provenance` and a widened `source` union; all consumers treat unlabeled legacy rows as before (analyzable, explicit-gated behavior unchanged for them).
- The old statusline read fabricated neutrals into displayed averages; the current `LIFEOS_StatusLine.sh` reads a `learning-cache.sh` whose producer is not in the tree, so no statusline change is included. If/when the cache producer ships, it should apply the same `source !== 'inference-failed' && confidence !== 0` filter now used in `LearningPatternSynthesis`.
- All five touched files pass `bun build --target bun` from `LifeOS/install/` (SatisfactionCapture with `--external yaml`, a pre-existing tree dependency).


danielmiessler:
Thank you @justinkatz94-glitch — this is a thoughtful analysis of a real corruption path. The catch: since the 2026-07-10 narrowing, the learning-synthesis corpus this fixes is no longer injected into context (LoadContext now reads back wisdom-frames only), so we're deliberately shrinking that subsystem rather than hardening it. Porting a med-risk change to it in this release doesn't fit that direction. Deferring — but the SatisfactionCapture provenance point is a good one we'll revisit.

PR #1475, opened 2026-07-13
fix(skill): route `doctor` in the LifeOS skill

erf1nd0r:
## Problem

`Doctor.ts` ships in the install payload (`LifeOS/install/LIFEOS/TOOLS/Doctor.ts`) and `INSTALL.md` documents it at step 8.5 — but `SKILL.md` never routes it.

Three places in `SKILL.md` omit `doctor`:

- the **Workflow Routing** table lists only `setup` / `interview` / `update` / `uninstall`
- `argument-hint` is `"[setup|interview|update|uninstall]"`
- the `USE WHEN` triggers in `description` carry no doctor phrasing

Net effect: after a successful install the capability prober exists on disk and is referenced by the install doc, but is unreachable through the skill. `/LifeOS doctor` falls through to no route, and the model has to go find `Doctor.ts` by search to honor the request. A capability checker you can't reach is exactly the silent degradation Doctor was built to prevent (the tool's own header cites LifeOS discussion #1461 — "capabilities that are assumed but never verified degrade silently").

## Fix

Single-file change to `LifeOS/SKILL.md`:

1. Add the `doctor` row to the routing table.
2. Add `doctor` to `argument-hint`.
3. Add doctor triggers to the `USE WHEN` list.
4. Add a `doctor` line to Examples.

`doctor` is the only **tool-backed** route — every other row points at a `Workflows/*.md`. Rather than add a workflow doc that would only say "run the tool and read the output" (dead scaffolding — `Doctor.ts` is already self-describing), the row points straight at the tool and a short note under the table explains why. The table's second column is renamed `Workflow` → `Target` to stay honest about that.

The note also carries Doctor's **declined-is-silent** contract, so the new route can't reintroduce nagging about capabilities the user has deliberately opted out of:

> Relay its table, offer the fix it names, and honor `decline` — a declined capability is a legitimate way to run LifeOS, never a defect to nag about.

The route target uses the `<configRoot>` placeholder, matching the convention `INSTALL.md` already uses throughout.

## Testing evidence

Claims in the new note are grounded in the shipped tool's source, not paraphrase:

- `LifeOS/install/LIFEOS/TOOLS/Doctor.ts:16` — `Four states: live | broken | declined | stale. Declined is first-class` / `and permanently silent — opted-out is not a defect.`
- `LifeOS/install/LIFEOS/TOOLS/Doctor.ts:31` — `bun Doctor.ts decline <cap>    # mark capability declined (silent forever)`
- `LifeOS/install/LIFEOS/TOOLS/Doctor.ts:28` — `bun Doctor.ts                  # probe (offline checks), table output`

Ran `bun LIFEOS/TOOLS/Doctor.ts` against an installed LifeOS instance: the tool executes clean, prints the capability table, and emits a concrete `fix:` line per broken capability — confirming the route target resolves and needs no workflow wrapper to be useful. (Output not pasted: it's a report of a specific machine's capability and key state.)

Docs-only change — no runtime code touched, nothing to regress in the installer.

## Notes

Deliberately left alone, as out of scope for this fix: `SKILL.md` still carries `version: 1.4.19` in frontmatter while its own Gotchas section says *"No `version:` in SKILL.md. Claude Code ignores it."* Happy to fix that separately if you want it.

### DIFF
--- a/LifeOS/SKILL.md
+++ b/LifeOS/SKILL.md
@@ -1,9 +1,9 @@
 ---
 name: LifeOS
 version: 1.4.19
-description: Install and onboard a user into LifeOS — the Life Operating System (current state → ideal state via TELOS + the Algorithm). The agentic installer detects your OS + harness, wires hooks with permission, scaffolds your USER tree, pulls in sources you provide, and runs the TELOS / current→ideal interview that seeds your Pulse dashboard. USE WHEN install LifeOS, set up LifeOS, lifeos setup, lifeos-setup, lifeos interview, onboard me, run the interview, integrate LifeOS into my harness, update LifeOS, uninstall LifeOS, first-time setup. NOT FOR building or cutting a LifeOS release (private release tooling), editing TELOS after onboarding (use Telos / Interview), or LifeOS system maintenance (use the private maintenance skill).
+description: Install and onboard a user into LifeOS — the Life Operating System (current state → ideal state via TELOS + the Algorithm). The agentic installer detects your OS + harness, wires hooks with permission, scaffolds your USER tree, pulls in sources you provide, and runs the TELOS / current→ideal interview that seeds your Pulse dashboard. USE WHEN install LifeOS, set up LifeOS, lifeos setup, lifeos-setup, lifeos interview, onboard me, run the interview, integrate LifeOS into my harness, update LifeOS, uninstall LifeOS, first-time setup, lifeos doctor, check my install, what capabilities are broken. NOT FOR building or cutting a LifeOS release (private release tooling), editing TELOS after onboarding (use Telos / Interview), or LifeOS system maintenance (use the private maintenance skill).
 disable-model-invocation: true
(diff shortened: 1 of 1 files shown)


erf1nd0r:
As a note, how I came up with this, is because after a fresh install I saw following message in the statusline and assumed that that the command for running the doctor should be /LifeOS doctor.

Screenshot 2026-07-13 at 12 09 35

danielmiessler:
Thanks @erf1nd0r — ported into source and verified.

The public LifeOS repo is a generated artifact (synced from a private tree at release), so a merge here would be overwritten on the next release. Instead I've applied your change at the source: the `doctor` route is now in the skill's description triggers, `argument-hint`, workflow-routing table, and examples, matching your diff.

Verified before merging the intent: `Doctor.ts` exists and emits exactly the four capability states your doc describes (live / broken / declined / stale) plus the fix command for anything broken. The routing text is accurate against the tool.

Closing since it's in the source now — it ships in the next release. Credited in the commit.

PR #1479, opened 2026-07-13
fix(telegram): self-explanatory permission denial — agents treat unexplained denials as bugs and probe workarounds

anikinsasha:
A field report turned one-liner. When a tool call gets denied without explanation, an agent reads the denial as breakage and starts probing workarounds — alternate tools, `sed` bypasses, retry loops. We watched a Telegram-side session do exactly this for 15 minutes against a working permission boundary before parking the task (incident documented in my #1143 comment thread).

A denial that explains itself converts the boundary from an obstacle the agent fights into an instruction it follows. This enriches the `canUseTool` deny message so the model is told, at the moment of failure: this is policy, working as intended — don't route around it, surface it to the principal.

No behavior change — nothing newly allowed or denied, message text only.

(Housekeeping note: I'm the author of #1457/#1458/#1459 — GitHub handle renamed `anikin-xyz` → `anikinsasha`, same account.)

### DIFF
--- a/LifeOS/install/LIFEOS/PULSE/modules/telegram.ts
+++ b/LifeOS/install/LIFEOS/PULSE/modules/telegram.ts
@@ -1002,7 +1002,7 @@ export async function startTelegram(config: TelegramConfig): Promise<void> {
               : String(input)
             if (cmd.includes("31337") || cmd.includes("/notify")) {
               log("warn", "canUseTool blocked /notify curl from SDK subprocess", { cmd: cmd.slice(0, 200) })
-              return { behavior: "deny", message: "Telegram mode: /notify and port 31337 are blocked. Voice is delivered via Telegram sendVoice, not the desktop speaker." }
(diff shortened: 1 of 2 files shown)


danielmiessler:
Thanks @anikinsasha — good field report, and the reasoning is right: an unexplained denial reads to an agent as breakage, and it burns time probing workarounds against a boundary that's working as intended.

Ported into source. The public repo is generated from a private tree at release, so I applied your one-line change directly to the source `telegram.ts` — the `canUseTool` `/notify` deny message now leads with "Denied by policy — working as intended, not a bug" and explicitly tells the model not to route around it and to surface the block to the principal instead. Message text only, no behavior change, exactly as you scoped it.

(Skipped the `PR_REVIEW_PACKAGE.md` that rode along — looks like PR meta rather than product.)

Closing since it's in the source now; ships next release. Credited in the commit. And noted the handle rename — same account, thanks for the continued fixes.

PR #1486, opened 2026-07-14
fix(ContextSearch): resolve ripgrep when no real rg binary exists

anikinsasha:
## Problem

ContextSearch's `ripgrep()` helper spawns `rg` via `Bun.spawn`. Claude Code's shell integration provides `rg` as a **shell function** wrapping the CLI's embedded ripgrep (`ARGV0=rg "$CLAUDE_CODE_EXECPATH"`), which `Bun.spawn` cannot see. On machines without a standalone ripgrep install this makes every rg-backed source (ISA bodies, conversation logs) die with:

```
ContextSearch error: Executable not found in $PATH: "rg"
```

— so the tool exits 1 on every query, silently losing two of its five sources' coverage. Easy to miss because machines with `brew install ripgrep` never hit it.

## Fix

Resolve the binary at spawn time:

1. A real `rg` on PATH (`Bun.which`) — used as before; zero behavior change for machines that have ripgrep.
2. Fallback: run the Claude Code executable (`CLAUDE_CODE_EXECPATH` or `~/.local/bin/claude`) with `argv0: "rg"` — the same mechanism the shell function uses to expose the embedded ripgrep.
3. Last resort: original `"rg"` literal, preserving the original error if neither exists.

## Testing

On a machine with no ripgrep binary (only the shell function): before the patch, every query exits 1 with the error above; after the patch, all five sources return results (verified against live `MEMORY/WORK` ISAs and conversation JSONLs, ~1s wall time). On a machine with real ripgrep, `Bun.which("rg")` short-circuits and behavior is byte-identical.

### DIFF
--- a/LifeOS/install/skills/ContextSearch/Tools/ContextSearch.ts
+++ b/LifeOS/install/skills/ContextSearch/Tools/ContextSearch.ts
@@ -299,10 +299,28 @@ function searchWorkDirs(tokens: string[], since: Date | null, until: Date | null
   return out.slice(0, 50);
 }
 
+/**
+ * Resolve a runnable ripgrep. On machines without a real `rg` binary, Claude
+ * Code's shell integration aliases rg to its embedded ripgrep via a shell
+ * FUNCTION — invisible to Bun.spawn, so every rg-backed source (ISA bodies,
+ * conversation logs) dies with "Executable not found in $PATH". Fall back to
+ * running the Claude Code executable with argv0="rg", the same trick the shell
+ * function uses.
+ */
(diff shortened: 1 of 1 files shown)


danielmiessler:
Ported to the private source tree essentially verbatim, with credit. The shell-function-invisible-to-`Bun.spawn` diagnosis was spot on, and the `argv0="rg"` trick mirroring what the shell integration does is clean. Verified ISA-body and conversation-log search working after the port.

Closing because the public repo is generated from private source at release time; your fix ships in the next release, credited in the commit. Thanks.

PR #1492, opened 2026-07-14
SMALL: StatusLine: per-section visibility toggles

thatsjet:
Every normal-mode section (header, state, effort, doctor, memory, versions, agents, context, usage, quote) gets a SHOW_* gate. Script defaults are all true — zero behavior change out of the box. Per-user overrides read from settings.json (preferences.statusline.sections) so they survive version upgrades that overwrite this script.

Hiding usage also skips its background OAuth/API fetch, not just the render. Separator lines are owned by their sections so hidden sections don't leave stacked rules. The settings.json eval is injection-safe: keys are regex- restricted to ^[a-z]+$ and values normalize to literal true/false.


danielmiessler:
Accepted for port. The design is right — script-level defaults with a settings.json override that survives version upgrades, and the sanitized jq eval (key regex + boolean normalization) holds up.

One honesty note on timing: our internal statusline has diverged a long way from the shipped one (~2400 lines now), so wiring the toggles through every section is its own careful pass. Leaving this open until it lands with your credit — you're already in the README [Credits section](https://github.com/danielmiessler/LifeOS/blob/main/README.md#credits).

PR #1511, opened 2026-07-16
feat(hooks): live per-turn time context (fixes stale session-start date)

anikinsasha:
## Problem

The harness injects `currentDate` as a **date-only** value anchored at **session start**. On a long-running session it goes stale: hours after midnight the assistant still reasons from the session-start date, and with only a date (no time) it can't distinguish morning from midnight. This causes small but real errors — greeting with the wrong time of day, scheduling against yesterday, "it's morning" at 1am.

## Fix

A small `UserPromptSubmit` hook (`TimeContext.hook.ts`) that recomputes the wall clock **every turn** and injects it as `additionalContext`:

```
<time-now>
🕐 Current time: Thu 2026-07-16 08:48 PDT (early morning)
(The harness date line is date-only and can be stale mid-session — trust this live value.)
</time-now>
```

Because it recomputes per turn, the value can never drift within a session.

## Notes

- **Timezone** comes from `settings.json` `principal.timezone` (the same source `lib/time.ts` reads), falling back to **UTC**. Read inline so a config problem can't break the hook.
- **Fails open**: any error is swallowed; the turn proceeds without the line. A timestamp is nice-to-have, never worth blocking a prompt.
- Registered in `hooks/hooks.json` under `UserPromptSubmit` (5s timeout), alongside the existing per-turn hooks.
- No new dependencies; ~40 lines.

Running on a live install. Happy to adjust the format or the relative-label thresholds.

### DIFF
--- a/LifeOS/install/hooks/TimeContext.hook.ts
+++ b/LifeOS/install/hooks/TimeContext.hook.ts
@@ -1,0 +1,44 @@
+#!/usr/bin/env bun
+/**
+ * @version 1.0.0
+ * TimeContext.hook.ts — inject a LIVE, per-turn wall-clock line on UserPromptSubmit
+ * so the assistant always knows the current time, date, and day of week.
+ *
+ * Why: the harness `currentDate` context is date-only and anchored at session start,
+ * so on a long-running session it goes stale — the model can believe it is still
+ * "morning" hours after midnight, or reason from yesterday's date. This hook
+ * recomputes the wall clock every turn and injects it as additionalContext, so the
+ * value can never drift within a session.
+ *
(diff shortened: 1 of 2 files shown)


danielmiessler:
Ported into source with credit — the file header cites you and this PR, and it's registered on UserPromptSubmit in the shipped template. Live-tested on our own install before closing.

Because the public repo is generated from a private source tree, community PRs get ported rather than merged (same fix, survives releases) — so closing this one, and it ships in the next release. Your rg-fallback from #1486 is already running in ContextSearch too. Two clean ones — you're in the README [Credits section](https://github.com/danielmiessler/LifeOS/blob/main/README.md#credits). Thank you.

PR #1512, opened 2026-07-16
feat(telegram): opt-in auto-purge of chat messages after N hours

anikinsasha:
## What

An **opt-in** privacy feature for the Telegram module: automatically delete the visible chat messages (both the user's and the bot's) after a configurable number of hours. **OFF by default** — existing installs are unchanged.

```toml
[telegram]
purge_enabled = true
purge_hours = 24   # default
```

## Why

The bot's real memory lives server-side in the `SessionStore` (and the conversation store), never in the Telegram chat history. The visible Telegram messages are just a UI surface, so for privacy many users will want them not to linger. This deletes them after `purge_hours` while losing zero context.

## How

- A durable ledger (`state/telegram/purge-ledger.json`) records `(chat, message_id, ts)` as messages flow: an incoming `bot.use` middleware for the user's messages, and a `bot.api.config.use` transformer that captures the `message_id` of **every** outgoing send (reply, voice, photo) centrally.
- A 30-minute loop (plus a boot sweep) deletes anything aged into `[purge_hours, ~48h)` and drops messages past that window. Bounded by the Bot API's hard **48h** `deleteMessage` ceiling.
- The ledger persists, so a restart resumes and past-due messages are swept on the next run.
- Fully gated on `purge_enabled`: `recordPurgeMsg`/`runPurge` are no-ops when off and the loop isn't scheduled. Zero behavior change for existing installs.

## Notes

- Running with `purge_enabled = true, purge_hours = 24` on a live install; verified end-to-end (a planted 25h-aged message is deleted on the next sweep).
- Known limit: out-of-band sends that bypass the bot's grammY instance (e.g. a raw `curl` to the Bot API) aren't recorded, so they aren't purged. All normal in-bot sends are captured via the transformer.
- +82 lines, no new dependencies.

### DIFF
--- a/LifeOS/install/LIFEOS/PULSE/modules/telegram.ts
+++ b/LifeOS/install/LIFEOS/PULSE/modules/telegram.ts
@@ -62,6 +62,12 @@ export interface TelegramConfig {
   // Opt out of Telegram voice-note summaries (PR #1458). undefined ⇒ on, so existing
   // installs are unchanged; set voice_summaries = false under [telegram] in PULSE.toml.
   voice_summaries?: boolean
+  // Auto-delete the visible Telegram chat messages (both sides) after purge_hours,
+  // for privacy. OFF by default (undefined ⇒ off, so existing installs are unchanged).
+  // The bot's memory lives server-side in the SessionStore, so purging the visible
+  // chat copies loses no context. Bounded by the Bot API's 48h delete window.
+  purge_enabled?: boolean
+  purge_hours?: number // default 24
 }
 
 // ── Constants ──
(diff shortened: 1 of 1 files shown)


danielmiessler:
Declining this one, and not on quality: we removed the Telegram integration entirely on 2026-07-15 (a retired-capability registry now gates it out of releases), so there's no surface for the purge to land on. Remote channels are iMessage and Siri for now.

The auto-purge idea itself is sound — if a messaging bridge comes back, opt-in retention limits like this are exactly what we'd want. Thanks for building it, and for #1511, which did land.

Issue #650, opened 2026-02-13
feat: Local patch tracking and update-safe SYSTEM file management


jlacour-git:
## Summary

PAI's SYSTEM/USER two-tier architecture (documented in `SYSTEM/SYSTEM_USER_EXTENDABILITY.md`) protects USER-tier customizations during updates. However, it doesn't address a common real-world scenario: **users who fix bugs in SYSTEM-tier files** before those fixes are merged upstream.

When a user patches a SYSTEM file locally (e.g., fixing a bug in `UpdateCounts.ts` or `statusline-command.sh`), that fix gets silently overwritten on the next PAI update. There's no mechanism to detect, warn about, or preserve these local modifications.

## Real-World Examples

**Example 1: UpdateCounts.ts** (see #642)
A user fixes the missing `work`/`research`/`ratings` counters. The fix lives in `hooks/handlers/UpdateCounts.ts` — a SYSTEM file. Next PAI release overwrites it with the old broken version.

**Example 2: Generated files with cached identity**
`SKILL.md` is generated by `CreateDynamicCore.ts` and caches `{DAIDENTITY.NAME}` at build time. After a PAI update that ships a new SKILL.md, the user's identity reverts to the default "PAI" instead of their configured name. There's no auto-rebuild trigger.

## Proposed Solution

Three lightweight mechanisms that extend the existing SYSTEM/USER architecture:

### 1. `LOCAL_PATCHES.md` Convention

A standardized file at `~/.claude/LOCAL_PATCHES.md` where users (or their DA) track local modifications to SYSTEM files:

```markdown
# Local Patches to PAI SYSTEM Files
## Active Patches
### UpdateCounts.ts - Missing counters
**File:** hooks/handlers/UpdateCounts.ts
**Issue:** #642
**PR:** _pending_
```

- Checked by the DA before modifying SYSTEM files
- Serves as a pre/post-update checklist
- Could be read by an install/update script to warn about conflicts

### 2. PAIUpgrade Conflict Check

Add a step to the PAIUpgrade Upgrade workflow (between scoring and output) that:
1. Reads `LOCAL_PATCHES.md` if it exists
2. Cross-references recommended techniques against locally-patched files
3. Flags conflicts in the upgrade report with a `⚠️ LOCAL PATCH CONFLICT` warning

This makes the upgrade workflow aware of local state, which is consistent with Thread 1's existing PAI State analysis.

### 3. Stale Generated File Detection

[...]

PR #1519, opened 2026-07-17
Update workflow: version diff never fetches — make step 2 a real release check

erf1nd0r:
## Problem

An installed user who runs the LifeOS skill's **Update** workflow always gets "already current", no matter how far behind their install is.

Step 2 says to compare "the release version being updated to" against the install marker — but nothing in the workflow ever *fetches* a release. The only version available on disk is the payload the skill itself shipped with, which is exactly what wrote the install marker. The comparison is self-referential, so it can never detect a newer release.

7.1.1 fixed the other half of this (`install.sh` now resolves the latest release dynamically), but no step tells an *existing* install to re-run the bootstrap — so the fix is unreachable from the update path. Hit in practice: an install on 7.0.0, with v7.1.1 published, reported "already current".

## Change (docs only, 2 files — the workflow + its nested payload mirror)

**Step 2 → a real release check:**
- installed version ← `<configRoot>/LIFEOS/VERSION` (the install marker, shipped since 7.1.1; absent → treat as behind)
- latest version ← `tag_name` from the same GitHub releases endpoint the bootstrap resolves against
- behind → run the shipped `install/install.sh` first (additive: replaces only the LifeOS skill dir, backs up the prior one), then continue from the **new** payload's workflow
- offline → degrade gracefully to re-applying the on-disk version
- plus an explicit note naming the anti-pattern (payload-vs-marker is not a release check)

**Step 3 additionally overwrites the `VERSION` marker** — the copyMissing deploys never touch an existing file, so without this the marker stays stale and step 2 re-trips on every future update.

## Possible follow-ups (intentionally not bundled)

- **Doctor.ts update awareness** — a capability-style line under `--network` ("installed 7.0.0, latest 7.1.1") would fit 7.1.1's install-awareness direction, but mapping "update available" onto the live/broken/declined states is a maintainer call, so it's left out here.
- **Changed-file delivery** — copyMissing delivers *new* files but leaves *changed* system-owned files stale after an update (adjacent to the concern in #650). Worth its own design pass; out of scope for this doc fix.

### DIFF
--- a/LifeOS/Workflows/Update.md
+++ b/LifeOS/Workflows/Update.md
@@ -12,8 +12,15 @@ curl -s -X POST http://localhost:31337/notify -H "Content-Type: application/json
 ## Steps
 
 1. **DetectEnv** — `bun Tools/DetectEnv.ts`. If `isDevTree` → STOP (the source repo updates itself via git, not this workflow).
-2. **Version diff** — the skill carries no version field and there is no plugin manifest; versioning lives at the distribution layer (the GitHub release tag + `LIFEOS_RELEASES/<version>/` + the `install.sh` fetch's `LIFEOS_VERSION`). Compare the release version being updated to against the user's current install marker. If equal, report "already current" and exit.
-3. **Re-overlay system** — re-copy the system templates (CLAUDE, system prompt, `settings.system.json` minus hooks). These are system-owned and safe to overwrite.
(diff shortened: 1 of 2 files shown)


danielmiessler:
Ported into source exactly as written — step 2 is a real release check and step 3 overwrites the VERSION marker; the workflow text credits you. Closing the PR because the public repo is release-generated (merge-back doctrine); your fix ships in the next release. The anti-pattern note you embedded ("payload-vs-marker is not a release check") is the kind of line that prevents regressions — appreciated.

PR #1535, opened 2026-07-18
feat: LIFEOS_DOWNLOADS_DIR — one knob for where generated artifacts land (+ Capture.sh stray-file fix)

anikinsasha:
## What this adds

A single opt-in env var, **`LIFEOS_DOWNLOADS_DIR`**, that controls where every LifeOS skill stages its generated artifacts — screenshots, images, thumbnails, video renders, web-design exports. Unset, everything behaves exactly as today (`~/Downloads`). Set (one line in `settings.json`'s `env` block, which Claude Code exports to every tool session), all output routes to the chosen directory:

```json
{ "env": { "LIFEOS_DOWNLOADS_DIR": "/path/to/your/downloads" } }
```

Motivation: operators who keep a dedicated workspace (synced drives, external volumes, tidy-home setups) currently have to override every skill's `~/Downloads` habit by prompt discipline. This makes it mechanical.

## Commit 1 — code

1. **`skills/Interceptor/Tools/Capture.sh`** — two things:
   - **Stray-file leak fix.** The `interceptor` CLI has no output-path flag: `--save` always writes `interceptor-screenshot-<epoch>.png` into the CLI process CWD, and unknown flags are silently swallowed — so Capture.sh's `--out "$OUT"` never reached the CLI, and its `resolve_saved()` reconciliation (`cp -f`) delivered the target while leaving a full-size duplicate in the calling agent's CWD on **every capture** (invisible because `.gitignore` ignores the pattern). Now every capture runs inside a private `mktemp -d` stage (trap-cleaned on all exit paths) and the CLI-reported `filePath` is **moved**, not copied, to the destination — CLI-version-proof, DOM-render and `--pixel` alike.
   - Default destination becomes `${LIFEOS_DOWNLOADS_DIR:-$HOME/Downloads}`; an unwritable destination is a hard failure (new exit 13) with remediation text, never a CWD fallback; `--selector <css>` / `--region X,Y,W,H` now pass through (the CLI supports both; the wrapper rejected them).
   - Preserved unchanged: preflight gate, target-deny checks, blank-frame degeneracy guard (still exit 12), wedge-recovery ladder, single-path-on-stdout contract, existing exit codes.
2. **`skills/Art/Tools/Generate.ts`, `ComposeThumbnail.ts`** — default `output:` honors `process.env.LIFEOS_DOWNLOADS_DIR ?? ${HOME}/Downloads` (usage text updated; `ThumbnailText.ts` comment example aligned).

## Commit 2 — docs

Mechanical sweep of every *output-destination* `~/Downloads` mention to the new contract (`$LIFEOS_DOWNLOADS_DIR`, default `~/Downloads/` when unset; command examples use `"${LIFEOS_DOWNLOADS_DIR:-$HOME/Downloads}"`):

- `LIFEOS/DOCUMENTATION/Skills/SkillSystem.md` (the "Output: Always ~/Downloads/ first" doctrine line)
- Interceptor: `SKILL.md` (also documents the CLI's silent-flag-swallowing gotcha and the staging behavior), `Workflows/ScreenshotForVlm.md`
- Art: `SKILL.md` + 5 workflows
- Remotion: `SKILL.md`, `CriticalRules.md`, 2 workflows, `Tools/Ref-ai-pipeline.md`
- Webdesign: 4 workflows

Deliberately untouched: input-file examples (AudioEditor's `Pipeline.ts ~/Downloads/podcast.mp3`, BiasCheck's "bias check ~/Downloads/report.pdf", RemoveBackground's input example) — users' downloaded files legitimately live in `~/Downloads`. Template placeholders (`{{DA_NAME}}`) preserved.

## Verification (run live, macOS, interceptor 0.22.6, isolated test profile)

- Var set → Capture.sh default-out capture lands in the configured dir; zero `interceptor-screenshot-*` strays anywhere; staging dir removed.
- Var unset (fake `$HOME`) → capture lands in `$HOME/Downloads` — stock behavior byte-for-byte, proven end-to-end.
- `--pixel` path → delivered via mv-reconciliation, zero strays.
- Art tools: the exact default-expression evaluated under bun with and without the var → correct both ways; both files transpile clean (`bun build`).
- Read-only destination → exit 13; bad args → exit 2 (unchanged).
- `patch -p1 --dry-run` against a pristine 7.1.1 tree → applies clean.

## Notes for reviewers

- Stock permission rules (`Write(~/Downloads/**)`) are untouched; operators pointing the var at a directory outside their working tree may need a matching allow rule.
- The knob is read at call time in shell (`${VAR:-fallback}`) and at module load in the TS tools — no new config file, no new dependency.
- `mv` across filesystems degrades to copy+unlink, so cross-volume destinations work.


danielmiessler:
Splitting this one. The Capture.sh stray-file fix is ported with credit: `resolve_saved()` now moves (not copies) the daemon-chosen file, verified live on both DOM and `--pixel` paths with zero strays — that was a real, reproducible leak and a great catch. The `LIFEOS_DOWNLOADS_DIR` knob itself is under consideration as a system-wide env contract (it touches doctrine across several skills), so leaving this PR open until that call is made. Either way the stray fix ships in the next release with your name on it.

danielmiessler:
Fully ported now — both halves. The stray-file fix landed earlier (mv-reconciliation in `resolve_saved`, verified live on DOM and `--pixel` paths, zero strays). The `LIFEOS_DOWNLOADS_DIR` knob is now in too: `Capture.sh` default destination, `Generate.ts` and `ComposeThumbnail.ts` defaults, plus the mechanical doc sweep (17 files, 104 output-destination substitutions, your input-example carve-outs honored). Verified end-to-end: var set → capture lands in the configured dir; var unset → byte-identical stock behavior. Credit comments cite you at every code site and you're in the README Credits. Closing for the usual mechanical reason — the public repo is release-generated, so ports land in source and ship with the next release. This was a model PR: verified claims, honest scope notes, and the silent-flag-swallowing discovery was a genuinely good catch.

Issue #1539, opened 2026-07-19
Installer ships 39 hook scripts and registers 28; nothing compares the two inventories (follow-up to #1508)


tzioup:
# The installer ships 39 hook scripts and registers 28; nothing compares the two inventories (follow-up to #1508)

> Structured for machine parsing: claims are atomic, and each carries the command that verifies it. Every command runs against a fresh clone of this repo — none depend on my machine, my install, or my configuration. Reproduction is at the bottom if you'd rather run it than read it.

## Summary

`InstallHooks.ts` performs two operations and verifies each independently:

1. `cpSync` the entire payload `hooks/` tree onto disk — verified by a file count (`hookFiles` vs `hookFilesCopied`).
2. Merge `hooks/hooks.json` into `settings.json` — verified by reporting `added`, `skipped`, `events`.

Both succeed. Neither is wrong. **Nothing compares the set of files copied against the set of hooks registered.** 39 hook scripts land on disk; 28 are named by the manifest merged beside them. The install reports success accurately, because "validated" here means "both operations completed," never "the things I copied are the things I registered."

Of the 11-hook difference, 8 are legitimately loaded via dispatcher imports. **Three are unreachable: `DriftReminder`, `PostToolObserver`, and `LoopDetector`** (whose only importer is `PostToolObserver`).

`Doctor.ts --reconcile` — which you shipped in v7.1.1 for exactly this, and which the same install run copies onto disk — reports all three correctly in under a second. The installer does not call it.

## Claims and verification

Run from the root of a fresh clone.

**C1 — the payload ships 39 hook scripts.**
```bash
ls LifeOS/install/hooks/*.hook.ts LifeOS/install/hooks/*.hook.sh | wc -l
# 39
```

**C2 — the manifest registers 28 distinct hooks.**
```bash
grep -oE '[A-Za-z]+\.hook\.(ts|sh)' LifeOS/install/hooks/hooks.json | sort -u | wc -l
# 28
```

**C3 — 11 shipped hooks are absent from the manifest.**
```bash
comm -23 \
  <(ls LifeOS/install/hooks/*.hook.ts LifeOS/install/hooks/*.hook.sh | xargs -n1 basename | sort) \
  <(grep -oE '[A-Za-z]+\.hook\.(ts|sh)' LifeOS/install/hooks/hooks.json | sort -u)
# 11 lines
```

**C4 — of those 11, two have no importer anywhere; they are unreachable by any path.**
```bash
comm -23 \
[...]

PR #1540, opened 2026-07-19
InstallHooks: reconcile copied hooks against registered hooks after apply

tzioup:
Fixes the detection gap in #1539.

## What this changes

`InstallHooks.ts` does two independent things and verifies each on its own: it `cpSync`s the payload `hooks/` tree onto disk (checked by file count), and it merges `hooks.json` into `settings.json` (checked by `added`/`skipped`/`events`). Both succeed. Nothing compares the set of files copied against the set of hooks registered, so a hook can land on disk with no settings entry naming it and the install still reports success — accurately.

After `--apply`, this shells `Doctor.ts --reconcile` (already on disk at that point, copied by the same install) with `CLAUDE_CONFIG_DIR` pointed at the target config root, and includes its `unwired` list in the existing report JSON.

No logic is duplicated — the reconciler you shipped in v7.1.1 already does the transitive closure through dispatcher imports correctly. This just asks it.

## Behaviour

On a clean v7.1.1 payload into an empty config root:

```json
{
  "ok": true,
  "added": 45,
  "events": 11,
  "hookFilesCopied": 72,
  "unwired": [
    "LoopDetector.hook.ts",
    "DriftReminder.hook.ts",
    "PostToolObserver.hook.ts"
  ]
}
```

If `Doctor.ts` isn't present at that path, the report carries a `reconcile` string saying so rather than skipping silently — matching the "LOUD blocker, never a silent no-op" rule in `Workflows/Setup.md`.

## Testing

Ran the patched installer against a scratch config root containing only `settings.json` (`{}`) and `LIFEOS/TOOLS/Doctor.ts`, with `--skill-root` pointed at this repo's `LifeOS/`. Output above is verbatim from that run. Existing report fields are unchanged; the two new keys are additive.

## Note

`unwired` is reported, not enforced — this doesn't fail the install or change exit codes. Happy to make it a blocker instead, or to drop the subprocess and inline the comparison, whichever you'd prefer.

### DIFF
--- a/LifeOS/Tools/InstallHooks.ts
+++ b/LifeOS/Tools/InstallHooks.ts
@@ -15,6 +15,7 @@
  */
 
 import { copyFileSync, cpSync, existsSync, mkdirSync, readdirSync, readFileSync, statSync, writeFileSync } from "node:fs";
+import { execFileSync } from "node:child_process";
 import { join } from "node:path";
 import { detectDevTree, mergeHooks } from "./InstallEngine";
 
@@ -101,7 +102,28 @@ function main(): void {
   cpSync(hooksPayloadDir, hooksDestDir, { recursive: true });
   const hookFilesCopied = countFilesRec(hooksDestDir);
 
-  console.log(JSON.stringify({ ...report, written: true, backup, hookFilesCopied }, null, 2));
+  // Copying hook scripts and merging the manifest are two independent operations;
+  // a hook can land on disk without any settings entry naming it. Ask the shipped
+  // reconciler whether every copied hook is actually reachable, so an unwired hook
+  // is loud at install time rather than silent until someone runs Doctor by hand.
+  const doctorPath = join(configRoot, "LIFEOS", "TOOLS", "Doctor.ts");
+  let unwired: string[] | undefined;
+  let reconcile: string | undefined;
+  if (existsSync(doctorPath)) {
+    try {
+      const out = execFileSync("bun", [doctorPath, "--reconcile"], {
+        encoding: "utf-8",
+        env: { ...process.env, CLAUDE_CONFIG_DIR: configRoot },
+      });
+      unwired = JSON.parse(out).unwired ?? [];
+    } catch (err) {
+      reconcile = `reconciler failed: ${err instanceof Error ? err.message : String(err)}`;
+    }
+  } else {
+    reconcile = `Doctor.ts not found at ${doctorPath} — hooks copied but not reconciled`;
+  }
+
+  console.log(JSON.stringify({ ...report, written: true, backup, hookFilesCopied, unwired, reconcile }, null, 2));
   process.exit(0);
 }
 


tzioup:
A generalization this PR naturally extends to — recording it here rather than opening a separate issue, since the machinery is identical and already shipped.

**Three pieces already exist; this PR connects the first two, for hooks. They generalize:**

1. **Per-component provenance is already computed, then discarded.** `DeployComponents.ts` builds a `ComponentResult { component, actions[], applied, blockers[] }` per component; `InstallHooks.ts` builds `report { added, skipped, events, hookFiles }`. Both are `console.log`'d and dropped — never persisted to disk.

2. **A durable, typed, tamper-evident manifest already exists.** `Doctor.ts` writes `LIFEOS/MEMORY/STATE/capabilities.json` (`interface Manifest`, `Doctor.ts:47/63/263`). Today it records external-capability *liveness* (`codex`/`interceptor`/`cloudflare`/`voice` → live/broken/declined/stale), not deployment *provenance*.

3. **The reconciler this PR calls is hooks-only by design** — `reconcileHooks()` is explicitly annotated `v1 scope: hooks only, two inventories` (`Doctor.ts:344`).

**Two low-cost generalizations, both building only on the above:**

- **Persist the provenance.** Fold the `ComponentResult` / `report` the deploy tools already compute into `capabilities.json` (or a sibling `install-manifest.json`) instead of discarding it. An install then carries a durable record of *what was deployed, what was skipped, and why* — which is what lets a later discrepancy be attributed (shipped-broken vs onboarding-declined vs onboarding-failed) instead of re-diagnosed from scratch on each machine.

- **Generalize the reconcile beyond hooks.** This PR wires `Doctor --reconcile` after hook apply; the same declared-vs-deployed check applies to every component `DeployComponents` places (statusline, agents, commands, launchd, pulse). The hook slice is exactly what this PR already does — this is the same move applied to (a) all components and (b) a persisted record.

No new subsystem is required; all three pieces ship today.

*Out of scope here, noted for completeness:* the broader class of "entries removed from **inside** a surviving config/doc structure" (e.g. a registration deleted from an array that otherwise stays valid) is wider than the two-inventory `reconcileHooks` covers, and a declared-vs-registered reconcile won't catch all of it. Closing that class needs either configs generated-from-source (so a hand deletion regenerates or fails loudly) or a broader reconciler — a separate, larger design question, not something to fold into this PR.

danielmiessler:
Ported into source, thank you. One note on mechanics: the local InstallHooks had diverged from the shipped copy (it already counts pre-existing files and skips collisions), so your diff was re-applied as intent onto that base rather than verbatim; behavior is exactly your design, reconcile after apply, `unwired[]` in the report, loud when Doctor.ts is absent. The provenance and generalize-beyond-hooks ideas from your follow-up comment are noted as separate design work. The public repo is generated at release, so this closes as ported with README credit. Ships with the next release.

Issue #1541, opened 2026-07-19
ReferenceCheck.ts finds 223 missing references in the shipped payload, has no invoker, and its orphan detection inspects one file


tzioup:
# ReferenceCheck.ts finds 223 missing references in the shipped payload, has no invoker, and its orphan detection inspects one file

> Structured for machine parsing: claims are atomic, each carries the command that verifies it. Every command runs against a fresh clone of this repo — none depend on my machine, my install, or my configuration. Reproduction at the bottom.

## Summary

`LIFEOS/TOOLS/ReferenceCheck.ts` is a full-surface reference validator. Run unmodified against the shipped payload it reports **223 missing references in 231ms**. Nothing invokes it — no hook, no command, no CI job. Its orphan mode reports 0 on every run because the check inspects exactly one file out of 1,324.

This is the same shape as #1539: the diagnostic exists, it is correct, and nothing asks it anything.

## Claims and verification

`$PAYLOAD` below means the release payload staged in the shape it ships as. `ReferenceCheck` resolves its root as `$HOME/.claude`, so:

```bash
git clone --depth 1 https://github.com/danielmiessler/LifeOS && cd LifeOS
mkdir -p /tmp/rc && cp -R LifeOS/install /tmp/rc/.claude
```

**B1 — the shipped payload contains 223 missing references.**
```bash
cd /tmp/rc/.claude/LIFEOS/TOOLS && HOME=/tmp/rc bun ReferenceCheck.ts --quiet | tail -1
# ReferenceCheck: 1174 files, 1558 refs, 223 missing, 0 stale, 0 orphan — 231ms
```

**B2 — roughly half are legitimate; the rest are not.** Classification of the 223:

| Class | Count | Defect? |
|---|---|---|
| Runtime-created state and logs (`MEMORY/STATE`, `OBSERVABILITY`, `DATA`, `LEARNING`, `RESEARCH`) | 90 | No — created on first use |
| `USER/` tree | 28 | No — populated at interview |
| References to private skills not in the release | 35 | **Yes** |
| References to dated `MEMORY/WORK/<slug>/ISA.md` artifacts | 13 | **Yes** |
| Other | 57 | Mixed |

```bash
grep -cE 'skills/_[A-Z]' /tmp/out.txt        # 35
grep -cE 'MEMORY/WORK/' /tmp/out.txt         # 13
[...]

PR #1542, opened 2026-07-19
ci: run ReferenceCheck against the shipped payload (report-only)

tzioup:
Makes the number in #1541 visible on every PR that touches the payload.

## What this does

`ReferenceCheck.ts` resolves its root as `$HOME/.claude`, so the job stages `LifeOS/install` in the shape it ships as and runs the validator there. No source change required.

```yaml
- run: cp -R LifeOS/install "$HOME/.claude"
- run: cd "$HOME/.claude/LIFEOS/TOOLS" && bun ReferenceCheck.ts --quiet
```

Triggers on `pull_request` and `push` to main, both filtered to `LifeOS/install/**`, plus `workflow_dispatch`.

## Report-only, deliberately

`continue-on-error: true`. It prints the summary line to the job summary and does not gate.

Against the current payload it reports:

```
ReferenceCheck: 1174 files, 1558 refs, 223 missing, 0 stale, 0 orphan — 231ms
```

Of those 223, roughly 118 are expected: 90 runtime-created paths under `MEMORY/{STATE,OBSERVABILITY,DATA,LEARNING,RESEARCH}` and 28 `USER/` files populated at interview. Gating on the raw count would be more than half false positives, and a check that cries wolf gets switched off — so this lands as report-only.

Making it blocking needs an ignore-list for runtime-created paths first. Happy to do that as a follow-up if you want it, or to fold it into this PR.

## Testing

Ran both `run` steps locally against a fresh clone at `4e33cc8` — staged the payload into a scratch `$HOME` and executed the validator, output above is verbatim. Workflow YAML parsed and validated.

I haven't been able to exercise it as a real Actions run (it has to be on the repo to trigger), so the first run on merge is the real test. `oven-sh/setup-bun@v2` is the only external action beyond `actions/checkout@v4`; swap it if you'd rather not add a dependency — `curl -fsSL https://bun.sh/install | bash` works equally well.

### DIFF
--- a/.github/workflows/reference-check.yml
+++ b/.github/workflows/reference-check.yml
@@ -1,0 +1,45 @@
+name: Reference Check
+
+on:
+  pull_request:
+    paths:
+      - 'LifeOS/install/**'
+  push:
+    branches: [main]
+    paths:
+      - 'LifeOS/install/**'
+  workflow_dispatch:
+
+jobs:
+  reference-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: latest
+
+      # ReferenceCheck resolves its root as $HOME/.claude, so stage the release
+      # payload in the shape it ships as. No source change required.
+      - name: Stage payload as an install
+        run: cp -R LifeOS/install "$HOME/.claude"
+
+      - name: Run ReferenceCheck against the shipped tree
+        id: refcheck
+        continue-on-error: true
+        run: |
+          cd "$HOME/.claude/LIFEOS/TOOLS"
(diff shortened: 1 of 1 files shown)


danielmiessler:
Declining this one, with appreciation for how carefully it was scoped. Two reasons. First, reference integrity already has teeth where we want them: a blocking release-build gate in the private pipeline runs ReferenceCheck before any staging is cut, so broken refs stop a release rather than reporting after a push. Second, by your own measurement ~118 of the 223 findings are expected runtime-created paths, and a permanently-half-red report-only workflow trains people to ignore CI. If a runtime-path ignore-list lands and the signal becomes clean, a public workflow becomes attractive and we'd revisit. The case fix from the companion issue is in source. Thank you for the patch either way.

PR #1544, opened 2026-07-19
Pulse: expand ${VAR} across all PULSE.toml string values, not just job.command

m8ryx:
`PULSE.toml` declares environment references in three places — `${NTFY_TOPIC}`, `${DISCORD_WEBHOOK}`, `${TWILIO_TO_NUMBER}` — but `resolveEnvVars()` is only applied to a job's `command`, so none of them are expanded. Consumers currently work around this by reading `process.env` directly in `dispatchSingle()`, so nothing is broken at runtime today; the config simply doesn't behave the way it reads.

Any module that trusts the config instead receives the literal string `"${VAR}"`. For a credential that means an authentication failure with no local error.

This moves expansion into the config layer via a new `parseConfigToml()`, so both readers of `PULSE.toml` — the job loader in `lib.ts` and `loadPulseConfig()` in `pulse.ts` — go through it, and the declared convention becomes real.

Substitution semantics are deliberately unchanged from the existing single-field behaviour: both `$VAR` and `${VAR}` are recognised, and an undefined variable resolves to the empty string. The pattern only matches `$` followed by an uppercase identifier, so a plain literal value (a hostname, a cron expression, a prose prompt) is returned byte-identical and existing installs are unaffected.

Includes tests covering nested-section expansion, literal pass-through, non-string scalars, `job.command` still expanding, undefined-variable semantics, and keys never being rewritten.

Note this is the repo's first tracked test file — there's no existing test convention or runner config, so it's colocated and run with `bun test`. Happy to move it if you'd prefer a different location or setup.

### DIFF
--- a/LifeOS/install/LIFEOS/PULSE/lib.test.ts
+++ b/LifeOS/install/LIFEOS/PULSE/lib.test.ts
@@ -1,0 +1,86 @@
+/**
+ * LifeOS Pulse — lib.ts config-layer tests
+ *
+ * Covers parseConfigToml()'s ${VAR} expansion: that env references resolve
+ * anywhere in the document, and — just as important — that plain literals are
+ * returned untouched so an existing install does not break on upgrade.
+ *
+ * Run from LIFEOS/PULSE (requires `bun install` for smol-toml):
+ *   bun test lib.test.ts
+ *
+ * Every credential-shaped value here is fake. Never put a real token in a test.
+ */
+
+import { test, expect } from "bun:test"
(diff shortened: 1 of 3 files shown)


danielmiessler:
Ported into source, including your test file, which has the honor of being the repo's first tracked test suite (7/7 green). The uppercase-identifier regex keeping literals byte-identical is what made this an easy yes: existing installs can't break on upgrade. Your `dispatchSingle` observation is right too; migrating its direct `process.env` reads onto the now-live config values is queued as a follow-on. The public repo is generated at release, so this closes as ported with README credit. Ships with the next release.

PR #1545, opened 2026-07-19
feat(install): support isolated custom LifeOS homes

sbusc:
## Summary

LifeOS can now be installed into an isolated custom home instead of always using the harness-wide `~/.claude` directory.

```bash
bash install.sh --home ~/MyProject/.claude
# or
LIFEOS_HOME=~/MyProject/.claude bash install.sh
```

This keeps plain Claude Code sessions vanilla while allowing users to opt into LifeOS deliberately, or run multiple isolated LifeOS instances without sharing settings, hooks, runtime state, or USER data.

## Why

Installing into `~/.claude` makes LifeOS global: its settings, hooks, skills, and `CLAUDE.md` affect every Claude Code session.

`CLAUDE_CONFIG_DIR` alone is not a suitable installer override because it relocates the harness's entire configuration and changes global/project merging behavior. `LIFEOS_HOME` controls where LifeOS is installed while preserving an explicit runtime launch strategy.

Previously, custom installs also required manually repairing hardcoded `~/.claude` paths after every update.

## What changed

- Added canonical root resolution with this priority:
  `--config-root` → `LIFEOS_HOME` → `CLAUDE_CONFIG_DIR` → persisted `dirname(LIFEOS_DIR)` → `~/.claude`.
- Non-default roots receive an isolated `<configRoot>/USER-data` data home.
- Plumbed `configRoot` and `configDir` through the complete Setup and Update workflows.
- Retargeted settings, permission globs, hook commands, statusline commands, and generated operational commands to the resolved root.
- Added shell-safe quoting for custom roots containing spaces or metacharacters.
- Added root-aware runtime resolvers for PULSE, TOOLS, hooks, services, and selected skills.
- Standardized model-read Markdown on semantic `{{LIFEOS_ROOT}}`, `{{LIFEOS_DIR}}`, and `{{LIFEOS_CONFIG_DIR}}` placeholders while keeping executable shell snippets on quoted `${LIFEOS_*}` variables.
- Grounded those semantic placeholders at every SessionStart, including subagent sessions, and added a PreToolUse fail-safe that rejects unresolved placeholders in path-bearing tool inputs.
- Made the bootstrap discover `/lifeos-setup` from arbitrary caller directories.
- Updated the `lifeos` launcher:
  - project-scoped `<project>/.claude` roots launch from the project parent and preserve merging;
  - arbitrary custom roots launch with `CLAUDE_CONFIG_DIR=<configRoot>`.
- Kept the authoritative setup skill and its shipped payload copy byte-for-byte synchronized.

### `LIFEOS_CONFIG_DIR` self-symlink collision

The shipped settings used `LIFEOS_CONFIG_DIR=$HOME/.claude/LIFEOS`, while `ScaffoldUser` and `LinkUser` interpreted the same variable as the private USER-data home.

An installer rerun could therefore resolve the data target to the live `LIFEOS/USER` directory itself.

The installer now:

- treats `LIFEOS_CONFIG_DIR` as the private data home;
- repairs the known legacy value during an update;
- keeps a final separation guard for explicitly unsafe `--config-dir` values.

### Lifecycle recovery

Later Setup or Update sessions no longer depend on the bootstrap shell's transient environment. They recover the custom root from persisted `LIFEOS_HOME` or `LIFEOS_DIR`.

### Instruction-layer path resolution

Claude Code does not natively substitute arbitrary `{{...}}` tokens in Markdown loaded through tools. `LoadContext.hook.ts` now places the three authoritative LifeOS paths into the model context at SessionStart and tells the model to resolve them before Read/Edit/Write/Glob/Grep/Bash calls.

`PreToolGuard.hook.ts` blocks unresolved LifeOS placeholders in path-bearing fields before they can reach the filesystem, while still allowing placeholder examples inside document content. This prevents a custom-home session from silently falling back to or contaminating the real `~/.claude` tree.

### Clean-machine installation

`InstallSettings` and `InstallHooks` now create a missing config root before writing, allowing fresh custom and default installations.

## Payload and template hygiene

The repository shipped a second copy of the LifeOS setup skill under `LifeOS/install/skills/LifeOS/`. The integration suite now verifies parity for all duplicated installer tools, workflows, templates, and settings.

The Telos Dashboard template also contained case-only duplicate trees (`App`/`app`, `Lib`/`lib`, and multiple UI variants). The removed files were verified as identical Git blobs; canonical lowercase Next.js/shadcn paths remain.

The template now:

- type-checks on case-sensitive systems;
- builds all 11 routes;
- includes a synchronized `bun.lock`;
- passes `bun install --frozen-lockfile`.

## Backwards compatibility

Without `LIFEOS_HOME`, LifeOS continues to install into `~/.claude`.

Default permission globs, hook command forms, and install locations remain unchanged. The intentional compatibility migration is the correction of the ambiguous legacy `LIFEOS_CONFIG_DIR` value.

## Validation

- `bun LifeOS/Tools/InstallIntegrationTest.ts` — **93/93 passed**
- Full custom-home installation into a throwaway fake `$HOME`
- Root containing spaces
- Bootstrap handoff to a fake Claude binary
- Fresh launcher process with root environment scrubbed
- Later-session recovery from persisted `LIFEOS_DIR`
- Default-install regression
- Legacy settings repair and explicit self-symlink refusal
- SessionStart grounding for main and subagent contexts, including stale/poisoned environment recovery
- PreToolUse rejection of unresolved path placeholders without blocking documentation content
- Markdown convention audit for semantic placeholders versus executable shell variables
- launchd and systemd Pulse rendering
- Complete setup-skill payload parity
- `bun install --frozen-lockfile` — passed for the Telos Dashboard template
- `bun run build` — passed; all 11 Dashboard routes compiled and prerendered
- Changed TypeScript and shell files passed syntax validation
- `git diff --check` — clean

## Scope note

The diff is large because the runtime and instruction-layer path refactor is mechanical across PULSE, TOOLS, hooks, services, agents, documentation, and skills, and because the Telos cleanup removes 27 case-only duplicate files. The behavioral change is centered on one invariant: every installer, runtime component, and model instruction resolves the same LifeOS root.


danielmiessler:
This is the most substantial contribution the repo has received, and we're not merging it, which deserves a real explanation. The public repo is generated from a private source tree at release; anything merged here is overwritten by the next release, so large features have to be authored in source. And this one is a system-wide change to path resolution and hook context injection, the surface where a regression breaks every session, so it needs to land incrementally with the release pipeline and settings-merge model in the loop. What we're keeping: the design. LIFEOS_HOME as the env contract, the CLAUDE_CONFIG_DIR rationale, the self-symlink collision fix, and the payload-parity integration suite are the blueprint for how isolated homes will land, and you're credited in the README now, with fix-site credit when the implementation ships. If you're up for it, the highest-leverage next contribution would be the integration suite as a standalone piece, since it's valuable independent of the path refactor.

Issue #1561, opened 2026-07-21
Harvest: add an eval before adopting behavior-changing candidates


rathko:
## The problem

Harvest is report-only, which is right. But when a candidate changes how the system behaves (Algorithm mechanics, a skill's workflow logic, hooks, routing policy), the only thing standing between "sounds great" and "adopted" is a review. A review can tell you a design is internally broken. It cannot tell you the new mechanism beats the one you already run. Without a measurement step, harvests slowly fill the system with plausible ideas that measure worse than what they replaced. That is where bloat comes from: every exciting blog post becomes a candidate, and enthusiasm is the only gate.

## What happened to me

I harvested multi-model "plan fusion" into my LifeOS-derived fork (from the Devin Fusion, OpenRouter fusion-beats-frontier and aider architect/editor posts): generate two plans with different frontier models at the commitment boundary, then have a judge synthesize one. On paper it looked like an obvious win. It survived a 10-concern adversarial cross-model design review after I folded every concern into a v2.

Then I ran an offline A/B before adopting:

- 7 historical high-effort ISAs as the task set
- ground truth from ISA Changelogs: `conjectured -> refuted_by` entries, the assumptions each real run only discovered were wrong late. A plan scores when it anticipates those pitfalls up front
- cost parity (3 model calls per arm), authorship-blind pairwise judging with position swap
- decision rule written down before any results existed: adopt at 0.65 win share or better, reject under 0.45

Fusion won 1 of 14 judgments and anticipated 3 fewer real pitfalls than the current plan -> advisor critique -> revise loop. Rejected. Seven tasks is a small sample and plans were judged rather than executed, so I treat the number as directional. Without the eval I would have shipped a measurably worse mechanism into the most leveraged part of the Algorithm, and the review had already blessed it.

## Proposal

Add a Step 7 to `skills/Harvest/Workflows/Harvest.md`, between Report and Ingest (Ingest becomes Step 8):

> ## Step 7 — Eval before adoption (behavior-changing candidates only)
>
[...]

PR #1562, opened 2026-07-21
Harvest: add eval-before-adoption step for behavior-changing candidates

rathko:
Behavior-changing candidates get a recommended offline comparison
 against the current mechanism, with the decision rule written down
 before results exist. Cosmetic candidates skip with a one-line note.
 Ingest renumbered to Step 8.

Addresses: https://github.com/danielmiessler/LifeOS/issues/1561


danielmiessler:
Ported in adapted form; see #1561 for the reasoning. The eval-before-adoption step is now in the Harvest mining workflow as an elected option for behavior-changing candidates rather than a near-mandatory gate, which keeps it inside our standing "evals are elected, not required" rule while preserving the idea's teeth: pre-registered decision rule, offline A/B against current behavior. Credited in the workflow and README. The public repo is generated at release, so this closes as ported rather than merged. Thank you for a genuinely well-argued proposal.

Issue #1496, opened 2026-07-14
Six bugs found during a fresh 7.1.1 install (TELOS parser, Doctor.ts, hook wiring, timezone default)


waveman2020-sudo:
# Bug report: multiple issues found during a fresh LifeOS 7.1.1 install

## 1. `GenerateTelosSummary.ts` — YAML frontmatter swallowed as a bogus first entry

`readTelosFile()` reads raw file content including leading YAML frontmatter. When a source file (e.g. `MISSION.md`) has no bullet/heading-ID items, `parseItems()` falls back to `paragraphItems()`, which splits on blank lines. The frontmatter block (`---\nkey: value\n...\n---`) forms one contiguous paragraph that doesn't start with `#` and isn't a bare `---` line, so it survives the filter and becomes a garbage first item — e.g. `M0: --- provenance: interview last_updated: ...`.

**Fix applied locally:** strip a leading `^---\n[\s\S]*?\n---\n` block in `readTelosFile()` before any parsing.

## 2. `GenerateTelosSummary.ts` — `parseGoals()` silently drops all goals under time-horizon headings

`bucketOf()` only recognizes headings containing "active", "deferred"/"ongoing", or "completed"/"done". A `GOALS.md` organized by time horizon ("## Long-Term (5+ years)", "## Medium-Term", "## Short-Term") matches none of these, so `current` becomes `null` after the first heading and every subsequent bullet is silently dropped from the summary. This was previously masked because bug #1's frontmatter-garbage item kept the goals count non-zero, so the fail-loud zero-item guard never fired.

**Fix applied locally:** treat any heading containing "term" as `active` (in addition to the existing checks), since Long-/Medium-/Short-Term are all currently-pursued goals.

## 3. `Doctor.ts` — Linux browser detection never checks for Brave

`chromeBinary()`'s candidate list includes the macOS Brave path but only `google-chrome`/`chromium`/`chromium-browser` on Linux — no Linux Brave path at all. A Linux user running only Brave gets a false "no Chrome/Brave/Chromium binary found" even with Brave installed and working.

**Fix applied locally:** added `/usr/bin/brave-browser` and `/usr/bin/brave-browser-stable` to the candidates list.

## 4. `Doctor.ts` — ElevenLabs 402 errors misdiagnosed as generic quota exhaustion

[...]

PR #1567, opened 2026-07-23
fix(doctor): detect browsers installed via Nix, Home Manager, or PATH

vibecrypto:
## Problem

`Doctor.ts`'s `chromeBinary()` only probes legacy install locations:

- `/Applications/Google Chrome.app` / `/Applications/Brave Browser.app`
- `/usr/bin/google-chrome`, `/usr/bin/chromium`, `/usr/bin/chromium-browser`

On machines where the browser is installed declaratively — nix-darwin / Home Manager put GUI apps under `/Applications/Nix Apps/` and expose the binary on PATH (e.g. `/run/current-system/sw/bin/google-chrome`) — none of these candidates exist, so Doctor reports:

```
✗ Browser verification (Interceptor) — broken
  no Chrome/Brave/Chromium binary found
```

…even though Chrome is installed and fully working. The cached "broken" verdict (7-day TTL in `capabilities.json`) then steers sessions into declaring Interceptor unusable and suggesting a redundant Chrome install.

## Fix

Two additions to `chromeBinary()`, no behavior change on machines where the current paths already match:

1. Add the `/Applications/Nix Apps/` locations for Chrome and Brave to the candidate list.
2. Fall back to a PATH lookup via a small `whichPath()` helper (like the existing `which()`, but returning the resolved path — the probe's detail string does `browser.split('/')`, so the fallback must yield a string, not a boolean). Covers `google-chrome`, `chromium`, `brave`, and `brave-browser` — the last also fixes the Linux-Brave false negative reported in #1496 (item 3), which was closed without the fix landing: `/usr/bin` is on PATH, so the lookup finds `/usr/bin/brave-browser` without more hardcoded candidates.

## Testing

On a nix-darwin machine with Chrome installed via Home Manager (`/Applications/Nix Apps/Google Chrome.app`, binary on PATH):

- Before: `✗ Browser verification (Interceptor) — broken — no Chrome/Brave/Chromium binary found`
- After: `✅ Browser verification (Interceptor) — live — skill present, browser found (Google Chrome)`

Unchanged on a machine with a standard `/Applications/Google Chrome.app` install (first candidate still wins).

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/Doctor.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/Doctor.ts
@@ -119,13 +119,24 @@ function envKey(name: string): string | null {
   return null;
 }
 
+function whichPath(bin: string): string | null {
+  const hit = (process.env.PATH || '').split(':').find(p => p && existsSync(join(p, bin)));
+  return hit ? join(hit, bin) : null;
+}
+
 function chromeBinary(): string | null {
   const candidates = [
     '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome',
+    '/Applications/Nix Apps/Google Chrome.app/Contents/MacOS/Google Chrome',
     '/Applications/Brave Browser.app/Contents/MacOS/Brave Browser',
+    '/Applications/Nix Apps/Brave Browser.app/Contents/MacOS/Brave Browser',
     '/usr/bin/google-chrome', '/usr/bin/chromium', '/usr/bin/chromium-browser',
   ];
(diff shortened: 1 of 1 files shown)


danielmiessler:
Ported with credit. The `/usr/bin/brave-browser` candidates from #1496 were already in source; your Nix Apps paths and PATH fallback close the rest — and the `whichPath()` string-not-boolean detail was a nice catch. (Closing rather than merging: the public repo is generated from private source; the port ships in the next release.)

PR #1570, opened 2026-07-23
docs(skills): document context/background; opt Knowledge out of 2.1.218 background default

elhoim:
## Problem

Two related gaps:

1. `SkillSystem.md` documents the skill frontmatter contract but never mentions `context:` or `background:` — despite seven shipped skills declaring `context: fork`.
2. `skills/Knowledge/SKILL.md` declares `context: fork` with no `background` value, so it inherits a default that changed underneath it.

## Why it matters

**Claude Code 2.1.218** changed skills declaring `context: fork` to run as **background subagents by default**. Previously they blocked the main thread.

A backgrounded skill returns via task notification rather than inline output. For long-running deliberation skills that is an improvement — detaching them is the point. But `Knowledge` is an interactive lookup skill (`effort: low`, argument-hint `[search|...|retrieve|<query>]`). A user asking *"what do we know about X"* expects the answer in that turn. Backgrounded, it doesn't arrive there.

Because neither field was documented, there was no way to discover the opt-out short of reading the changelog.

## Fix

- Documents `context` and `background` in `SkillSystem.md` with a field table and the 2.1.218 gotcha, plus the test for which way to set it.
- Adds `background: false` to `Knowledge` only.

The other six fork skills — Council, Evals, Ideate, Research, RootCauseAnalysis, SystemsThinking — are all `effort: high` deliberation and are deliberately left on the new background default.

## Affected versions

Introduced in Claude Code **2.1.218**. `background` is a real parsed frontmatter field in that build, verified against the shipped binary.


danielmiessler:
The Knowledge `background: false` half had already landed in source independently — you called the right skill. The documentation half is now ported: SkillSystem.md documents `context` and `background` with a field table and the 2.1.218 default-change gotcha, including your test for which way to set it. Credited; ships in the next release.

PR #1579, opened 2026-07-23
lifeos: stop a --local launch from the main checkout when worktrees are in use

asdf8675309:
## What
'lifeos --local' starts a session wherever you are. If that's a repo's main checkout and the repo uses .claude/worktrees, the whole session runs on whatever branch the root is parked on — usually not what you meant. Easy footgun if you live in worktrees.

## Change
One file. New inMainCheckoutWithWorktrees() (exported). It's plain git: when --git-dir and --git-common-dir match, you're in the main checkout, not a worktree; then it just checks for a .claude/worktrees dir. No version-specific flags, so it works on any git. When it's true, --local refuses with a clear message. Set LIFEOS_ALLOW_ROOT=1 to override.

## Why it's safe
It only fires in one spot: --local, from a main checkout, in a repo that actually has .claude/worktrees, with the override off. Anything unexpected falls through to false and the launch proceeds — it never blocks you without an escape hatch, and it touches nothing outside --local. Not a git repo? Returns false, no-op.

## Verification
Tested on macOS, Bun 1.3.14.

| context | inMainCheckoutWithWorktrees() | --local | |---------|-------------------------------|---------| | not a git repo | false | proceeds |
| main checkout, no .claude/worktrees | false | proceeds | | linked worktree (git-dir != git-common-dir) | false | proceeds | | main checkout WITH .claude/worktrees | true | refuses unless LIFEOS_ALLOW_ROOT=1 |

## Scope
Just lifeos.ts, just the --local path. No new dependency. Nothing changes for non-worktree repos or for normal worktree launches.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/lifeos.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/lifeos.ts
@@ -394,6 +394,26 @@ function cmdWallpaper(args: string[]) {
 }
 
 
+// True when the current directory is a git repo's MAIN checkout (not a linked
+// worktree) AND the repo uses the .claude/worktrees convention. The main checkout
+// has --git-dir == --git-common-dir; a linked worktree's --git-dir points into
+// .git/worktrees/<name>, so they differ. Used to stop a --local launch from running
+// a whole session on whatever branch the root happens to be parked on.
+export function inMainCheckoutWithWorktrees(): boolean {
+  try {
+    // No --path-format flag → works on any git version; the two paths stay directly
+    // comparable (both ".git" in a main checkout, divergent in a linked worktree).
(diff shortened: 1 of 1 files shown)


danielmiessler:
Ported with credit — and porting it caught a real edge worth handing back: the harness creates an EMPTY `.claude/worktrees` dir as a side effect of one-off agent isolation (our own install has one), and the as-submitted check would have locked `--local` out of the main checkout there. The ported version requires the dir to be non-empty — "worktrees in use" literally. Verified across five contexts: non-git, main-without-worktrees, main-with-empty-dir, and inside-a-linked-worktree all proceed; main-with-populated-worktrees refuses. (Closing rather than merging — the public repo is generated from private source; the port ships next release.)

PR #1580, opened 2026-07-23
Add CoreUpgrade skill: scoped, reversible major-version upgrade for existing installs

asdf8675309:
## The problem

The installer is deliberately additive — `DeployCore` copies only *missing* files and never overwrites a populated target. That's correct for a fresh install, but it means there is **no supported path to move an existing install across a major version**. Deploying a new payload over an old install leaves every old SYSTEM file in place, producing a broken hybrid where new and retired machinery coexist. The only alternatives today are a full reinstall (loses local state) or hand-deleting directories (one wrong `rm -rf` and USER data is gone). There's a missing primitive: a **scoped, reversible clear-and-replace** with the preserve boundary encoded as tested code rather than operator care.

This is the migration gap raised in discussion #1417: installing a newer version over a customized install force-overwrites existing hook scripts with no backup — the installer backs up `settings.json` but not the hooks it replaces.

## The change

A new `CoreUpgrade` skill (`skills/CoreUpgrade/`) with one tool, `LifeosUpgrade.ts`:

1. **Plan** — the clear-list is enumerated from the *live* install; an entry is cleared only if it passes the preserve check **and** the new payload ships a replacement for it. A wrong or malicious payload can neither inject a path nor clear a preserved zone; an empty payload clears nothing. Upstream deletions are honored; local content the payload doesn't know about survives.
2. **Backup** — exactly the to-be-cleared entries are copied to a backup dir *outside* the install (enforced), verified per-entry, **before any delete**.
3. **Clear + deploy** — every delete goes through a guarded remove (preserve-denylist + path-escape guard, case-insensitive for APFS, failing toward preservation). The payload then fills the cleared slots.
4. **Rollback** — on any error, deploy-created artifacts are removed and cleared entries are restored from the backup, with partial failures named explicitly (with their backup path) rather than reported as success.

Preserved untouched: `LIFEOS/USER/`, `LIFEOS/MEMORY/`, `.env*`, `settings*.json`, private `skills/_*`, and any command/agent/skill the payload does not ship.

**Optional `--split-claude-md`** ends CLAUDE.md drift: it diffs your `CLAUDE.md` against the payload's `CLAUDE.template.md` (deterministic markdown block-diff, no model calls), auto-extracts pure *additions* into an imported `LIFEOS/USER/CUSTOMIZATIONS/GLOBAL.md`, and **flags** modified base lines for manual placement rather than guessing which version you meant. So the base CLAUDE.md can be pulled fresh on every future upgrade.

Dry-run is the default; nothing mutates without the literal `--apply`.

## Why it's safe

- **Dry-run by default** — `--apply` is the only mutating path, and no delete happens unless its backup is verified on disk first. Every cleared entry — the hook scripts #1417 flagged included — is copied to an external backup and verified before any delete.
- **Payload-aware clearing** — only entries the payload ships are ever cleared; the preserve denylist + path-escape guard gate every single delete and fail toward preservation.
- **Refuses to run on the wrong thing** — a LifeOS *source checkout* (dev tree) and an ambiguous target (a bare dir with a stray `CLAUDE.md`) are both rejected.
- **Named-failure rollback** — anything that couldn't be auto-restored is listed with its backup path, never silently dropped.
- **Symlink-correct** — payload symlinks are never deployed (matching the installer), and backup/rollback verification is `lstat`-based so symlink-bearing zones round-trip.

## Verification

**Validated on a real upgrade, not only in tests.** This tool performed the actual `v6.0.5 → v7.1.1` upgrade of a heavily-customized live install (`--apply --split-claude-md`): it created a scoped backup of only the cleared SYSTEM entries, ran the CLAUDE.md split (additions extracted to `GLOBAL.md`), preserved USER / MEMORY / private `_*` skills / `.env` / settings intact, and the install came up healthy on v7.1.1.

- **`bun test`** — 27 tests / 138 assertions, green (macOS 25.5, Bun 1.3.14). Covers: preserve-zone protection (incl. case variants), payload-aware clearing, empty/wrong-payload no-op, dev-tree + ambiguous-target + missing-payload refusal, symlink safety, backup scoping + pre-existing-backup-dir refusal, rollback (restore + new-artifact removal + named failures with causes), the CLAUDE.md split (additions vs flagged modifications vs noise, no-clobber with a write-time re-check, failure atomicity), full split+rollback composition under a simulated deploy failure, and two subprocess integration tests of `main()` itself.
- **Clone rehearsal** — the whole cut was rehearsed on a throwaway clone (`--config-root <clone>`) before the live apply.
- **Transpiles clean** (`bun build --no-bundle --target=bun`, exit 0).

**Honest scope of the live proof:** the live apply *succeeded*, so the rollback path was exercised by the unit/e2e tests and the clone rehearsal, **not** fired in the production run. And the tool does the scoped core replacement + optional CLAUDE.md split only — post-upgrade steps (deps install/build, `InstallHooks`, settings reconcile, re-applying any custom edits made inside replaced SYSTEM dirs) are **manual and documented** in the skill's "Post-Upgrade Steps," printed by the tool at the end of a run.

## Scope

Additive — one new skill directory (`SKILL.md` + `Tools/LifeosUpgrade.ts` + `Tools/LifeosUpgrade.test.ts`) and one line added to `CLAUDE.template.md` (the commented `@…/CUSTOMIZATIONS/GLOBAL.md` import the split activates). No existing behavior changes; the installer and every current flow are untouched. `disable-model-invocation: true` — it only runs when a human calls it.


danielmiessler:
Not closing this one either way yet — it's under active consideration. The engineering is excellent (the test matrix, verified backup-before-delete, and the honest-scope section especially), and the migration gap it targets is the loudest recurring theme in these discussions. The open question is doctrine, not code: the supported cross-major path has deliberately been fresh-install + port USER, and adopting this means owning an in-place upgrade primitive and its support surface long-term. I want to decide that deliberately rather than in a triage pass — you'll hear back here either way.

asdf8675309:
Sounds good - I worked on it so that I could use it to go from v6 to v7.  After going through a pretty manual upgrade path to go from v5 to v6 -- with 77 customizations.  So shifting some of the customizations into a portion that doesn't change alongside working on what could potentially go upstream was the approach this time around.  Perfect - nah - but if nothing else I was hoping it would give a start to some thoughts about a potential upgrade script.

asdf8675309:
Understood on the doctrine call, no rush from my side.

Pushed one more commit because I hit a real failure with this: upgrading my own install cleared `skills/Interceptor` and took its `Extension/` with it. That directory is a local pin built by the skill's own `Pin.sh`, so the payload ships the skill but can't ship the pin, and nothing puts it back. It stayed invisible for two days because Chrome keeps an already-loaded unpacked extension running from memory until the next restart. The commit adds a post-deploy re-pin — `planRepin` (pure, 5 tests) plus `runRepin` — that only fires when that skill was actually cleared and the operator's local build exists, and sits outside the rollback-guarded block.

Two flags rather than quiet decisions: it executes a payload-shipped script, which I documented in the header as a deliberate exception to the payload-distrust stance; and a general upgrade tool naming one specific skill is a smell. Generalizing to "any `Pin.sh` in a cleared skill" widens that execution surface, so I left it narrow — happy to go either way, or drop the commit if it muddies the doctrine question.

Also sent #1602 for the other half of this: Interceptor's preflight only detects a *stale* pin, never an absent one, so it reported OK the entire time the extension was gone. That one stands alone and doesn't depend on this PR.

greg-bandy:
Field data from an in-place 6.0.5 → 7.1.1 upgrade this week, plus one detail I don't think is captured yet: the *documentation* now asserts the overwrite that `DeployCore` can't perform, so following the official Update workflow reads as success while silently half-upgrading.

**The doc/tool contradiction**

`Workflows/Update.md` step 3 (as shipped in 7.1.1) says:

> **Re-overlay system** — re-copy the system templates (CLAUDE, system prompt, `settings.system.json` minus hooks). These are system-owned and safe to overwrite.

No shipped tool does this. Every copy in `DeployCore.ts` routes through `InstallEngine.copyMissing`, which is documented as *"Recursive, existsSync-GUARDED copy. Copies only files/dirs absent at dst — NEVER overwrites a populated target."*

#1519 was a docs-only change that rewrote step 3 this way; the tooling side was left as an explicit follow-up ("copyMissing delivers new files but leaves changed system-owned files stale after an update — worth its own design pass"). The result is that an operator or agent following Update.md believes step 3 overwrote the templates, and nothing surfaces that it didn't.

**Measured impact on a real install**

Comparing the 7.1.1 payload against a live 6.0.5 tree (excluding `USER/` and `MEMORY/`):

| | files |
|---|---|
| New — would be added | 82 |
| Identical — no-op | 228 |
| **Changed — silently skipped, stay at 6.0.5** | **222** |

The 222 include `VERSION`, `LIFEOS_SYSTEM_PROMPT.md`, and effectively all of `TOOLS/`.

**Why it's self-concealing**

`VERSION` is in the skipped set, so after running Update the marker still reads `6.0.5`. With #1519's step 2 now doing a real release check, that guarantees the *same* diff re-trips on every subsequent run — each one reporting work done, each one changing nothing. There's no state in which the install converges, and no signal that it isn't converging. #1519 anticipated this ("step 3 additionally overwrites the `VERSION` marker") but that overwrite has no implementing tool either.

**Two smaller things**

- `install.sh` carries `LIFEOS_FALLBACK_TAG="v7.11.0"`, but no such tag exists — latest is `v7.1.1`, and `v7.11.0` sorts *above* it under semver. If the redirect probe and the API both fail, the fallback resolves to a nonexistent release. Given the stamp comment's history (the 2026-07-12 silent v7.0.0 install), the fallback may want an existence check.
- The payload ships `PULSE.toml` with `[bunker] enabled = true`, pointing at `~/Projects/bunker`. On installs without that repo the module errors on every refresh. Defaulting a module to enabled when its target is a path outside the install may be worth reconsidering — though as a config file, `copyMissing` correctly leaves it alone on an additive deploy, so this only bites paths that overwrite it.

**What I ended up doing**

Given the doctrine noted above (fresh-install + port USER), I took the unsupported route deliberately: `rsync` of the payload runtime over the live tree, excluding `USER`/`MEMORY`/`node_modules`/`.git` — the same set as `DeployCore`'s `RUNTIME_SKIP` — with a git snapshot as rollback, then `InstallHooks.ts`, `ScaffoldUser.ts`, `ActivateImports.ts` as normal. Of the 222 overwritten files exactly one was a config file (`PULSE.toml`); the rest were code or generated docs. That's a narrow enough blast radius that a supported primitive along #1580's lines looks very achievable — the preserve boundary in practice was `USER/`, `MEMORY/`, and the handful of `.toml` files.

Not arguing for any particular resolution on the doctrine question — just noting that until it's settled, Update.md step 3 promises something no tool delivers, and the stale `VERSION` marker keeps that invisible.

danielmiessler:
Status after today's triage: still very much wanted—the gap it fills (scoped, reversible major-version upgrade) is real and today's batch fixed several of its neighbors (overlay completeness, substitution on update, a minimal retirement pass). A delete-capable engine earns a dedicated adversarial review session before porting; the dry-run-default + verified-backup-before-delete design is exactly why it's still on the table. Sorry for the wait—it's earned a proper look, and that's what it'll get.

Drizzt321:
At a minimum I'd still very much like this, even if just purely as a Skill, since ANY sort of customizations results in an annoying upgrade between tagged version.

PR #1581, opened 2026-07-23
fix(installer): bun install nested runtime packages so Pulse can start

elhoim:
## Problem

On a fresh install, **Pulse dies on startup**:
```
error: Cannot find package 'smol-toml' from '.../LIFEOS/PULSE/pulse.ts'
```
`smol-toml` *is* correctly declared in `LIFEOS/PULSE/package.json`. The bug is not a missing manifest entry — it is that the installer never runs `bun install` **inside the nested package**.

Several directories in the deployed runtime ship their own `package.json` (`LIFEOS/PULSE/`, `LIFEOS/PULSE/Observability/`, `LIFEOS/TOOLS/`, …). Node/Bun module resolution stops walking up at the first ancestor that owns a `package.json`, so the shared `node_modules` the installer creates at the config root never satisfies these nested packages. Each needs its own `bun install`.

The Observability dashboard has a second layer of the same problem: it ships as a Next.js static export (`Observability/out/index.html`) that must be **built**. Until it is, `pulse.ts` reports `status: degraded` and the dashboard route returns HTTP 503.

## Reproduction

1. Fresh install.
2. `bun run pulse.ts` → dies on `Cannot find package 'smol-toml'`.
3. After manually `bun install`-ing `LIFEOS/PULSE/`, Pulse starts but `GET /` 503s until `Observability/` is installed and built.

## Fix

Adds a `nested-dependencies` deploy step to `DeployCore.ts` that, after the existing dependency step:
- walks the deployed runtime tree for every `package.json` (skipping `node_modules`/`.git`) and runs `bun install --cwd <dir>` on each;
- additionally runs `bun run build` for the Observability dashboard when its static export is missing.

Respects the existing `--apply` / dry-run contract (dry-run lists the exact `bun install` / `bun run build` commands it would run) and records per-package failures without aborting the whole deploy.

## Files
- `LifeOS/Tools/DeployCore.ts`
- `LifeOS/install/skills/LifeOS/Tools/DeployCore.ts` (install-payload mirror)

### DIFF
--- a/LifeOS/Tools/DeployCore.ts
+++ b/LifeOS/Tools/DeployCore.ts
@@ -43,7 +43,7 @@ function arg(a: string[], flag: string): string | undefined {
 }
 
 interface DeployResult {
-  what: "skills" | "runtime" | "memory" | "dependencies";
+  what: "skills" | "runtime" | "memory" | "dependencies" | "nested-dependencies";
   src: string;
   dst: string;
   present: boolean;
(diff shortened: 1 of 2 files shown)


danielmiessler:
Ported with credit. Confirmed the gap in source — the installer ran exactly one root-level `bun install` and nothing for the nested packages — and your fix is in: the walk, the per-package `--cwd` installs, and the Observability build-when-missing. Verified the dry-run lists the right actions for every nested `package.json`, including the build line when `out/index.html` is absent. This closes a genuine first-boot Pulse killer, which is squarely against our own "full Pulse on a fresh install" release bar. (Closing rather than merging — generated-repo doctrine; ships next release.)

PR #1587, opened 2026-07-24
telos: make TelosFreshness and InterviewScan split-file aware

asdf8675309:
## What
LifeOS supports a split-file TELOS: TELOS.md is a thin index, and each section lives in its own ALLCAPS file (MISSION.md, GOALS.md, ...). GenerateTelosSummary already reads those, but TelosFreshness and InterviewScan only ever looked at TELOS.md's H2 headings. So on a split-file setup, freshness misses every section and the interview report scores them as "does not exist."

## Change (two files, one concern)
- TelosFreshness.ts: adds SPLIT_FILE_BY_SLUG (the inverse of GenerateTelosSummary's LEGACY_FILE_TO_SECTION) and an exported splitFilePathForSlug(). readTelosFreshness now surfaces each split file as its own section, and the freshness bump writes the split file. One shared extractPreview() helper feeds both the H2 scan and the split scan.
- InterviewScan.ts: extractSectionBody reads the split file's body as the real source. It imports from TelosFreshness, which is why this is one PR.

When a section shows up as both a TELOS.md H2 and a split file, the split file wins — same as GenerateTelosSummary, and same as what TELOS.md itself says. In a normal thin-index setup nothing lives in both, so this only settles the odd overlap case.

## Why it's safe
For a plain unified TELOS.md there's no split file, so none of this runs and nothing changes. The split path only kicks in when the sibling file exists. A populated split file wins; an empty one falls back to the H2, so a half-made file can't hide real content. Nothing gets counted twice. The only write is the existing freshness bump, now pointed at the right file.

## Verification
Tested on macOS, Bun 1.3.14, against a real split-file TELOS (thin index plus 21 ALLCAPS files).

| TELOS shape | TelosFreshness sections | InterviewScan on Mission | |-------------|-------------------------|--------------------------| | unified (H2 per section) | all H2 sections (unchanged) | reads H2 body (unchanged) | | split, BEFORE this change | 3 (only the index H2s) | 0% "section does not exist" | | split, AFTER this change | 24 (3 H2 + 21 split, real per-file freshness) | 100% (reads MISSION.md body) | | split, sibling genuinely absent | not surfaced | 0% "section does not exist" (correct) |

## Known follow-up
SPLIT_FILE_BY_SLUG and LEGACY_FILE_TO_SECTION are the same mapping in two directions, kept by hand in two different tools — add a TELOS concept and you have to update both. A single shared map would kill the drift risk. Left it out here to keep this to one concern.

## Scope
TelosFreshness.ts + InterviewScan.ts. No new dependency. Unified-TELOS users see no change.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/InterviewScan.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/InterviewScan.ts
@@ -25,7 +25,7 @@ for (const __k of ["LIFEOS_DIR", "LIFEOS_CONFIG_DIR", "PROJECTS_DIR"]) {
 
 import { readFileSync, existsSync } from "fs";
 import { join } from "path";
-import { readTelosFreshness, sectionSlug, type SectionFreshness } from "./TelosFreshness";
+import { readTelosFreshness, sectionSlug, splitFilePathForSlug, parseFrontmatter, type SectionFreshness } from "./TelosFreshness";
 
 // Normalize env path vars that Claude Code injects without shell expansion (LifeOS#1404)
 for (const k of ["LIFEOS_DIR", "LIFEOS_CONFIG_DIR", "PROJECTS_DIR"]) {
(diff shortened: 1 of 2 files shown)


danielmiessler:

PR #1589, opened 2026-07-24
docs: define Hindsight memory port schema

AronAxe:


danielmiessler:
Appreciate the work that went into this, but it can't merge: it rewrites core docs and five install tools to describe a different product also named Hermes, with its own external memory bank — which collides directly with LifeOS's Hermes sidecar doctrine (one brain, the sidecar is another way into it, no second memory). It also carries personal paths and a compiled `.pyc` artifact. The port-schema idea itself is better kept in your fork as a standalone doc; closing here.

PR #1591, opened 2026-07-24
fix(hooks): correct DocIntegrity trigger naming to SessionEnd (docs still said Stop)

anikinsasha:
DocIntegrity.hook.ts runs on **SessionEnd** — the registration (`hooks/hooks.json` "SessionEnd" block) and `HookSystem.md` (which records the move off Stop during the consolidation) are already correct. But 13 lines across 10 files still describe the mechanism as Stop-triggered:

- Handler docstrings: `DocCrossRefIntegrity.ts:11`, `RebuildArchSummary.ts:11`, `MemoryDirIntegrity.ts:14` (all three siblings called by DocIntegrity.hook.ts)
- `IntegrityCheck.hook.ts:7` — its cross-reference to DocIntegrity says "(Stop event)"
- `ArchitectureSummaryGenerator.ts` — the hard-coded PIPELINE_ROUTER string "Stop-hook cross-reference checks" (this is the actual source of the ARCHITECTURE_SUMMARY.md row; editing the .md alone gets overwritten on regen), plus the generated row in `ARCHITECTURE_SUMMARY.md` aligned in the same commit
- `LifeosSystemArchitecture.md` — "(Stop)" in the Pipeline Topology intro and the Doc Integrity pipeline row
- `hooks/README.md` — DocIntegrity relocated from the Stop fire-order diagram, event table, Stop-hooks table, and Stop→UserPromptSubmit bridge into the SessionEnd section (it was absent there), placed per the hooks.json fire order
- `settings.enhancements.json` spinner tips (both copies: install root and skills/LifeOS/install)

Docs and comments only — no behavior change.


danielmiessler:

PR #1594, opened 2026-07-24
refactor(memory)!: rename <pai-memory*> context tags to <lifeos-memory*>

anikinsasha:
## Breaking change — PAI→LifeOS naming leftover

The memory context-block tags injected into prompts still carry the pre-rename PAI naming. This renames them to match LifeOS:

- `<pai-memory>` → `<lifeos-memory>`
- `<pai-memory-delta>` → `<lifeos-memory-delta>`
- `<pai-memory-health>` → `<lifeos-memory-health>`

All coupled surfaces change in this one commit — the emitters (LoadMemory, MemoryDeltaSurface), the MemoryTurnStart dispatcher docs, FormatGate's detection, and the system prompt's 🧠 MEMORY-line contract. These string-match each other and must never be renamed separately; a COUPLING comment now marks the FormatGate site. FormatGate keeps accepting the legacy `pai-` tag so sessions started before the rename still register their deltas.

**Anyone with downstream prompt customizations referencing the old tag names must update them** — hence the breaking-change label. The `pai-freshness-v1` frontmatter convention is a data-format value stamped across user files, not a tag, and is deliberately NOT renamed here.

Stacked on #1593 (`fix/memory-marker-canonical-rebuild`) — the diff shown includes that PR until it merges; this PR's own change is the tag rename commit only.

### DIFF
--- a/LifeOS/install/LIFEOS/LIFEOS_SYSTEM_PROMPT.md
+++ b/LifeOS/install/LIFEOS/LIFEOS_SYSTEM_PROMPT.md
@@ -47,7 +47,7 @@ You ARE the DA. Speak as yourself — "I", "me", "my system", "our work." Never 
 
 [Short bullets: the evidence — whenever CHANGE appears]
 
-🧠 MEMORY: [verbatim hook-fed line when a <pai-memory-delta> block is present; omit otherwise]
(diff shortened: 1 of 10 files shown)


danielmiessler:

PR #1595, opened 2026-07-24
docs(hooks): sync README event inventory to hooks.json registration truth

anikinsasha:
## What

`hooks/README.md` still documents the pre-consolidation hook topology (its header says *Last updated 2026-05-06*). The 2026-07-10/11 consolidation reshaped the hook tree — TabState/EventLogger/StopGates/PreToolGuard/MemoryTurnStart dispatchers, TheRouter retired, several hooks removed — but the README's event inventory was never resynced to `hooks/hooks.json`.

This PR rewrites the README's inventory to match what `hooks.json` actually registers, section by section (fire-order diagram, event table, per-event registry tables, inter-hook bridges, data-flow diagram, troubleshooting, migration notes).

## The drift, quantified

- **15 hook names in the README exist neither on disk nor in any registration**: TheRouter, ArtWorkflowGuard, SetQuestionTab, QuestionAnswered, TelosSummarySync, ToolActivityTracker, ToolFailureTracker, ResponseTabReset, StopFailureHandler, PreCompact, RestoreContext, ConfigAudit, InstructionsLoadedHandler, ULWorkSync (private hook, not in the public payload per WorkSystem.md), RelationshipMemory (deleted 7.0.0 per HookSystem.md).
- **10 registered, firing hooks were missing from the README**: HookHealer, MemoryTurnStart, PreToolGuard, TabState, EventLogger, AlgorithmNudge, ISARenderOnStop, StopGates, MemoryReviewFire, MemoryHealthGate.
- **3 documented event types have no registration and no files**: PreCompact, PostCompact, InstructionsLoaded. (PermissionRequest, which IS registered, was missing from the event table.)

## Changes

- Fire-order diagram, event table, and all per-event registry tables now mirror `hooks.json` array order exactly.
- New **Consolidation Dispatchers** section: the registered-dispatcher vs standalone-sub-hook pattern (StopGates → Format/Verification/WritingGate; PreToolGuard → SystemFileGuard/CommunicationSkillGuard/EgressClassGuard; MemoryTurnStart → LoadMemory/MemoryDeltaSurface; plus the single-file event dispatchers TabState/EventLogger/Safety), with dispatch order and fail policies from the hook headers.
- Prose sections updated: TheRouter's classification flow replaced with the current prompt-start flow (HookSystem.md records TheRouter as retired entirely, no successor classifier); Stop bridge now lists all six registered Stop hooks; maintenance checklists point at `hooks.json` + `settings.json`.
[...]

PR #1597, opened 2026-07-24
fix: remove stale TheRouter references from live code comments, prompt text, and skill docs

anikinsasha:
## Problem

`TheRouter.hook.ts` was retired entirely in release 7.0.0 (`3baba670`, the 2026-07-11 hooks-BPE pass): per-prompt mode/tier classification abolished, no successor classifier, its router libs deleted with it (`HookSystem.md` documents the retirement). But a set of live references survived, still describing it as running machinery — including **prompt text that ships to the model on every Telegram/iMessage turn**:

> "TheRouter has injected a TELEGRAM_DIRECTIVE in this turn's additionalContext."

That sentence is false twice: nothing injects anything (the directive IS the hardcoded `systemPrompt.append` literal in the bridge module itself, passed to the Agent SDK `query()`), and `additionalContext` was the retired hook's mechanism (`hookSpecificOutput` on UserPromptSubmit).

## Changes (wording-only — every hunk is a comment, docstring, prompt string, or doc prose)

| File | Fix |
|---|---|
| `PULSE/modules/telegram.ts`, `imessage.ts` | Directive attribution sentence reworded to the truth; the operative DO-NOT-emit list is byte-identical |
| `PULSE/lib/strip-mode-scaffolding.ts` | Layer-1 (prevention) attribution corrected: the channel bridges append the directive, not a hook |
| `hooks/lib/isa-utils.ts` | `markSessionNative` / `markAlgorithmStarting` docblocks marked **no live caller** (their only caller was the retired classifier); the `markAlgorithmStarting` docblock was physically attached to the wrong function — moved to its definition; `upsertSession`'s one-way guard comment now states no downgrade path remains |
| `hooks/lib/tab-setter.ts` | `setModeToken` docblock marked no-live-caller (nothing stamps fresh `N`/`E{n}` tokens) |
| `hooks/PromptProcessing.hook.ts` | Module NOTE + mode-token comments corrected: `isNativeMode()` is now the only live mode signal (feeds `work.json` `currentMode`); tab-token recovery preserves legacy state only |
| `LIFEOS/TOOLS/Inference.ts` | Four present-tense classifier claims ("it now runs at high", "it fires on every prompt", "only the TheRouter classifier does") moved to past tense, resolving an internal contradiction |
| `skills/ISA/Workflows/Scaffold.md` | Removed the `GOAL_SIGNAL` classifier handshake and `INTERVIEW_ELIGIBLE` additionalContext clause — tier inference is the only eligibility path |
| `skills/CMUX/DESIGN.md` | Token-authority sentences corrected; section flagged as predating the retirement |
| `skills/Hardening/Workflows/PropertyTest.md` | Oracle example swapped to a live component |
| `modes/native.md`, `LifeosSystemArchitecture.md`, `CoreComponents.md`, `HookSystem.md` | Qualified as retired/history (HookSystem also gains the public squash hash alongside the unpublished pre-squash `4dd0fbe19`) |

## Evidence

- Caller analysis: `markSessionNative`, `markAlgorithmStarting`, `setModeToken` have **zero call sites** anywhere in the payload — their docstrings described a calling convention that no longer exists. (Removing the dead exports themselves is deliberately NOT in this PR — that's a behavior change; happy to file it as a follow-up issue.)
- A sweep checker (walks the payload, asserts every remaining `TheRouter` mention is an explicit historical note or has a retirement marker within ±3 lines) passes on this branch except for 12 hits in `hooks/README.md` — deliberately untouched here because **PR #1595** owns that file's rewrite. After both merge, every remaining reference is an explicit historical/migration note.

No logic changes; no template variables touched (`{{PRINCIPAL_NAME}}`/`{{DA_NAME}}` context lines preserved).


danielmiessler:

PR #1601, opened 2026-07-24
fix(pulse): memory-consolidation must be a script job, not type:claude

asdf8675309:
The memory-consolidation job runs two deterministic CLIs (SessionHarvester, LearningPatternSynthesis) but was typed as "claude". spawnClaude() launches the CLI with --tools "" (PULSE/lib.ts), so an agent told to *execute* those commands has no tool to do it — it emits a few seconds of narration and exits, which the runner logs as "completed". The job has therefore never run.

Convert it to type:"script"; spawnScript() runs the command via bash -c, so the && chain executes both tools directly. Drops the (never-reached) LLM "summarize" step and the now-unused model key.


danielmiessler:

Issue #1550, opened 2026-07-20
TestingDoctrine.md specifies `test/harness.ts` as the probe behind every ISC, but no harness and no tests ship in the release


tzioup:
---

`LIFEOS/DOCUMENTATION/Testing/TestingDoctrine.md` (v1.0.24) defines `bun test` as "the canonical probe behind every ISC claim" and describes a shared harness at `~/.claude/test/harness.ts` in the present tense, including its size and export count. Neither the harness nor any test file is present in the public release.

Reproduced against a clean clone of `main` at `4e33cc8`.

## Reproduce

```bash
git clone --depth 1 https://github.com/danielmiessler/LifeOS /tmp/lifeos-check
cd /tmp/lifeos-check

find . -name 'harness.ts' -not -path '*/.git/*' | wc -l   # → 0
find . -name '*.test.ts' -not -path '*/.git/*' | wc -l    # → 0
find . -type d -name test  -not -path '*/.git/*' | wc -l  # → 0

# the doctrine that references them
grep -n 'harness.ts' LifeOS/install/LIFEOS/DOCUMENTATION/Testing/TestingDoctrine.md
```

## What the doctrine states

- Line 14: *"The shared harness lives at `~/.claude/test/harness.ts` and exports zero-external-dep helpers (`paiTestEnv`, `tempDir`, `claudeFixture`, platform predicates, custom matchers). Tests live in a parallel `~/.claude/test/` tree… Coverage is corpus-based: every documented hook, skill workflow, and tool surface gets at least one test file."*
- Line 32 describes it concretely: *"`test/harness.ts` (1985 lines, 98 exports, header comment forbids external deps)"*.
- Line 24: *"What's been missing is the invocable probe behind each ISC. This doctrine closes that gap by making `bun test path/to/foo.test.ts` the canonical answer."*

The document contains no hedging language — `grep -cE "aspiration|planned|not yet|TODO|future|will be"` returns 0 — so it reads as describing shipped state, not a roadmap.

## The consequence

`IsaFormat.md` makes `## Test Strategy` name a `tool:` per ISC, and the doctrine makes `bun test path/to/foo.test.ts` the canonical value for that field. On a fresh public install, any ISC written to that contract names a probe that cannot run: there is no harness to import and no test tree to put the file in. The ISC still gets checked off — just not by an executable probe.

[...]

PR #1603, opened 2026-07-24
docs(testing): scope the test-location rule — subsystem packages keep their suite in-package

anikinsasha:
TestingDoctrine rule 2 maps every test into the repo-root `test/` tree, and the anti-patterns list bans co-location without qualification. That is right for the repo-root surface (hooks, skills, single-file tools) but leaves one shipped class with no home in the rule: self-contained subsystem packages — tool directories with their own `package.json`, in the current payload `LIFEOS/TOOLS/TokenXray/` and `LIFEOS/TOOLS/llcli/`. A package with its own dependencies and lifecycle is the unit that travels; routing its suite into the repo-root tree strands the tests whenever the package is developed, vendored, or deployed on its own.

Changes (docs only, no behavior change):

- Rule 2 gains a **Subsystem-package clause**: a directory-tool with its own `package.json` keeps its suite in-package at `<tool>/test/` with a `package.json` `test` script — the suite travels with the deployable package. The parallel repo-root `test/` tree continues to govern the repo-root surface.
- The co-location anti-pattern row is scoped to match (repo-root surface only; subsystem packages keep in-package suites per Rule 2). A package-local `test/` directory is the parallel-tree pattern at package scope, not co-location — co-location means a test file next to its source file.

Deliberately not touched: no test tree or harness is added (that packaging question is #1550's, whose resolution this clause is consistent with — the clause governs where tests live when they exist); no bunfig or code changes; rules 1 and 3–11 unchanged. The corpus coverage rule (every hook/skill/tool surface gets at least one test file) loses nothing — a repo-root `bun test` walk still reaches in-package suites.

We run this split on our install: two subsystem tools keep their suites in-package (a 14-file suite behind a bare `bun test` script; a second package whose script scopes unit tests apart from live-integration runners), run per-package, with the packages' deploy scripts shipping `test/` alongside the code — tests traveling with the package is exactly what the clause writes down.

### DIFF
--- a/LifeOS/install/LIFEOS/DOCUMENTATION/Testing/TestingDoctrine.md
+++ b/LifeOS/install/LIFEOS/DOCUMENTATION/Testing/TestingDoctrine.md
@@ -78,6 +78,8 @@ Mirror the source path, not co-locate. Map:
 
 Snapshot files at `test/<surface>/__snapshots__/<name>.test.ts.snap`.
 
+**Subsystem-package clause.** A directory-tool with its own `package.json` (`LIFEOS/TOOLS/<tool>/`) keeps its suite in-package at `<tool>/test/` with a `package.json` `test` script — the suite travels with the deployable package. The parallel `~/.claude/test/` tree governs the repo-root surface (hooks, skills, single-file tools).
+
 ### 3. Every test file is independently runnable.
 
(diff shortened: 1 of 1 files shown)


danielmiessler:
Thanks @anikinsasha — the test-location scope clause is already in TestingDoctrine.md in source, credited. Closing here since the public repo is generated from that source; it'll be in the next release.

PR #1606, opened 2026-07-24
Add Grill discovery-interview workflow to the ISA skill

asdf8675309:
## The gap

ISA already covers known-structure work — Scaffold, Interview, CheckCompleteness, Reconcile, Seed, Append. What's missing is the front end for when you don't know the shape yet. If an idea's still half-formed, you can't scaffold it honestly — you just bake in assumptions nobody checked. That's the most common way agentic work goes off the rails: the agent runs with the wrong mental model before you get a chance to correct it.

## The change

A Grill workflow that interrogates a half-formed idea until its shape is clear, then hands it to Scaffold. Three pieces:

- **`skills/ISA/Workflows/Grill.md`** — walks the design tree one question at a time, recommends an answer each turn, explores the codebase instead of asking when it can, and checkpoints every turn to `MEMORY/WORK/{slug}/grill.md` so a long session doesn't lose earlier answers to context drift. Runs a pre-mortem→draft-ISC pass, then hands off to Scaffold.
- **`skills/ISA/SKILL.md`** — one new routing row (`"grill me" / "discovery interview" / "figure out the shape"` → `Workflows/Grill.md`) and a Grill clause in the description.
- **`commands/grill-me.md`** — a thin `/grill-me <topic>` command that routes into `Skill("ISA", "grill me …")`. Model-invocation is off, so it only fires when you ask.

Same split as the skill it's based on: a reusable interview loop plus a thin command to trigger it.

## Why it's safe

- Additive. One new workflow, one new command, one routing row, one description tweak. Nothing existing changes.
- Fails safe. If Grill mis-routes it falls back to normal ISA behavior, and it never writes outside `MEMORY/WORK/{slug}/grill.md` — scratch output that sits next to where the ISA lands.
- Human-gated. The command is `disable-model-invocation`; Grill runs only when you call it.

## Verification

- Running live in a real install: `commands/grill-me.md` → `Skill("ISA", "grill me …")` → the SKILL.md Grill row → `Workflows/Grill.md`, all resolve. The ISA skill validates — every routing row points at a real file, and the description is under the frontmatter cap.
- The Grill → Scaffold handoff works end to end: the `grill.md` checkpoint feeds `Skill("ISA","scaffold from <slug>/grill.md")`.

## Scope

Additive. `Workflows/Grill.md` and `commands/grill-me.md` are new, plus two small edits to `ISA/SKILL.md`. No other skill, tool, or hook touched.

## Credit

This adapts Matt Pocock's **`grilling` / `grill-me` skills** ([github.com/mattpocock/skills](https://github.com/mattpocock/skills) — MIT, © 2026 Matt Pocock). His "grill me" prompt is the core idea: *interview me relentlessly about every aspect of this plan; walk each branch of the design tree, resolving dependencies one at a time; recommend an answer per question; ask one at a time; explore the codebase instead of asking.* This is an independent implementation wired into LifeOS's ISA discovery→scaffold pipeline — it checkpoints each turn to `MEMORY/WORK/{slug}/grill.md` instead of a project-root `brainstorms/` folder, and adds a pre-mortem→draft-ISC pass before the Scaffold handoff.

### DIFF
--- a/LifeOS/install/commands/grill-me.md
+++ b/LifeOS/install/commands/grill-me.md
@@ -1,0 +1,17 @@
+---
+name: grill-me
+description: Relentless checkpointed discovery interview that interrogates a half-formed idea until its shape is clear, then hands off to ISA Scaffold. Routes to the ISA skill's Grill workflow. USE WHEN /grill-me, grill me, discovery interview, figure out the shape, brainstorm before scaffolding an ISA. NOT FOR filling an existing ISA's thin sections (use ISA Interview).
+argument-hint: <topic> [slug]
+---
+
+# grill-me — ISA Grill workflow
+
+Invoke the ISA skill's **Grill** workflow on `$ARGUMENTS`:
+
+```
+Skill("ISA", "grill me on $ARGUMENTS")
+```
+
+If `$ARGUMENTS` is empty, ask what idea/plan/decision to grill, then proceed.
+
+Grill walks the design tree one question at a time (recommending an answer each turn, exploring the codebase instead of asking when it can), checkpoints every turn to `MEMORY/WORK/{slug}/grill.md`, runs a pre-mortem→ISC pass, and hands off to `ISA Scaffold`. Full procedure: `~/.claude/skills/ISA/Workflows/Grill.md`.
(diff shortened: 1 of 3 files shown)


danielmiessler:
Ported: the Grill discovery-interview workflow ships in the ISA skill (skills/ISA/Workflows/Grill.md) with your credit, routed from /grill-me. Closing rather than merging because the public repo is generated from private source; your work is in the source of truth and in v7.40.4.

PR #1607, opened 2026-07-24
docs: discoverable doc headers for every flat tool in LIFEOS/TOOLS

anikinsasha:
## Summary

Every flat (single-file) tool in `LIFEOS/TOOLS/` now carries a doc header that says what it is, how to run it, and where its governing doc lives. Comment-only — zero code changes.

Agents and humans alike decide what a tool does by reading its head. Most tools here already follow a strong de-facto convention (see `Inference.ts`, `WorkSweep.ts`, `ISARender.ts`); this PR brings the stragglers up to the same bar: on a census of the shipped root, ~25 tools were missing an identity line and/or usage, and several were completely headerless (`ForgeProgress.ts`, `HealthSnapshot.ts`, `SplitAndTranscribe.ts`).

## The header contract (codified from the corpus, not invented)

A flat tool's leading comment block carries:

1. **Identity line** — `<Name>.ts — <one-line purpose>`
2. **Purpose prose** — 1–6 lines: what it does, why it exists, its cardinal boundary if it has one
3. **Usage** — runnable invocation(s) for invocable tools; pure library modules instead state "Library, not a CLI" plus a `Consumed by:` line; dual modules carry both
4. **`@see` pointer** — only where a governing doc actually exists (skill, system doc, subsystem README)

That's the whole contract, stated here in full — this PR intentionally cites no standard file, since none ships in the repo today. A short standard doc codifying the contract may follow as a separate contribution.

## What's in the diff

- Header additions/gap-fixes across the flat root (including `PiSync.sh`), each derived from the file's actual code — argv handling read before writing usage lines.
- Banner theme variants (`BannerMatrix`, `BannerNeofetch`, `BannerPrototypes`, `BannerRetro`, `BannerTokyo`, `NeofetchBanner`) are honestly labeled as alternative prototypes not wired to the launcher (only `Banner.ts` is invoked).
- One substance fix: `GetTranscript.ts`'s usage examples pointed at a `skills/Videotranscript/` location that does not exist in the repo; corrected to the tool's real path.

## Verification

- Diffs are comment-only (checked mechanically: no non-comment line added or removed).
- A deterministic header audit (regex over the element contract) passes for every tool in the root after this PR.

---

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/AddBg.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/AddBg.ts
@@ -1,16 +1,16 @@
 #!/usr/bin/env bun
 
 /**
- * add-bg - Add Background Color CLI
+ * AddBg.ts — add a solid background color to a transparent PNG image
  *
  * Add a solid background color to transparent PNG images.
- * Part of the Images skill for LifeOS system.
+ * Companion to RemoveBg.ts in the Art toolchain.
  *
  * Usage:
  *   add-bg input.png "#EAE9DF" output.png
  *   add-bg input.png --ul-brand output.png
  *
- * @see ~/.claude/skills/Images/SKILL.md
+ * @see ~/.claude/skills/Art/SKILL.md
  */
 
 import { existsSync } from "node:fs";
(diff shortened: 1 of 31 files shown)


danielmiessler:

PR #1623, opened 2026-07-26
fix(pulse): build the /projects dashboard page (nav linked a missing page → 404)

elhoim:
## Problem

The dashboard sidebar links a **PROJECTS** item to `/projects` (`nav-manifest.ts:48`), but no page was ever built for that route — there's no `src/app/projects/page.tsx`. Clicking PROJECTS returns **404**.

The backend is fine: `GET /api/projects` already exists and returns real data parsed from `USER/PROJECTS.md`. Only the frontend page was missing. (Same class of bug as the recently-fixed `/security` page.)

## Fix

Adds a read-only `src/app/projects/page.tsx` that renders the existing `/api/projects` endpoint, modeled on the sibling pages (security/performance) and using the shared chrome kit (`PageShell`/`PageHeader`/`Panel`/`PanelHeader`/`StatTile`/`Pill`). It shows:

- A project count and an open-sessions count (StatTiles).
- A per-project list: name, path, stack, deploy command, and live URL, with an "open session" pill for resumable projects.

No backend change (the endpoint already exists), and no nav change (the nav entry already points here — this just fills the gap). The `out/` build artifact is intentionally not included; the build regenerates it.

## Verification

- `bun run build` in `Observability/` → compiled successfully, `/projects` route emitted (5.15 kB).
- `curl http://127.0.0.1:31337/projects` → **200** (was 404). No Pulse restart needed (static frontend rebuild).

## Affected file
- `LifeOS/install/LIFEOS/PULSE/Observability/src/app/projects/page.tsx` (new)

### DIFF
--- a/LifeOS/install/LIFEOS/PULSE/Observability/src/app/projects/page.tsx
+++ b/LifeOS/install/LIFEOS/PULSE/Observability/src/app/projects/page.tsx
@@ -1,0 +1,155 @@
+"use client";
+
+import { useState, useEffect, useCallback } from "react";
+import { FolderGit2, ExternalLink, Rocket, Layers, Radio } from "lucide-react";
+import EmptyStateGuide from "@/components/EmptyStateGuide";
+import {
+  PageShell,
+  PageHeader,
+  Panel,
+  PanelHeader,
+  StatTile,
+  Pill,
+} from "@/components/ui/chrome";
+
+interface Project {
+  name: string;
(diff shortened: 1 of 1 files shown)


danielmiessler:
The `/projects` page exists in source, so this can't merge, but your diagnosis of the 404 was right on the cut you tested. It ships with the next release. Closing as superseded, with thanks.

PR #1624, opened 2026-07-26
fix(pulse): add read-only System→Security page + GET /api/security (was 404)

elhoim:
## Problem

The dashboard's **System → Security** nav item (`nav-manifest.ts:80`) links to `/security`, but no page was ever built for that route, so it returned **404**. (Same class of bug as the `/projects` page.)

## What this adds — read-only only

A read-only security view. It surfaces the live three-layer model without touching any of it:

- **L1** — a static explainer of the constitutional (model-is-the-boundary) layer.
- **L2** — `permissions.deny` / `permissions.ask`, read live from `settings.json`.
- **L3** — the `DANGEROUS_PATTERNS` / `CREDENTIAL_PATHS` / `INJECTION_SHAPES` shape categories, imported **directly** from `hooks/lib/safety-classifier.ts` (not transcribed), so the page can't drift from what `Safety.hook.ts` actually enforces.
- **Telemetry** — a summary of `permission-cache.json` / `permission-decisions.jsonl` when present, with a graceful empty state when they aren't.

**No mutation surface anywhere.** `/api/security` is GET-only (any other method → 405). Crucially, this does **not** touch the deprecated `POST /api/security/patterns` / `/rules` routes in `observability.ts`, which deliberately return **410 Gone** (PATTERNS.yaml-style pattern management was removed in the 2026-05-06 security simplification — see `DOCUMENTATION/Security/README.md` § "What's NOT Here"). Requests to those paths still fall through to that tombstone unchanged.

## Files

| File | Change |
|------|--------|
| `LifeOS/install/LIFEOS/PULSE/modules/security.ts` | **new** — `GET /api/security` handler; reads L2/L3 live, summarizes telemetry, GET-only |
| `LifeOS/install/LIFEOS/PULSE/Observability/src/app/security/page.tsx` | **new** — the read-only page, modeled on the performance page, shared chrome kit |
| `LifeOS/install/LIFEOS/PULSE/pulse.ts` | **modified** — declares the module, loads it, and dispatches only exact `GET /api/security` (everything else under `/api/security/*` falls through to the existing 410 tombstone) |

`out/security.html` (build artifact) is intentionally excluded — the build regenerates it. `nav-manifest.ts` needs no change; the entry already existed.

## Verification (live)

- `bun run build` → `out/security.html` generated; `sudo systemctl restart lifeos-pulse` → clean, `status: ok`.
- `curl /security` → **200** (was 404).
- `curl /api/security` → real JSON: `l2.counts {deny:36, ask:8}` (live settings.json), `l3.counts {dangerous:22, credential:4, injection:7}` (live safety-classifier), telemetry empty-state.
- `curl -X POST /api/security/patterns` → **410 Gone** (pre-existing tombstone, untouched); `curl -X POST /api/security` → **405**.

### DIFF
--- a/LifeOS/install/LIFEOS/PULSE/Observability/src/app/security/page.tsx
+++ b/LifeOS/install/LIFEOS/PULSE/Observability/src/app/security/page.tsx
@@ -1,0 +1,303 @@
+"use client";
+
+import { useState, useEffect, useCallback } from "react";
+import {
+  ShieldCheck,
+  ShieldAlert,
+  KeyRound,
+  MessageCircleWarning,
+  Ban,
+  UserRoundPen,
+  BookOpenCheck,
+  Database,
+  Clock,
+} from "lucide-react";
+import EmptyStateGuide from "@/components/EmptyStateGuide";
+import {
+  PageShell,
+  PageHeader,
+  Panel,
+  PanelHeader,
+  StatTile,
+  Pill,
+} from "@/components/ui/chrome";
+
+interface ShapeItem {
+  label: string;
+  description: string;
+  pattern: string;
+}
+
+interface RuleItem {
+  rule: string;
+}
+
+interface DecisionRow {
+  ts: string;
+  tool: string;
+  cmd_prefix: string;
+  decision: string;
+  reasons: string[];
+  matched_pattern?: string;
+  cache: "hit" | "miss" | "n/a";
+}
+
+interface SecuritySnapshot {
+  generatedAt: string;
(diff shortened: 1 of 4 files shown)


danielmiessler:
A richer version already exists in source: a read-only Security page with L1/L2/L3 panels plus attack-surface and hooks-detail sections, and the same 410-tombstone handling you correctly avoided re-enabling. Your import-the-patterns-directly instinct (so the page can't drift from the enforcing code) is exactly how the source version works. The public cut predates it — closing as superseded.

PR #1625, opened 2026-07-26
docs(cli): drop hardcoded '6.0.0' version banner (install is 7.1.1)

elhoim:
## Problem

`Tools/Cli.md` opens with a hardcoded banner reading **LifeOS 6.0.0**, but this install is on 7.1.1 (`LIFEOS/VERSION`), and the master doc + `ARCHITECTURE_SUMMARY` both say 7.1.1. The hardcoded number drifts on every release.

## Fix

Drop the hardcoded version from the banner and point to `LIFEOS/VERSION` as the source of truth.

## File
- `LifeOS/install/LIFEOS/DOCUMENTATION/Tools/Cli.md`

### DIFF
--- a/LifeOS/install/LIFEOS/DOCUMENTATION/Tools/Cli.md
+++ b/LifeOS/install/LIFEOS/DOCUMENTATION/Tools/Cli.md
@@ -2,7 +2,7 @@
 version: 1.1.10
 ---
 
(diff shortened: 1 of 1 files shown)


danielmiessler:
Already fixed in source with the same conclusion you reached: the hardcoded version banner is gone in favor of no version claim. Public cut lag — closing as superseded.

PR #1626, opened 2026-07-26
docs(pulse): note the Linux systemd unit alongside the launchd plist

elhoim:
## Problem

`Pulse/PulseSystem.md` describes Pulse as launchd/macOS-managed ("single Bun process managed by launchd"), but Linux installs run it under the shipped systemd unit `com.lifeos.pulse.service`. A Linux operator following the doc looks for a launchd plist that does not apply.

## Fix

Note the Linux systemd unit as the non-macOS equivalent.

## File


danielmiessler:

PR #1627, opened 2026-07-26
docs(bunker): note the Bunker repo is per-machine and module degrades when absent

elhoim:
## Problem

The **Bunker** row documents a canonical repo `~/Projects/bunker` and a `bunker.isa.md` state file as if always present, but they are per-machine/private and absent on a fresh install. `Pulse/modules/bunker.ts` loads by default and reads the missing repo.

## Fix

Note that the Bunker repo is private/per-machine and the Pulse module degrades to an empty state when `~/Projects/bunker` is absent.

## File
- `LifeOS/install/LIFEOS/DOCUMENTATION/LifeosSystemArchitecture.md`


danielmiessler:
Declining the patch, but you found a real doc bug in the process: the path itself was stale (Bunker was folded into the main repo), so annotating degradation onto a dead path would have cemented the wrong fact. The doc now points at the real location, with credit to this PR for surfacing it.

PR #1628, opened 2026-07-26
feat(pulse): agents Actions list — newest-first default, date on non-today events, sort toggle

elhoim:
## What this improves

Three behaviors on the **Agents → Actions** event list (`ObservabilityDashboard` → `EventTimeline` / `EventRow`):

1. **Most-recent-first by default** when Actions opens. Sorting is now explicit on `timestamp` rather than a blind `reverse()`, so order is deterministic regardless of the incoming array order.
2. **Date on non-today events.** The Time column shows time only for today's events; for older events it prepends a zero-padded day + 3-letter month (`03 Jul`), and across a year boundary it also shows the year (`31 Dec 2025 07:31:22`).
3. **Sort toggle** next to the Time column header — an arrow button flips between most-recent-first and oldest-first. Auto-scroll-to-top only fires in most-recent-first mode (otherwise it would scroll away from new events).

## Files
- `LifeOS/install/LIFEOS/PULSE/Observability/src/components/activity/EventTimeline.tsx` — `sortOrder` state, explicit sort, header toggle button, scroll gate
- `LifeOS/install/LIFEOS/PULSE/Observability/src/components/activity/EventRow.tsx` — `formatTime()` date-aware rendering

Frontend-only; no backend change. `out/` build artifacts excluded (regenerated by `bun run build`).

## Verification
- `bun run build` in `Observability/` → compiled successfully, `/agents` route emitted; `curl /agents` → 200.
- Visual interaction (toggle, date rendering) not browser-verified — this headless host has no Chrome for Interceptor. Logic compiles clean; rendered behavior is deferred-verify.

### DIFF
--- a/LifeOS/install/LIFEOS/PULSE/Observability/src/components/activity/EventRow.tsx
+++ b/LifeOS/install/LIFEOS/PULSE/Observability/src/components/activity/EventRow.tsx
@@ -138,12 +138,31 @@ const EVENT_TYPE_LABELS: Record<string, string> = {
 
 function formatTime(timestamp?: number): string {
   if (!timestamp) return "";
-  return new Date(timestamp).toLocaleTimeString("en-GB", {
+  const d = new Date(timestamp);
+  const time = d.toLocaleTimeString("en-GB", {
     hour: "2-digit",
     minute: "2-digit",
     second: "2-digit",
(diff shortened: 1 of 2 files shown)


danielmiessler:

PR #1629, opened 2026-07-26
docs(hooks): fix hook census (29 registered / 9 dispatched) + document FormatGate

elhoim:
## Problem

`Hooks/HookSystem.md` line 22 gives a stale hook census: it says **30** distinct `.hook.ts` are registered in `settings.json` and **8** are dispatched-only. The actual numbers are **29 registered** and **9 dispatched-only**, and `FormatGate.hook.ts` (dispatched via `StopGates.hook.ts`) is missing from the list entirely.

## Fix

- `30 distinct … (31 counting …)` → `29 … (30 …)`
- `The 8 files …` → `The 9 files …`
- Add `FormatGate` to the `StopGates` dispatch group.

Verified against the live install: `settings.json` parse yields 29 distinct registered `.hook.ts`; 38 on disk; `FormatGate.hook.ts` present and referenced 4x in `StopGates.hook.ts`.

## File
- `LifeOS/install/LIFEOS/DOCUMENTATION/Hooks/HookSystem.md`


danielmiessler:
Superseded twice over: the source doc already lists FormatGate under StopGates, and the census has since been recomputed against settings.json to 37 registered / 48 on disk — so both your numbers and the doc's old ones were stale in different directions. The underlying observation, that a hand-maintained census rots, was right and prompted the recount. Closing with thanks.

PR #1630, opened 2026-07-26
chore(pulse): delete dead pulse-old.ts and pulse-unified.ts (superseded by pulse.ts)

elhoim:
## Problem

`LIFEOS/PULSE/` ships two dead, superseded entrypoints alongside the live `pulse.ts`:
- `pulse-old.ts`
- `pulse-unified.ts`

No doc references either, and every launcher (`start-pulse.sh`, `com.lifeos.pulse.service`, `manage.sh`) runs `pulse.ts`. Verified: zero references to `pulse-old`/`pulse-unified` anywhere in the tree outside the files themselves.

## Fix

Delete both files.

## Files
- `LifeOS/install/LIFEOS/PULSE/pulse-old.ts` (deleted)
- `LifeOS/install/LIFEOS/PULSE/pulse-unified.ts` (deleted)


danielmiessler:
Half of this was already done — `pulse-unified.ts` is gone from source. The `pulse-old.ts` half I'm declining, and for a reason worth stating: the Pulse ISA lists it explicitly under Out of Scope, "kept as history." You're right that it has zero code references; that's the point of keeping it. Deleting would mean reversing a written decision, which isn't a drive-by.

Thanks for the sweep — the dead-file instinct is the right one, and the `pulse-unified.ts` half of it landed.

PR #1633, opened 2026-07-26
docs(claude): fix wrong-case ISA/ paths to Isa/ in the routing table

elhoim:
## Problem

The routing table lists the ISA docs with a wrong-case `ISA/` path (`ISA/IsaSystem.md`, `ISA/IsaFormat.md`). The directory is `LIFEOS/DOCUMENTATION/Isa/` (case-sensitive on Linux), so those links fail an existence check. `ARCHITECTURE_SUMMARY` already uses the correct `Isa/` case.

## Fix

Change the two rows to `Isa/IsaSystem.md` and `Isa/IsaFormat.md`.

## File


danielmiessler:

PR #1634, opened 2026-07-26
docs(claude): correct ActivateImports.ts path to skills/LifeOS/Tools/

elhoim:
## Problem

`CLAUDE.md` references `Tools/ActivateImports.ts` (twice), with a note that paths are relative to `~/.claude/`. But `~/.claude/Tools/` does not exist; the real file is `skills/LifeOS/Tools/ActivateImports.ts`.

## Fix

Point both mentions at `skills/LifeOS/Tools/ActivateImports.ts`.

## File


danielmiessler:

PR #1635, opened 2026-07-26
docs(claude): add missing subsystems to the routing table

elhoim:
## Problem

`CLAUDE.md` calls itself "the routing table — it tells you where everything lives," but the on-demand doc list omits several shipped subsystems: Amber, Conduit, Router, Background services (Services/), Spinner, CoreComponents.md, and BrandAssets.md. All exist under `LIFEOS/DOCUMENTATION/`.

## Fix

Add the missing rows to the routing list, pointing at each subsystem's actual doc:
`Amber/AmberSystem.md`, `Conduit/ConduitSystem.md`, `Router/RouterSystem.md`, `Services/BackgroundServices.md`, `Spinner/SpinnerSystem.md`, `CoreComponents.md`, `BrandAssets.md`.

## File


danielmiessler:

PR #1636, opened 2026-07-26
docs(work): note USER/WORK config files are created by work-system setup, not shipped

elhoim:
## Problem

`Work/WorkSystem.md` (and CLAUDE.md) list `USER/WORK/config.yaml`, `labels.yml`, and `work_repo.json` as user-config files, but on a fresh install `USER/WORK/` contains only `README.md` and sample dirs — those three files don't exist yet. The doc presents them as if always present.

## Fix

Note in the config-file row that `labels.yml` / `config.yaml` / `work_repo.json` are created by the work-system setup (`SetWorkRepo.ts --bootstrap`, already documented lower in the same file), and only `README.md` ships by default.

## File
- `LifeOS/install/LIFEOS/DOCUMENTATION/Work/WorkSystem.md`


danielmiessler:

PR #1637, opened 2026-07-26
feat(lifeos): add 'doctor' subcommand so the advised 'lifeos doctor' works

elhoim:
## Problem

`Doctor.ts` and the statusline advise running `lifeos doctor` on a capability regression (`Doctor.ts:270,338` write "capability regression — lifeos doctor" into the statusline). But `lifeos.ts` has no `doctor` subcommand, so `bun LIFEOS/TOOLS/lifeos.ts doctor` errors with `Unknown command: doctor`. The working invocation is the more obscure `bun LIFEOS/TOOLS/Doctor.ts`.

## Fix

Add a `doctor` subcommand to `lifeos.ts` that execs `Doctor.ts` (forwarding any extra args like `decline <name>`), plus a help line. Verified live: `bun lifeos.ts doctor` now runs the capability check (exit 0) instead of erroring.

## File
- `LifeOS/install/LIFEOS/TOOLS/lifeos.ts`

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/lifeos.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/lifeos.ts
@@ -623,6 +623,7 @@ USAGE:
 
 COMMANDS:
   k update                 Update Claude Code to latest version
+  k doctor                 Run the LifeOS capability check (Doctor.ts)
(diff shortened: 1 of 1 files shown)


danielmiessler:

PR #1638, opened 2026-07-26
docs(pulse): document the 18 missing Pulse modules in the subsystem table

elhoim:
## Problem

`Pulse/PulseSystem.md`'s subsystem-module table lists only a handful of modules (Hooks, Telegram, iMessage, UserIndex, Doctor), but `LIFEOS/PULSE/modules/` ships 25. Eighteen shipped modules (bunker, memory, security, siri, syslog, wiki, work, telos, usage, projects, content, books, assets, hypotheses, local-intelligence, menubar, conduit, example-module, …) are undocumented.

## Fix

Add the missing modules to the table, each with a one-line description taken from the module's own header comment and its `modules/<name>.ts` path.

## File


danielmiessler:

Issue #1392, opened 2026-07-03
v6.0.0: shipped cron job assistant-heartbeat references missing module Assistant/checks/heartbeat.ts — healthz permanently degraded on fresh install


badosanjos:
### Summary
On a fresh v6.0.0 install the shipped cron job `assistant-heartbeat` fails permanently: the module it runs does not exist in the payload.

### Observed
```
{"level":"error","msg":"assistant-heartbeat failed","error":"Error: Script exited 1: error: Module not found \"Assistant/checks/heartbeat.ts\"\n","failures":3}
```
After 3 consecutive failures the scheduler skips it, and `/healthz` reports `status: "degraded"` forever on an otherwise healthy fresh install (`job assistant-heartbeat: 3 consecutive failures`).

### Environment
Ubuntu 24.04, LifeOS v6.0.0, Pulse via manage.sh (systemd).

### Expected
Either ship `PULSE/Assistant/checks/heartbeat.ts` or don't register the job by default; a missing optional module shouldn't keep health permanently degraded on day one.

catchingknives:
Confirming on a fresh **6.0.5** install (macOS): same root cause, and the job that trips the degraded state first is the every-minute `assistant-tasks` (`* * * * *`), not `assistant-heartbeat`. After 3 consecutive failures the scheduler parks it and `/healthz` reports `status: degraded` with `reasons: ["job assistant-tasks: 3 consecutive failures"]` permanently.

The startup half is also present: `pulse.ts` logs `Assistant module not available — Cannot find module ./Assistant/module from pulse.ts` because the `PULSE/Assistant/` directory is not in the 6.0.5 payload, which is the import-warning side tracked in #1419. So both halves are the same missing-`Assistant/`-directory cause; which of the four `assistant-*` jobs surfaces first is the only variation.

danielmiessler:
Thanks — and agreed on the principle: a missing optional module should never leave `/healthz` permanently degraded on day one. Root cause confirmed: the Assistant module is private and stripped from the release payload, while the shipped `PULSE.toml` still registers the job.

Fixed in the development source at the class level: the cron scheduler now preflights every script job at startup and disables any whose referenced script file isn't present on the install — one clear warning, no repeated failures, no degraded health. That covers this job and any future payload/config drift of the same shape. Ships with the next release. Closing as integrated.

PR #1639, opened 2026-07-26
docs(pulse): reconcile DA/Assistant module path and mark it private/stripped

elhoim:
## Problem

The DA/Assistant Pulse module is documented inconsistently:
- `DaSubsystem.md` says it lives at `modules/da.ts`, status "pending implementation".
- `PulseSystem.md` lists a live **Assistant** module at `Assistant/module.ts`.

Neither `modules/da.ts` nor `Assistant/module.ts` exists on a public install — it's a private module stripped from the public release (`pulse.ts` existence-checks `./Assistant/module` and omits the `/assistant` routes when absent, #1419).

## Fix

Reconcile both docs on the actually-imported path (`Assistant/module.ts`) and note it is a private module absent from the public release, so `/assistant` API routes don't exist publicly.

Note: the `assistant-*` cron jobs in `PULSE.toml` are already correctly `enabled = false` with an explanatory comment (#1392), so no PULSE.toml change is needed — the audit's claim that they ship enabled was stale.

## Files
- `LifeOS/install/LIFEOS/DOCUMENTATION/Pulse/DaSubsystem.md`


danielmiessler:

PR #1640, opened 2026-07-26
fix(docs): wire GenerateKnowledgeSchemaDoc into the doc-integrity pass

elhoim:
## Problem

`LIFEOS/TOOLS/GenerateKnowledgeSchemaDoc.ts` regenerates `MEMORY/KNOWLEDGE/_schema.md` from `KnowledgeSchema.ts` (the source of truth), and four docs describe both as live. But nothing ever invokes it — so it's dead code and `_schema.md` doesn't exist on a fresh install.

## Fix

Wire the generator into the doc-integrity pass (`DocIntegrity.hook.ts`, which already rebuilds the arch-summary on Stop): a small `RebuildKnowledgeSchema` handler runs the generator as a subprocess (it can't be imported — it runs `main()` on import), keeping `_schema.md` in sync with the schema and present on every install. Non-fatal on failure, mirroring the sibling handlers.

Verified: hook + handler typecheck; running the generator writes `_schema.md` (90 lines).

## Files
- `LifeOS/install/hooks/DocIntegrity.hook.ts` (wire the handler)
- `LifeOS/install/hooks/handlers/RebuildKnowledgeSchema.ts` (new)


danielmiessler:

PR #1641, opened 2026-07-26
docs: annotate private _* skill paths as absent from the public release

elhoim:
## Problem

System docs reference private, underscore-prefixed skills (`skills/_LIFEOS/…`, `skills/_ULWORK/…`, `_WRITING`, `_SURFACE`, `_NETWORK`, `_LIFELOG`, `_INCIDENT_RESPONSE`, `_HOMESECURITY`, `_HARVEST`) as if present, but `_`-prefixed skills are excluded from the public release — none of those paths exist on a public install. 53 such references span 12 docs, so a reader can't tell the absent paths are intentional.

## Fix

Add a one-line legend near the first private-skill reference in each of the 12 affected docs, noting that `_`-prefixed skill paths are private and absent from public installs. This covers every occurrence in the file without tagging each of the 53 mentions inline (keeps the docs readable). ALLCAPS tokens that merely look similar (`_API_KEY`, `_IDENTITY`, `_ALLCAPS` the convention term, etc.) are deliberately NOT touched — only real `skills/_NAME` paths.

## Files (12)
DOCUMENTATION: SystemUserBoundary.md, Memory/MemorySystem.md, Tools/Containment.md, Amber/AmberSystem.md, LifeosSystemArchitecture.md, Work/WorkSystem.md, Pulse/PulseSystem.md, Security/README.md, Writing/AIWritingPatterns.md, Config/ConfigSystem.md, LifeOs/LifeOsSchema.md, Security/NpmRapidResponse.md


danielmiessler:

PR #1643, opened 2026-07-26
fix: write settings.json and other critical config atomically

elhoim:
## The problem

`MergeSettings.ts` rewrites `settings.json` with a truncate-then-write. That file is the harness's entire configuration: hooks, permission rules, environment variables, statusline.

It runs at SessionStart as `bun SettingsBackport.ts; bun MergeSettings.ts --output <settings>` under a 15-second timeout. If the pair overruns — cold bun cache, slow disk, a busy machine — the harness kills the process group. If the kill lands after the file was truncated but before the ~60KB body is flushed, or the disk is full, what remains is partial JSON. The next launch reports invalid settings and every hook, permission rule and env var is gone.

There is no crash needed for this to bite: the window is opened by an ordinary timeout on an ordinary startup path.

The repo already ships the right primitive at `PULSE/lib/atomic-write.ts` (write to a unique temp file, then `renameSync` into place), and `settings.json` even allowlists a temp-file rename in its own permission rules. It just was not used here.

## The fix

All critical config writes go through the existing atomic primitive rather than a second implementation. Beyond the two sites originally identified, a search for the same pattern found five more, all now converted:

- `LIFEOS/TOOLS/MergeSettings.ts`
- `LIFEOS/TOOLS/SettingsBackport.ts`
- `LIFEOS/TOOLS/SyncIdentityToSettings.ts`
- `hooks/lib/work-config.ts`
- `skills/LifeOS/Tools/DeployComponents.ts`
- `skills/LifeOS/Tools/InstallHooks.ts`
- `skills/LifeOS/Tools/InstallSettings.ts`

`atomic-write.ts` itself gained failure-path cleanup so a failed write cannot strand a temp file that a later run might mistake for real config.

The content written is unchanged. This is purely about how it lands on disk.

## Tests

`test/pulse/lib/atomic-write.test.ts` and `test/tools/MergeSettings.test.ts`, 11 cases:

```
11 pass
0 fail
18 expect() calls
```

Separately verified with an independent probe that induces a real failure — a read-only target directory, so the rename genuinely cannot complete — rather than mocking it:

```
✓ normal write lands correct content
✓ no temp file left behind
✓ failed write reports failure
✓ ORIGINAL survives byte-identical
✓ survivor is still valid JSON
✓ no stray temp after failure
✓ contrast: naive open('w') truncates before any write — 0 bytes remained
✓ contrast: truncated file is invalid JSON
```

The last two cases demonstrate the old behaviour for contrast: `open(path, "w")` empties the file the instant it is called, before a single byte of new content exists. That is the window this PR closes.

### DIFF
--- a/LifeOS/install/LIFEOS/PULSE/lib/atomic-write.ts
+++ b/LifeOS/install/LIFEOS/PULSE/lib/atomic-write.ts
@@ -1,16 +1,39 @@
-import { writeFileSync, renameSync, mkdirSync } from "node:fs";
+import { writeFileSync, renameSync, mkdirSync, rmSync } from "node:fs";
 import { dirname } from "node:path";
 
-export function atomicWriteJSON(filePath: string, data: unknown): void {
+/**
+ * Write to a uniquely-named sibling temp file, then rename it over the target.
+ *
+ * `renameSync` within a filesystem is atomic, so a reader either sees the whole
+ * previous file or the whole new one — never a truncated prefix. This matters
+ * for any file whose loss breaks the next launch (settings.json et al): a
+ * truncate-then-write leaves invalid JSON behind if the process is killed
+ * mid-write (SessionStart hook timeouts kill the process group) or the disk
+ * fills. Here both of those fail on the temp file and the target is untouched.
+ *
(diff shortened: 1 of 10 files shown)


danielmiessler:

PR #1645, opened 2026-07-26
fix(algorithm): close ISC criteria only on a structured worker verdict

elhoim:
## The problem

`algorithm.ts` decided whether a parallel worker passed its criterion with a bare substring test:

```ts
if (stdout.includes(`RESULT: ${cId} PASS`) || stdout.includes(`${cId} PASS`)) {
```

The second clause subsumes the first, so the `RESULT:` anchor protected nothing. Both of these strings satisfied it while describing a **failure**:

- `RESULT: ISC-3 FAIL: I could not make ISC-3 PASS despite two attempts`
- `Verified whether ISC-3 PASSES: it does not.`

The worker's exit code was ignored for pass detection, used only for a display label. The criterion checkbox was then flipped to `[x]`, and once `postCriteria.passing >= postCriteria.total` the ISA status was set to COMPLETE with no re-verification.

So a failing worker whose prose happened to mention its own criterion id near the word PASS got that criterion checked off, and enough of those drove the artifact to COMPLETE. For a system whose entire premise is that a claim closes only on evidence, that inverts the doctrine: it reports verification it never performed.

## The fix

**A structured verdict line, not prose matching.** Workers now emit a sentinel-prefixed line:

```
ISC_RESULT v1 id=<criterion-id> verdict=PASS|FAIL reason=<one line, FAIL only>
```

The contract lives in one place (`IscResult.ts`) and is rendered into the worker prompt from the same constants the parser uses, so the two ends cannot drift.

**Fail closed on anything less than a clean verdict.** The judge returns `PASS`, `FAIL`, `CRASHED`, `AMBIGUOUS` or `NO_SIGNAL`, and only `PASS` closes a criterion. Contradictory output — a PASS and a FAIL line for the same id — is `AMBIGUOUS`, not a pass. A non-zero exit is `CRASHED` regardless of what stdout contains, so exit code is a necessary condition for passing.

**Regression checks parsed with the same rigour,** and a regression FAIL from any worker vetoes that criterion for the whole iteration, since one worker's fix can break what another claims to have passed.

**Operator-visible distinctions.** `NO_SIGNAL` and `AMBIGUOUS` are displayed as themselves rather than collapsed into a generic failure, because "the worker said it failed" and "the worker said nothing parseable" call for different responses. The dead `fallback detection` comment that had no code beneath it is gone.

## Tests

`test/tools/IscResult.test.ts`, 19 cases:

```
19 pass
0 fail
27 expect() calls
```

Separately verified with an adversarial probe written to break the parser rather than confirm it. All 19 cases behave correctly:

- Both original false-positive strings → not a pass
- Sentinel quoted inside prose, or wrapped in markdown bold → `NO_SIGNAL`
- Contradictory PASS + FAIL for one id → `AMBIGUOUS`
- Valid PASS line with a non-zero exit → `CRASHED`
- Verdict addressed to a different criterion → `NO_SIGNAL`
- **Id-prefix collision**: `id=ISC-30 verdict=PASS` does not close `ISC-3` → `NO_SIGNAL`
- Malformed or missing verdict key → `AMBIGUOUS`
- Empty and whitespace-only output → `NO_SIGNAL`
- Genuine PASS closes, genuine FAIL reports as FAIL, both alone and embedded in surrounding output
- Regression FAIL captured from the structured line while prose mentioning another id near "PASS" is ignored

## Note for review

This changes the worker contract, so a worker still emitting the old `RESULT: <id> PASS` shape now yields `NO_SIGNAL` and its criterion stays open. That is the intended direction — fail closed rather than open — but it is a behaviour change worth being deliberate about, since an in-flight run started before the upgrade will not close criteria until its workers use the new line.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/IscResult.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/IscResult.ts
@@ -1,0 +1,129 @@
+/**
+ * ============================================================================
+ * ISC RESULT — structured verdict contract between loop workers and the parent
+ * ============================================================================
+ *
+ * PURPOSE:
+ * A parallel loop worker reports whether its criterion passed over stdout. The
+ * parent checks the criterion's box in the ISA on that report, so the signal
+ * has to be something a *failing* worker cannot emit while narrating its
+ * failure in prose. Substring matching is not that: "RESULT: ISC-3 FAIL: I
+ * could not make ISC-3 PASS" contains a pass-shaped substring and describes a
+ * failure.
+ *
+ * THE CONTRACT:
+ *   ISC_RESULT v1 id=<criterion-id> verdict=PASS
+ *   ISC_RESULT v1 id=<criterion-id> verdict=FAIL reason=<one line>
+ *   ISC_REGRESSION v1 id=<criterion-id> verdict=PASS
+ *   ISC_REGRESSION v1 id=<criterion-id> verdict=FAIL reason=<one line>
+ *
+ * A verdict line must occupy a whole line, start at column 0, and match the
+ * grammar exactly. Prose mentioning a criterion id near the word PASS never
+ * parses. The worker prompt (see buildWorkerPrompt in algorithm.ts) states the
+ * contract as `verdict=PASS|FAIL`, so no complete, valid PASS line appears
+ * anywhere in the prompt — a worker echoing its instructions cannot produce
+ * one.
+ *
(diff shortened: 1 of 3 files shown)


danielmiessler:
Correct analysis against the released cut, but the file it fixes no longer exists: `algorithm.ts` was deleted in the phase-machinery strip, and there is no standalone runner anymore — the Algorithm runs in-session against the ISA. Nothing remains to port the fix into, so closing. The substring-PASS hazard you identified is a good general lesson and noted.

PR #1648, opened 2026-07-26
fix(pulse): map adapter models to inference levels that exist

elhoim:
## The problem

`AdapterRunner.ts`'s `modelToLevel()` could only return `"fast"`, `"smart"` or `"standard"`. Those names were removed from the inference layer on 2026-06-10. `Inference.ts` now validates against `low | medium | high | max` and throws on anything else:

```
[Inference] unknown level 'smart' — use low | medium | high | max
```

So every level the adapter could produce was rejected. Nothing caught it: `callInference` and `runAdapter` have no catch, `AdapterCli.ts` top-level-awaits `runAdapter` so the process dies, and `RebuildAll.ts` uses `Promise.allSettled`, printing each rejection as an unlabelled "unhandled" row. Only the cache-hit, no-sources and missing-prompt early returns avoid inference, so **100% of actual adapter builds failed**.

## Why it compiled

This is the part worth fixing properly. `Inference.ts` declared its own copy of the level names:

```ts
export type InferenceLevel = 'low' | 'medium' | 'high' | 'max';   // a copy
```

`models.ts` already had `EffortLevel` as the real source of truth. Two hand-maintained lists of the same names means a rename in one leaves the other stale and still type-checking — which is exactly what happened. The type mismatch didn't compile-error because the adapter was typed against a copy that agreed with itself.

## The fix

**Bind both ends to one source.**

```ts
export type InferenceLevel = EffortLevel;                              // alias, not a copy
const VALID_LEVELS: readonly InferenceLevel[] = Object.keys(EFFORT_MODEL) as InferenceLevel[];
```

The type now IS `models.ts`'s level type, and the runtime validator derives from `EFFORT_MODEL`'s keys. Rename a level in `models.ts` and every stale call site fails at the type level rather than at runtime. The error message and the CLI's own validation also read from `VALID_LEVELS` instead of repeating the literals a third and fourth time.

**Mapping**, extracted to `adapters/model-level.ts` and typed `InferenceLevel`:

| model family | level |
|---|---|
| haiku | `low` |
| sonnet | `medium` |
| opus | `high` |
| fable | `max` |
| anything else | `medium` |

Cheap models map to cheap tiers, and an unrecognised model gets a sane middle default rather than the most expensive one.

**Failures stop being unlabelled.** `callInference` catches a rejection and returns the same structured failure shape as every other error path, so one bad page reports as a named page with a real message and the batch keeps going.

## Tests

`test/pulse/adapters/model-level.test.ts`, 8 cases:

```
8 pass
0 fail
49 expect() calls
```

Separately verified that every level the mapping can emit is one the inference layer actually accepts, checked against `EFFORT_MODEL`'s real keys rather than a hardcoded list:

```
VALID_LEVELS (from EFFORT_MODEL): max, high, medium, low
✓ claude-haiku-4-5       → low
✓ claude-sonnet-5        → medium
✓ claude-opus-5          → high
✓ claude-fable-5         → max
✓ gpt-5                  → medium
✓ <empty>                → medium
✓ weird-unknown-model    → medium
```

No API calls are made by the tests or the check.

### DIFF
--- a/LifeOS/install/LIFEOS/PULSE/Tools/AdapterCli.ts
+++ b/LifeOS/install/LIFEOS/PULSE/Tools/AdapterCli.ts
@@ -19,7 +19,15 @@ if (!manifest) {
   process.exit(1);
 }
 
-const result = await runAdapter(manifest, { force });
+// Top-level await: an escaping rejection would kill the process with a bare
+// stack and skip the index refresh below. Name the page and exit cleanly.
+let result;
+try {
+  result = await runAdapter(manifest, { force });
+} catch (err) {
+  console.error(`error: adapter "${manifest.id}" crashed: ${err instanceof Error ? (err.stack ?? err.message) : String(err)}`);
+  process.exit(1);
+}
 
 // Refresh _index.json so the Pulse v2 sidebar reflects the new state.
 const all = loadAllManifests();
--- a/LifeOS/install/LIFEOS/PULSE/Tools/RebuildAll.ts
+++ b/LifeOS/install/LIFEOS/PULSE/Tools/RebuildAll.ts
@@ -18,6 +18,7 @@ const results = await Promise.allSettled(manifests.map((m) => runAdapter(m, { fo
 const wallMs = Date.now() - start;
 
 const entries: IndexEntry[] = [];
+const crashes: string[] = [];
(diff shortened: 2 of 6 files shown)


danielmiessler:

PR #1649, opened 2026-07-26
feat(delegation): add Persist workflow for work that must outlive the session

elhoim:
## What this adds

A `Persist` workflow in `Delegation` for the one case its other patterns don't reach: **local work that must outlive the session**. Agent teams and subagents end when the session does. Managed Agents survive disconnection but run in a cloud sandbox, so they can't touch this host's filesystem, repo, or `.env`. Detached tmux is what's left.

## Why it's this small

This started as a proposal for a full tmux cockpit skill — a wrapper plus five workflows, mirroring the macOS-only `CMUX` skill for Linux. A red team killed most of it before any code was written, and the two findings that mattered were:

- **Split-pane teammate visibility is already a setting, not something to build.** `teammateMode` accepts `in-process` (default since v2.1.179), `auto`, `tmux`, `iterm2`. `auto` selects tmux when the leader is already running inside tmux. Rebuilding boot-team/fleet/monitor would have duplicated a maintained backend.
- **Driving agents by keystroke puts them outside the control plane.** An agent started by typing `claude ...` into a pane is a sibling process — no subagent registry entry, no concurrency accounting, screen text instead of a result object, and whatever permission flags its own invocation carried. A `teammateMode` teammate inherits the lead's permission mode and its prompts bubble up to the lead. That asymmetry is now stated in the Gotchas, because it's the thing a reader most needs and nothing in the skill said it.

So the workflow is scoped to processes rather than agents: detached persistence, long-running non-Claude work, and TUIs with no API.

## The gotchas are measured, not inferred

Every one was reproduced against tmux 3.6:

| Gotcha | Evidence |
|---|---|
| `-t` resolves exact → fnmatch → **unambiguous prefix** | with `prod` and `production-work` present, `kill-session -t production` destroyed `production-work` — no wildcard, exit 0 |
| `send-keys -l` executes on an embedded newline | text containing `\n` ran with no Enter sent; the same bytes via `load-buffer` + `paste-buffer -p` did not |
| `remain-on-exit` is a **window** option | `set-option -t "$SID" remain-on-exit on` returns success and does nothing; panes still vanish. `-w` against a window target works and preserves `#{pane_dead_status}` |
| default `remain-on-exit off` destroys exited panes | exited pane absent from `list-panes`; `capture-pane` errors — a crashed job leaves no trace |
| `pane_current_command` resolves one level | `sleep 60` reports `sleep`; `bash wrap.sh` running the same thing reports `bash`, so "back to a shell" ≠ finished |
| `set-hook -gu` cleanup is destructive | removes the operator's own global hooks for that event alongside yours |
| `pane_current_path` can break `-F` templates | carries raw filesystem bytes, so a path may contain the delimiter or a newline; `#{q:}` doesn't escape them and `#{b64:}` doesn't exist in 3.6 |

The `remain-on-exit` one was found by running my own documented command sequence and watching step 5 fail — the first draft of this workflow had the silently-ineffective session-scoped form in it.

## Verification

The documented sequence was executed end to end after the fix:

```
session=$0 window=@0
  running: %0 dead=0 cmd=sleep
  after exit: %0 dead=1 status=7
teardown by ID ok
```

All probing ran on private `-L` sockets that were torn down; the default socket was never touched.

## Notes for review

- Delegation had no `## Workflow Routing` section, since its patterns aren't workflow-routed. I added one with a single row rather than restructuring the skill.
- Worth considering separately: the official troubleshooting advice for orphaned team sessions is `tmux kill-session -t <session-name>`, which is exactly the prefix-matching hazard above. A short session name there can kill a longer-named session the operator cares about.

### DIFF
--- a/LifeOS/install/skills/Delegation/SKILL.md
+++ b/LifeOS/install/skills/Delegation/SKILL.md
@@ -224,6 +224,12 @@ The tier delegation floors set a *minimum* fan-out. This gate sets the *ceiling*
 
 **Every inline brief and Workflow subagent prompt articulates the ideal state, not the procedure.** State WHAT a done result looks like (as testable outcomes), the CONSTRAINTS, and the TOOLS available — then trust the agent to find HOW. A brief that choreographs the agent's reasoning ("first search, then extract, then verify, then synthesize") is BPE-violating scaffolding: it caps a capable agent and rots as models improve. This is the highest-volume prompting surface in the system (subagent briefs are written constantly at runtime), so the discipline matters most here. Keep the four keep-classes — safety-gate, verified-gotcha, tool-contract, output-format-contract — and delete the choreography. Prefer stating the coverage OUTCOME over a hardcoded fan-out count ("trust a match only when 3+ independent sources align", not "spawn exactly 15 agents"). Full standard: `skills/Prompting/Standards.md` § Ideal-State Prompting.
 
+## Workflow Routing
+
+| Trigger | Workflow |
+|---------|----------|
+| "survive disconnect", "outlive the session", "run detached", "keep running after I log off", "long-running local process", "drive a TUI that has no API", "reattach later" | `Workflows/Persist.md` |
+
 ## Anti-Patterns (Don't Do These)
 
 - Don't delegate what Grep/Glob/Read can do in <2 seconds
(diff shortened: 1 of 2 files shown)


elhoim:
Correction pushed — the original version of this workflow carried a **wrong security instruction** and I'd rather flag it loudly than quietly amend it.

It said untrusted text should be sent via `load-buffer` + `paste-buffer -p`, on the belief that bracketed paste stops an embedded newline from executing. That is false. Retested into a ready `bash` pane:

| mechanism | result |
|---|---|
| `send-keys -l` | EXECUTED |
| `paste-buffer` | EXECUTED |
| `paste-buffer -p` | EXECUTED |
| `paste-buffer -r` | EXECUTED |
| `paste-buffer -p -r` | EXECUTED |

`man tmux`: on output, LF in a paste buffer is replaced with a separator (CR by default), and `-p` inserts bracket codes only *if the application has requested bracketed paste mode* — which no `#{...}` format exposes, so a caller cannot detect whether `-p` did anything at all. `-r` skips the LF→CR rewrite, but readline binds `accept-line` to both `\r` and `\n`.

The rule is now: **the control is the payload, not the transport.** Refuse text containing `\r`/`\n` unless submitting is explicitly intended. `load-buffer`/`paste-buffer` stays documented as ergonomics for large payloads, never as a safety boundary.

Also added, found while retesting: named paste buffers are server-global and readable from any session, a same-name `load-buffer` silently overwrites a concurrent caller's payload, `buffer-limit` only trims *automatic* buffers so named ones leak indefinitely, and `paste-buffer -d` does not delete the buffer when the paste fails.

PR #1650, opened 2026-07-26
fix(conduit): schedule capture + insight via Pulse cron so Conduit works on Linux

elhoim:
## Problem

Conduit's two scheduled jobs — the 120 s capture poll and the hourly insight read — are installed **only** via macOS `launchd` (`Conduit/InstallConduit.ts` and `InstallConduitInsight.ts`, which write to `~/Library/LaunchAgents` and call `launchctl`). On Linux there is no launchd, so neither job ever runs: nothing is captured, the hourly `BuildInsight.ts` never fires, and the Pulse **/conduit** page stays permanently empty. There was no non-macOS install path.

## Fix

Schedule Conduit through **Pulse's own cron heartbeat loop** instead of an OS-level daemon. Pulse already runs a cross-platform cron scheduler (`matchesCron`/`isDue` in `pulse.ts`), active on every platform Pulse runs on. Moving Conduit's two jobs there makes them work on **Linux and macOS alike**, with no launchd/systemd dependency:

```toml
[[job]]
name = "conduit-capture"
schedule = "*/2 * * * *"
type = "script"
command = "bun run Conduit/conduit.ts capture"
output = "log"
enabled = true

[[job]]
name = "conduit-insight"
schedule = "0 * * * *"
type = "script"
command = "bun run Conduit/BuildInsight.ts"
output = "log"
enabled = true
```

Jobs run with `cwd = PULSE/`, so the relative `Conduit/…` commands resolve, and Pulse's missing-script preflight already disables them cleanly on an install where the scripts aren't present.

### Why the insight job is safe under Pulse's minimal PATH

`BuildInsight.ts` uses the cheapest inference rung (haiku via `TOOLS/Inference.ts`), which resolves the `claude` binary **explicitly** (`Bun.which("claude")` → `~/.local/bin/claude` fallback). That's the same minimal-PATH problem the launchd insight installer solved by baking a PATH into the plist — so it's a non-issue here.

## Verified on Linux (headless)

- Pulse's cron loop fired `conduit-capture` on load → `captured 1 event(s)`, dispatched to log.
- `BuildInsight.ts` read the real events and wrote a content-type read (`Development 100%` with a narrative) — vs. the empty `[]` before.

## Scope / compatibility

- The macOS launchd installers are left in place for anyone who prefers standalone scheduling; this simply gives Conduit a working out-of-the-box path on every platform via Pulse.
- Single-file change to the public template `PULSE.toml`. No code changes.

### DIFF
--- a/LifeOS/install/LIFEOS/PULSE/PULSE.toml
+++ b/LifeOS/install/LIFEOS/PULSE/PULSE.toml
@@ -132,6 +132,30 @@ command = "bun run checks/airgradient-poll.ts"
 output = "log"
 enabled = true
 
+# ── Conduit (sensory capture) ──
+# Cross-platform scheduling via Pulse's own cron heartbeat — no OS-level daemon
+# (launchd/systemd) required, so Conduit works on Linux and macOS alike. These
+# replace the macOS-only launchd installers (Conduit/InstallConduit*.ts), which
+# never ran on Linux. capture is a cheap deterministic poll; the hourly insight
+# read uses the lowest inference rung (haiku, via Inference.ts — subscription
+# billing, resolves `claude` explicitly so Pulse's minimal PATH is a non-issue).
+
+[[job]]
+name = "conduit-capture"
+schedule = "*/2 * * * *"
+type = "script"
+command = "bun run Conduit/conduit.ts capture"
+output = "log"
+enabled = true
+
+[[job]]
+name = "conduit-insight"
+schedule = "0 * * * *"
+type = "script"
+command = "bun run Conduit/BuildInsight.ts"
+output = "log"
+enabled = true
+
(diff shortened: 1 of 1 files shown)


danielmiessler:
Declining this and #1650 together, since they're one decision.

Your diagnosis is right: those installers are launchd-only, so the jobs genuinely don't run on Linux. But the fix routes scheduling for jobs that currently run under launchd through Pulse cron, and on a macOS install that risks two schedulers owning the same job. I prototyped a single-owner guard to make it safe and then decided against the whole direction — adding a guard to protect the primary machine's scheduling from a change that only benefits a platform we can't test is the wrong trade.

The standing rule here is that platform support should be as universal as possible but strictly additive: a second backend beside the first, never a change to the path the primary install already runs. #1692 (systemd `--user` beside launchd for WorkSweep) is exactly that shape and landed today.

If you want to take these two again in that shape — a Linux-only backend inside each installer, launchd untouched — that would land.

PR #1655, opened 2026-07-26
docs: correct subagent nesting default (reversed in 2.1.219) and stale Algorithm LATEST pointer

elhoim:
Two stale-fact corrections in `DOCUMENTATION/`. No behaviour change.

## Subagent nesting default

The default has moved twice in three releases — 5 until 2.1.217, 1 in 2.1.217, 3 from 2.1.219 — and the guidance still said subagents cannot spawn subagents, steering away from a pattern the platform now enables by default. Records the current depth-3 default with `CLAUDE_CODE_MAX_SUBAGENT_SPAWN_DEPTH=1` to flatten it, plus the version history so the next move is easy to spot.

Also adds the two per-session ceilings (200 subagent spawns, 200 WebSearch calls) that a long fan-out session can actually reach.

Per your note on #1569, the concurrency-cap wording stays neutral on what happens to over-cap spawns — the changelog states the cap but not whether they queue or are refused.

## Algorithm version pointer

The row that teaches how to resolve the Algorithm version named 8.1.0 while `ALGORITHM/LATEST` reads 8.4.0. Line 26 of the same file was already correct, so this was a single missed edit rather than systemic lag.

---

Two notes so you can close this fast if it is already handled:

- #1569 is already ported, so the delegation half may be redundant. The genuinely new part is the **depth-3 correction** — #1569 covered the concurrency cap and nested-spawn-default-off, which was accurate for 2.1.217 but was reversed in 2.1.219.
- You mentioned `DelegationSystem.md` is retired to history-only status. This edit is offered on the same reasoning you gave there — a known-false claim shouldn't ride under a RETIRED banner. If the retirement means you'd rather not touch it at all, closing is fine.

### DIFF
--- a/LifeOS/install/LIFEOS/DOCUMENTATION/Algorithm/AlgorithmSystem.md
+++ b/LifeOS/install/LIFEOS/DOCUMENTATION/Algorithm/AlgorithmSystem.md
@@ -34,7 +34,7 @@ Spend follows the work, discovered from the task and its evidence gates — neve
 |---|---|---|
 | System prompt | Constitution: modes, output contracts, security, verification doctrine | `LIFEOS/LIFEOS_SYSTEM_PROMPT.md` |
 | Dynamic context | CLAUDE.md routing table, @-imports, hook-injected session context | `DOCUMENTATION/Config/ConfigSystem.md` |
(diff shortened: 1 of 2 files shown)


danielmiessler:
Neither stale fact exists in source: the Algorithm doc uses a `v{LATEST}` placeholder throughout, and DelegationSystem.md is retired history (which your PR text anticipated). The public cut lagged. Thanks for checking the pointers — closing as superseded.

PR #1656, opened 2026-07-26
fix: pin background:false on the remaining context:fork skills; bump CURRENT.opus to claude-opus-5

elhoim:
Two places where the harness or the model lineup moved and the payload didn't.

## `context: fork` skills were silently detached

2.1.218 changed skills declaring `context: fork` to dispatch as detached background agents by default, opt-out via `background: false`.

Seven skills declare `context: fork` — Research, Council, Ideate, Evals, SystemsThinking, RootCauseAnalysis, Knowledge — and their workflows all return results inline to the calling thread. Under the new default their output arrives as a task notification instead, so the invoking turn sees nothing. Pins the previous behaviour explicitly rather than inheriting a default that changed underneath the skill.

Note on overlap with #1570: that PR opted **Knowledge** out and you ported the documentation half into `SkillSystem.md`. This is the same reasoning applied to the **other six** skills, which have the same inline-return shape. If the ported `SkillSystem.md` guidance intends some of them to stay backgrounded, say which and I'll trim the set.

**Verified** — all seven `SKILL.md` files re-parsed after the edit; `background` lands inside the frontmatter block in each, with `context` and `background` both present.

## `CURRENT.opus` still pinned `claude-opus-4-8`

Claude Opus 5 shipped 2026-07-24 and is the default Opus model. `models.ts` calls itself the single edit point on a model release, and its header cites exactly this drift as the cautionary example (`ContextAudit` checking for `claude-opus-4-7` long after `claude-opus-4-8` shipped).

Applied via the repo's own `UpdateModels.ts --apply opus claude-opus-5`.

Blast radius is small — consumers taking the tier ALIAS were already auto-resolving, so only `pinnedModelForEffort()` and the drift scanner change behaviour.

**Verified** — `pinnedModelForEffort("high")` → `claude-opus-5`; `isCurrent("claude-opus-4-8")` → false.

Deliberately left alone: the dated probe records in `LifeosSystemArchitecture.md` and `ALGORITHM/changelog.md` are history and should keep naming the model that actually ran, and `cost-aggregator`'s `includes("opus")` fallback already returns correct pricing since Opus 5 holds Opus 4.8's $5/$25 rate.


danielmiessler:
Both halves were already in source when this arrived: the registry pins opus to claude-opus-5, and all seven fork skills carry `background: false`. The public cut you audited predates that. Correct call on both — just already made. Closing as superseded.

PR #1657, opened 2026-07-26
feat(interceptor): add measured zoom so appearance claims can close on real pixels

elhoim:
Adds `skills/Interceptor/Tools/Zoom.ts` — measured zoom for appearance verification.

## The problem

Images are downscaled to roughly 1568px on the long edge before a model sees them. On a 2000px-wide page a 40px logo arrives as a smudge, and cropping the *delivered* screenshot cannot recover it — the detail was destroyed upstream. That is the gap between "a screenshot proves this element exists" and "a screenshot proves what it looks like".

## The tool

Takes absolute pixel coordinates in the **original** image, crops from the full-resolution source, and magnifies that region to the image budget, so the pixels spent land on the thing under test.

```
bun Tools/Zoom.ts <image> --x N --y N --w N --h N [--out <path>] [--budget 1568]
```

- Nearest-neighbour on the upscale — magnifies without inventing detail.
- Never downscales: a region already past the budget passes through at native size, since shrinking it early would only throw away detail.
- Prefers `magick` (the binary `Capture.sh` and `VerifyImageProbe.ts` already use) and falls back to `ffmpeg`, so it works on hosts without ImageMagick.
- Out-of-bounds, zero-area and negative-origin crops exit 2 with the source dimensions in the message rather than emitting a misleading image.

A DOM-coordinate read (`eval`) gives you the region to pass.

## Verification

A 2000×1200 source with a 40×24 marker — the kind of detail a full-page screenshot destroys. Zooming a 60×44 region around it produced 1568×1150 at 26.13×, marker colour preserved, and the output was inspected visually: a speck in the source fills the frame in the zoom with the requested 10px margins on each side.

All eight guard paths exercised: out-of-bounds, zero width, negative origin, missing argument, missing file, no args, non-numeric budget, and an oversized region (passes through at 1.00×, confirmed still 2000×1200).

Also adds a gotcha to `Interceptor/SKILL.md` describing when to reach for it — logo, glyph, font rendering, icon, or few-pixel spacing claims.

No `.help.md`: none of Interceptor's eight tools carry one, and the tool self-documents via its usage error.


danielmiessler:

PR #1661, opened 2026-07-26
docs: document the remaining external tools (gh, ffmpeg, yt-dlp, fabric, jq) and add them to Doctor

elhoim:
Follow-on to #1660, applying the same logic to the rest of the tree.

## Method

Rather than grep for tool names, I enumerated what shipped code actually spawns:

- `Bun.spawn`/`spawnSync`/`execFile` array literals across `*.ts`
- command invocations in shipped `*.sh`
- every `command -v` / `Bun.which` guard — the strongest signal, since a guard means the code already knows the tool may be absent

That guard set came back as: `bash bun bunx claude codex fabric gws interceptor jq kitten magick rtk wrangler`. Four were documented, `bun` is in INSTALL.md, and `bash`/`claude` are foundational. The rest, plus the unguarded spawns, are what this PR covers.

## Load-bearing and undocumented — added to Doctor and the page

| Tool | Powers | Absence today |
|------|--------|---------------|
| `gh` | Work System's system of record | **fails outright** |
| `ffmpeg` | AudioEditor, transcript splitting, Conveyor | fails at spawn |
| `yt-dlp` | Feed YouTube source, Upgrade channel scan | step skipped |
| `fabric` | Fabric skill's pattern library | falls back to native prompting |
| `jq` | shell hooks that parse JSON | hook skips silently |

`gh` is the significant one. `WorkSweep`, `CommitmentSweep`, `CommitmentLog`, `RetagSweepTypes`, the Pulse work module and `ReminderRouter` all spawn it, **none with a fallback** — and `WorkSystem.md` states the GitHub repo *is* the system of record. So on a host without `gh`, work capture doesn't degrade, it fails, and nothing said so in advance. On the machine I audited from, `gh` is genuinely absent.

## Narrower ones — short table, no Doctor row

`rtk`, `kitten`, `cmux`, `mkcert`, `gws`. Each is tied to one terminal, one platform, or one opt-in workflow, and each skips cleanly when absent — a Doctor row would be a red ❌ on most hosts for something the user never wanted. They're listed so nothing is undocumented, without turning the check into noise.

## Deliberately excluded

`whisper` reads like a dependency — it appears in `ExtractTranscript.ts`, `SplitAndTranscribe.ts` and `Conveyor/Runner.ts`. It is the OpenAI `whisper-1` **API model**, not a local binary (`model: "whisper-1"` at `SplitAndTranscribe.ts:98`). That's an API-key concern, not an external tool, so documenting it here would have been wrong.

## Verified

- `Doctor.ts` runs clean with all 11 capabilities and exits 0, honouring its never-install-fatal contract.
- On this host it correctly reports `gh` **broken** with its fix command, while `ffmpeg`, `yt-dlp`, `fabric` and `jq` report live — so both branches are exercised by a real run, not just the happy path.
- Cross-check: every `id` in Doctor's registry has a corresponding mention in `GETTING-STARTED.md` — 11/11.

### DIFF
--- a/LifeOS/GETTING-STARTED.md
+++ b/LifeOS/GETTING-STARTED.md
@@ -95,7 +95,7 @@ These are narrower — tied to one terminal, one platform, or one opt-in workflo
 
 | Tool | Powers | Install |
 |------|--------|---------|
-| `rtk` | Transparent command compression on the Bash pre-tool hook. Skips silently without it (also needs `jq`). | see the rtk project |
+| `rtk` | Transparent command compression on the Bash pre-tool hook — compresses what the model *watches* (git status, test and build runs), never what it *reads*. Skips silently without it (also needs `jq`). | [rtk-ai/rtk](https://github.com/rtk-ai/rtk) — `brew install rtk`, or take the release binary. **Do not run `rtk init -g`**: that installs rtk's own global hook alongside LifeOS's `ContextReduction.hook.sh` and you get two rewriters on the same command. rtk also keeps a local `history.db` of proxied commands, and its telemetry is opt-in — `rtk telemetry disable` makes that explicit. |
 | `kitten` | Terminal tab naming and the retro banner, on the kitty terminal only. | ships with [kitty](https://sw.kovidgoyal.net/kitty/) |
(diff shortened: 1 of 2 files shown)


danielmiessler:

PR #1663, opened 2026-07-27
feat(tools): scheduled USER-tree backup to a private repo (Linux scheduling)

elhoim:
## What

Adds `LIFEOS/TOOLS/BackupUserData.ts`, a scheduled job that mirrors the USER tree to its private GitHub repo, plus its README and a `ConfigSystem.md` section documenting it.

## Why

`ConfigSystem.md` § Two-repo sync specifies the USER tree as its own private repo, kept in sync by `~/.claude/.git/hooks/pre-push`. That trigger has two preconditions: the operator has to actually push the SYSTEM tree, and `~/.claude` has to be a git repo. Neither holds on every install.

Where they don't hold, the USER tree — identity, DA identity, TELOS, projects, operational rules — has no off-machine copy at all. `BackupBeforeUpgrade.sh` doesn't close that gap either: it archives the SYSTEM tree and stores `LIFEOS/USER` as a bare symlink entry, deliberately.

This is the trigger-independent half of the same mechanism, not a replacement for the pre-push hook. On an install where the hook works, the two compose: the hook syncs on push, this syncs on a schedule.

## The four refusals

Fail-closed throughout; each is a non-zero exit with a logged reason, never a silent skip.

1. **Non-private remote.** Visibility is re-checked on *every* run via `gh repo view --json visibility,isPrivate`. `PUBLIC`, an API error, an unparseable answer, or a missing `gh` all block the push. Per-run rather than once at setup, because a repo can be flipped public at any time.
2. **The distribution repo.** A remote whose repo is named `LifeOS` (any case) or whose owner is the upstream is rejected outright. USER data never goes near the public release path.
3. **Secret material.** Staged files are scanned for private-key blocks and GitHub / OpenAI / Anthropic / Slack / Google / AWS credentials and JWTs before the commit is created. A hit unstages everything and aborts.
4. **A concurrent run.** An exclusive lock means two overlapping invocations produce one commit, not a race. Locks older than 30 minutes are treated as stale.

When nothing changed it exits `0` with `status=noop` and creates no commit, so history records real change rather than the clock. One JSONL line per run lands in `MEMORY/OBSERVABILITY/user-backup.jsonl`.

## Scheduling is Linux-only

Worth stating plainly, since it's the main limit on this PR. The tool itself is platform-neutral: it shells out to `git` and the GitHub CLI and nothing else. The **schedule** is a Linux user crontab, and that is the only scheduler it has been verified on. macOS operators should wrap the same command in a `launchd` agent (the shape LifeOS already uses for `com.lifeos.pulse.plist`) rather than assuming the crontab path works. The README and the doc section both say so.

## Verification

Verified on Linux against a real private repo, not a fixture:

- Round-trip: wrote a marker file, ran, read it back **from the remote** via `gh api`; deleted it, ran, the same call returned 404.
- No-op path: two consecutive runs on an unchanged tree left `HEAD` identical and created no commit.
- Privacy gate: `--check-visibility` against a real public repo refused (exit 1); against a private one, passed.
- Distribution-repo guard: rejected both `danielmiessler/LifeOS` and a fork's `LifeOS`, accepted a normal user-data slug.
- Lock: two parallel runs on a dirty tree produced exactly one commit; the loser logged `status=locked`.
- Cron: proven end-to-end under the real cron daemon with a temporary every-minute entry, and separately under `env -i` with a stripped cron environment. No auth or PATH failures.

Not verified on macOS or with `launchd` — see above.

## Notes for review

- No secrets are ever mirrored: `.env*`, key material, and the credentials directory are excluded, and credentials stay in `~/.claude/.env` as they do today.
- Auth is expected to come from a git credential helper, so no token ever lands in a remote URL or a log line.
- `LIFEOS_GH_BIN` exists because the GitHub CLI may be a wrapper rather than a plain `gh` on `PATH` (e.g. when LifeOS runs as a system user distinct from the operator's own account).
- Happy to reshape this as a skill rather than a bare tool if that fits the distribution better.

### DIFF
--- a/LifeOS/install/LIFEOS/DOCUMENTATION/Config/ConfigSystem.md
+++ b/LifeOS/install/LIFEOS/DOCUMENTATION/Config/ConfigSystem.md
@@ -1,5 +1,5 @@
 ---
-version: 1.2.5
+version: 1.2.6
 ---
 
 # LifeOS Configuration
@@ -80,6 +80,12 @@ The two trees are physically separate git repos:
 4. Post-push HEAD verification on both repos
 
 **Pre-flight refuses to proceed if the public LifeOS repo appears in either remote** — the workflow is explicitly for the two PRIVATE repos only. Public LifeOS release goes through the separate shadow-release pipeline with the 14 ShadowRelease gates.
+
+### The unattended path (`BackupUserData.ts`)
+
+The `pre-push` hook above only fires when the principal pushes the SYSTEM tree, and it presupposes `~/.claude` is itself a git repo, which not every install has. `LIFEOS/TOOLS/BackupUserData.ts` is the trigger-independent half: a scheduled job that commits and pushes the USER tree whenever it changed, and no-ops when it didn't. It carries the same boundary gates in tool form (remote confirmed `PRIVATE` on every run, failing closed on any other answer; distribution-repo remotes rejected; staged files scanned for secrets before commit; post-push HEAD comparison) plus an exclusive lock so it can't race itself or a manual push. One JSONL line per run lands in `MEMORY/OBSERVABILITY/user-backup.jsonl`.
+
+**Scheduling is Linux-only.** The tool is platform-neutral, but the shipped schedule is a Linux user crontab and that is the only scheduler it has been verified on. On macOS, wrap the same command in a `launchd` agent instead. Full contract, restore procedure, and known limits: `LIFEOS/TOOLS/BackupUserData.README.md`.
 
 ## Public releases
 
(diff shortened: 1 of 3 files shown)


danielmiessler:
Thanks @elhoim — the guards on this are seriously built (the per-run visibility check, the secret scan, the exclusive lock). I'm not going to land it, for three reasons, and one of them you'll want regardless. It runs `git add -A` in the user-data repo, which our conventions forbid outright. It duplicates a daily backup job that already exists. And the owner-heuristic that decides which remote is "safe" rejects any repo under the author's own GitHub account — which means the tool refuses to run on my machine, and it'll do the same to anyone whose user-data repo sits under their own account. That last one is a real bug worth fixing for your own use of it. Closing here.

PR #1664, opened 2026-07-27
fix(evals): derive judge model levels from the registry (claude-opus-5 was silently demoted to medium)

elhoim:
The three model-based graders each hardcoded a judge-model → inference-level map. The lineup moved and the maps didn't, so both current top-tier IDs fell through to the `?? 'medium'` default:

| judge_model | resolved to | should be |
|---|---|---|
| `claude-opus-5` | `medium` | **`high`** |
| `claude-sonnet-5` | `medium` | `medium` (right only by accident) |

The opus case is the damaging one: an eval configured with the strongest judge silently ran on Sonnet-tier inference, and nothing surfaced the demotion — the grader still returned scores, just from a weaker model than asked for.

## The fix

Derive the current lineup from `EFFORT_MODEL`/`CURRENT` in `models.ts` rather than restating it. That file already describes itself as the single edit point on a model release, and its header cites exactly this failure (`ContextAudit` checking for `claude-opus-4-7` long after `claude-opus-4-8` shipped) as the reason it exists.

```ts
const levelMap: Record<string, InferenceLevel> = {
  ...legacyLevels,
  ...Object.fromEntries(
    Object.entries(EFFORT_MODEL).map(([level, tier]) => [CURRENT[tier], level as InferenceLevel]),
  ),
};
```

Superseded IDs (`claude-opus-4-8`, `claude-opus-4-6`, `claude-sonnet-4-6`, and the two dated 2025 IDs) stay in a small `legacyLevels` table so existing eval configs keep resolving. `claude-fable-5` and `claude-haiku-4-5-20251001` came out of the hardcoded list because they're now produced by the derivation with identical values.

A future model bump flows through automatically instead of stranding a judge here.

## Verified

Resolution asserted against the shipped expression:

- `claude-opus-5` → `high`, `claude-sonnet-5` → `medium`, `claude-fable-5` → `max`, `claude-haiku-4-5-20251001` → `low`
- all five legacy IDs unchanged
- unknown model and empty string still fall back to `medium`

All three graders transpile clean.

Companion to #1656, which bumped `CURRENT.opus` to `claude-opus-5` — that bump is what made the stale maps observable.

### DIFF
--- a/LifeOS/install/skills/Evals/Graders/ModelBased/LLMRubric.ts
+++ b/LifeOS/install/skills/Evals/Graders/ModelBased/LLMRubric.ts
@@ -6,6 +6,7 @@
 import { BaseGrader, registerGrader, type GraderContext } from '../Base.ts';
 import type { GraderConfig, GraderResult, LLMRubricParams } from '../../Types/index.ts';
 import { inference, type InferenceLevel } from '../../../../LIFEOS/TOOLS/Inference.ts';
+import { CURRENT, EFFORT_MODEL } from '../../../../LIFEOS/TOOLS/models.ts';
 import { readFileSync, existsSync } from 'fs';
 
 export class LLMRubricGrader extends BaseGrader {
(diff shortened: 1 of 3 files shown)


danielmiessler:
Good diagnosis, and it was independently fixed in source the same day you filed: the three graders now share one `judgeLevelForModel()` derived from the model registry, resolving any generation by tier pattern rather than a lookup table, plus a deliberate grading ceiling. Your PR was correct against the released cut; the source fix supersedes it. Closing with thanks — this one crossed in flight.

Issue #1584, opened 2026-07-23
Support installing LifeOS into an isolated `CLAUDE_CONFIG_DIR`


ramazanpolat:
## Problem

LifeOS installs into `~/.claude` — the user's primary Claude Code config. Across `LifeOS/install/`, **391 files hardcode `~/.claude` / `$HOME/.claude` (1,780 occurrences)**, so the install root isn't really a parameter.

That means you can't try LifeOS without adopting it, can't develop it without mutating the assistant you're currently using, and can't run two configurations (work vs. personal, stable vs. dev, two principals on one machine) — which is a shame for a system that's fundamentally about per-principal identity.

The failure mode is quiet, too. `USER/CONFIG/LIFEOS_CONFIG.toml` ships `user_dir = "~/.claude/LIFEOS/USER"`. Move the config root without rewriting that and `LifeosConfig` still loads fine — it just resolves to a path that doesn't exist, `identity.ts` reads no frontmatter, and the install silently reports its DA as "LifeOS" and its principal as "User". Nothing throws, so the `try/catch` fallback in `paiUserDir()` never fires.

## Fix

Claude Code already supports this — it reads everything from `CLAUDE_CONFIG_DIR`:

```bash
claude                                                 # normal setup, ~/.claude
CLAUDE_CONFIG_DIR=~/.claude-lifeos claude              # isolated LifeOS
```

```bash
# ~/.zshrc
alias lifeos='CLAUDE_CONFIG_DIR="$HOME/.claude-lifeos" claude'
```

So the ask is just: **stop hardcoding the path, read the config root, default it to `~/.claude`.** No behavior change for existing users, and it can land incrementally by area (`LIFEOS/`, `skills/`, `hooks/`, `agents/`, `USER/`) rather than as one huge diff.

Happy to do that work and split it into reviewable chunks if you want it.

## Context

I maintain [claude-playbooks](https://github.com/ramazanpolat/claude-playbooks) — a playbook is just an isolated Claude Code instance, i.e. a directory you point `CLAUDE_CONFIG_DIR` at.

[...]

PR #1666, opened 2026-07-27
Support project-scoped installs (honor CLAUDE_CONFIG_DIR/LIFEOS_DIR everywhere)

### LINKED ISSUE #1665: Project-scoped installs (--config-root/CLAUDE_CONFIG_DIR) silently fall back to ~/.claude in ~60 files
The install Tools accept `--config-root`/`--config-dir` (or `CLAUDE_CONFIG_DIR`/`LIFEOS_DIR` env vars) to install LifeOS somewhere other than the global `~/.claude` — useful if you want it scoped to one project instead of leaking into every Claude Code session.

But past the install step, ~60 files across `install/LIFEOS`, `install/hooks`, and `install/skills` hardcode `homedir() + '.claude'` (or `$HOME/.claude` in shell) directly, bypassing those flags entirely. In practice:

- Hooks fire with commands pointing at files that were never placed there (`No such file or directory`) — `hooks.json` ships commands hardcoded to `$HOME/.claude/...`, and `InstallHooks.ts` merges them verbatim into `settings.json` without rewriting the prefix to the actual `--config-root`.
- `install/hooks/lib/paths.ts`'s `getClaudeDir()` — the central resolver several core hooks depend on (`PromptProcessing`, `EventLogger`, `LoadContext`, `AgentInvocation`) — only honors `CLAUDE_PLUGIN_ROOT`, not `CLAUDE_CONFIG_DIR`, so it silently falls back to the real home dir.
- Dozens of memory/knowledge/Pulse tools construct their working paths as `pathResolve(homedir(), ".claude/...")` with no env-var check at all.

Net effect: a project-scoped install (`--config-root <project>/.claude`) works during setup, then silently reads/writes the real global `~/.claude` once you actually use it.

Found and fixed while doing a real project-scoped install. PR incoming with the fix (60 files, all additive — env var checked first, falls through to the exact original default, zero behavior change for a normal global install).

rogdex24:
Fixes #1665.

LifeOS's install Tools already accepted `--config-root`/`--config-dir`, but ~60 files across the install payload hardcoded `$HOME/.claude` (or `homedir()`) directly, bypassing those flags. A project-scoped install (e.g. `--config-root <project>/.claude`, so LifeOS doesn't leak into every Claude Code session) would silently read/write the real global `~/.claude` instead once past the install step — hooks pointing at files that were never placed there, memory/knowledge tools writing to the wrong tree, etc.

## Changes

- **`InstallHooks.ts`** — `hooks.json` ships commands hardcoded to `"$HOME/.claude/..."`; rewrite that prefix to the actual `--config-root` before merging into `settings.json`.
- **`InstallSettings.ts`** — env template values under `$HOME/.claude` now expand to `--config-root` instead of unconditionally expanding to the real home dir.
- **`install/hooks/lib/paths.ts`** — `getClaudeDir()` now honors `CLAUDE_CONFIG_DIR` (it already honored `CLAUDE_PLUGIN_ROOT` for packed-plugin installs). This is the central resolver several core hooks depend on (`PromptProcessing`, `EventLogger`, `LoadContext`, `AgentInvocation`).
- **~55 further files** across `LIFEOS/TOOLS`, `PULSE`, `hooks/`, and `skills/*/Tools` — added the same `CLAUDE_CONFIG_DIR`/`LIFEOS_DIR` fallback ahead of each hardcoded `homedir()+".claude"` path, matching the convention already used correctly in `Conveyor/Ledger.ts` and `PULSE/modules/content.ts`. Every change is additive (env var checked first, falls through to the exact original default) — zero behavior change for a normal global install with these vars unset.
- **`ContextAudit.ts`** — its cross-reference validator only recognized `~/.claude/LIFEOS/...` as a valid absolute prefix, so it would flag every correct project-scoped doc reference as broken; added the missing branches.
- **Docs** — `INSTALL.md` and `Workflows/Setup.md` now document the existing `CLAUDE_CONFIG_DIR`/`LIFEOS_DIR` mechanism as the supported way to scope an install to one project, since nothing previously surfaced that this was possible.

Left untouched (confirmed cosmetic, no functional path): usage/doc strings, `console.log`/`error` message text, and a couple of files already correctly guarded by an existing env-var check.

## Test plan

- [x] `bash -n` on all edited shell scripts — pass
- [x] `bun build`/transpile parse check on all edited TypeScript files — pass
- [x] Verified via a real project-scoped install: `getClaudeDir()`/`getHooksDir()`/`getSkillsDir()`/`getSettingsPath()`/`getLifeosDir()` and Conduit's `CLAUDE_ROOT`/`DATA_ROOT` all resolve to the project path with `CLAUDE_CONFIG_DIR`+`LIFEOS_DIR` set
- [x] Confirmed no regression to default (unset-env-var) behavior — every fix is an additive `||`/`${VAR:-default}` fallback


danielmiessler:
Thanks @rogdex24 — the underlying complaint is legitimate: a bunch of files hardcode the config root past the install step. I can't take this diff as-is, though. It's 231 files touching credential-path resolution, and it makes `.env` resolution overridable by an environment variable, which cuts against a standing rule that `~/.claude/.env` is the one canonical location — a process that can set one env var shouldn't be able to redirect credential reads. It also edits modules that were retired weeks before this cut. The narrow, defensible piece — teaching `getClaudeDir()` about `CLAUDE_CONFIG_DIR` — I'll look at on its own. And whether LifeOS commits to being installable as a fully isolated instance is a real product-direction question I'm tracking in #1584. Closing this PR.

PR #1686, opened 2026-07-28
feat(knowledge): report note conformance on the doc-integrity pass (KnowledgeLint is invoked by nothing)

elhoim:
**Stacks on #1685.** That PR keeps `_schema.md` honest against `KnowledgeSchema.ts`, so the *doc* can no longer drift from the *code*. Nothing checks whether the **notes** conform to either.

## The gap this closes

`KnowledgeLint.ts` exists and works. It is invoked by **nothing in the tree** — every reference to it is prose, telling a human to run it. So note-level conformance is only ever measured when someone remembers to type the command.

That is how an archive reached **0.9% conformance** without a single warning:

1. A bulk import wrote 1,744 notes bypassing `MemorySystem.renderInitialNote`.
2. Later writers read a neighbouring note and copied its frontmatter, spreading the legacy dialect.
3. Nothing ever compared the notes to the schema, so none of it surfaced.

#1685 makes step 3's *documentation* half impossible. This makes the *data* half visible.

## The change

A read-only handler alongside the regen, in the pass that already runs.

- Calls `validate()` / `parseNote()` from `KnowledgeSchema` **directly** rather than shelling out to the CLI — one pass, no subprocess.
- Reports conformance percentage, per-directory breakdown, and the top violation classes.
- Names the repair command (`MigrateKnowledge.ts`) in the warning, so the next step is obvious.
- Emits `doc.integrity.knowledge_conformance` to `STATE/events.jsonl` for Pulse.
- **Reports only.** Never edits, never blocks. Non-conformance is a soft warning, matching `MemoryDirIntegrity`'s contract.

## Verified

On a 1,801-note archive:

```
[KnowledgeConformance] === Check complete (451ms, 100.0% conformant) ===
```

And with one note deliberately downgraded to `convention: kb-v2`:

```
[KnowledgeConformance] 1/1801 notes off-schema (99.9% conformant). Repair: bun LIFEOS/TOOLS/MigrateKnowledge.ts
[KnowledgeConformance]   1× convention — not kb-v3 (dialect not migrated)
```

Positive and negative both exercised. 451ms over 1,801 notes, so it is not a meaningful cost on the Stop path.

### DIFF
--- a/LifeOS/install/hooks/DocIntegrity.hook.ts
+++ b/LifeOS/install/hooks/DocIntegrity.hook.ts
@@ -20,6 +20,7 @@ import { handleDocCrossRefIntegrity } from './handlers/DocCrossRefIntegrity';
 import { handleRebuildArchSummary } from './handlers/RebuildArchSummary';
 import { handleMemoryDirIntegrity } from './handlers/MemoryDirIntegrity';
 import { handleRebuildKnowledgeSchema } from './handlers/RebuildKnowledgeSchema';
+import { handleKnowledgeConformance } from './handlers/KnowledgeConformance';
 
 async function main() {
   const input = await readHookInput();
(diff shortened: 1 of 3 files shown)


danielmiessler:

PR #1687, opened 2026-07-28
feat(knowledge): warn at write time when a note lands off-schema

elhoim:
**Stacks on #1686.** That PR reports non-conformance on the Stop pass. This one reports it at the write, while the author is still in the loop and the fix is one edit rather than a migration.

## Why a write-time signal

`MemorySystem.renderInitialNote` is the sanctioned way a note is born on the schema. **Nothing prevents a writer from bypassing it.**

A bulk import did exactly that and wrote 1,744 notes off-schema. Every writer afterwards read a neighbouring note, copied its frontmatter, and spread the legacy dialect further — including, on the archive where this was found, two agents that were explicitly trying to match house style. Copying a sibling is the intuitive move, and it is precisely the wrong one.

The convention was documented and unenforced. This is the smallest thing that makes it visible at the point of failure.

## Contract

- **Advisory only.** Emits `additionalContext` naming the violations and pointing at `renderInitialNote`.
- **Never blocks.** The note is already on disk by `PostToolUse`, and a hard gate on note-writing would be worse than the drift it prevents.
- **Narrow.** Silent for every path outside `MEMORY/KNOWLEDGE/<Type>/*.md`, including `_index.md` and `_schema.md`.
- Validates in-process via `KnowledgeSchema.validate()` — one file, no subprocess.

## Verified

| Input | Result |
|---|---|
| Conformant note | silent |
| `_schema.md` | silent |
| `Ideas/_index.md` | silent |
| `settings.json` path | silent |
| Note downgraded to `convention: kb-v2` | warns, names the violation, gives the repair route |

The warning it emits:

```
⚠ Knowledge note written off-schema — `Ideas/<slug>.md`
  - `convention` — not kb-v3 (dialect not migrated)

Notes are born on the schema via `MemorySystem.renderInitialNote`. Do not
hand-write frontmatter by copying a sibling — that is how a legacy dialect
spreads. Fix this note now, or run `bun LIFEOS/TOOLS/MigrateKnowledge.ts`
(dry-run by default) for a bulk repair.
```

## The set

Four PRs close this end to end: #1684 fixes a rule that could never be satisfied, #1685 stops the doc drifting from the code, #1686 makes note-level drift visible, and this makes it visible at the moment it happens.

### DIFF
--- a/LifeOS/install/hooks/DocIntegrity.hook.ts
+++ b/LifeOS/install/hooks/DocIntegrity.hook.ts
@@ -20,6 +20,7 @@ import { handleDocCrossRefIntegrity } from './handlers/DocCrossRefIntegrity';
 import { handleRebuildArchSummary } from './handlers/RebuildArchSummary';
 import { handleMemoryDirIntegrity } from './handlers/MemoryDirIntegrity';
 import { handleRebuildKnowledgeSchema } from './handlers/RebuildKnowledgeSchema';
+import { handleKnowledgeConformance } from './handlers/KnowledgeConformance';
 
 async function main() {
   const input = await readHookInput();
@@ -64,6 +65,12 @@ async function main() {
     await handleRebuildKnowledgeSchema();
   } catch (err) {
     console.error('[DocIntegrity] Knowledge-schema regen failed:', err);
+  }
+
+  try {
+    await handleKnowledgeConformance();
+  } catch (err) {
+    console.error('[DocIntegrity] Knowledge-conformance check failed:', err);
   }
 
   process.exit(0);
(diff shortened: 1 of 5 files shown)


danielmiessler:
Ported into source with credit — thank you. The public repo is generated from a private tree at release time, so the intent landed in source with your handle and this PR number at the fix site rather than merging here.

One honest caveat: the hook is in the tree but **not yet registered**, so it doesn't fire yet. Registration is a settings.json change that needs the maintainer's sign-off, and it's queued for that decision rather than slipped in. The guard itself is advisory-only and fail-silent, as you designed it.

Closing as ported; the registration decision is tracked separately.

PR #1692, opened 2026-07-29
Add systemd --user backend for WorkSweep scheduling on Linux

takanorinishida:
## Summary

`WorkSweep.ts` (the hourly Work System sweep — session catch-up, stale-issue flagging, project checks, TELOS-goal derivation) is already platform-neutral: it only uses `existsSync`/`readFileSync`/`Bun.spawn`, no macOS-specific calls. But `InstallWorkSweep.ts`, the script that schedules it, only knew how to talk to launchd. On Linux there was no way to get WorkSweep running on a recurring cadence at all.

This PR adds a `systemd --user` backend selected via `process.platform`, mirroring the pattern `LIFEOS/PULSE/manage.sh` already uses for Pulse itself (launchd on darwin, `systemctl --user` on linux). Two new unit templates pair with the existing plist template:

- `com.lifeos.worksweep.service.template` — oneshot service that runs `WorkSweep.ts`, same log target (`MEMORY/STATE/com.lifeos.worksweep.log`) as the launchd `StandardOutPath`/`StandardErrorPath`
- `com.lifeos.worksweep.timer.template` — `OnBootSec=2min` + `OnUnitActiveSec=60min` + `Persistent=true`, reproducing the plist's `RunAtLoad` + `StartInterval 3600` behavior

`InstallWorkSweep.ts` now branches on `process.platform`: the existing darwin/launchd code path is untouched, and a new linux/systemd path materializes both templates into `~/.config/systemd/user/`, runs `daemon-reload`, calls `loginctl enable-linger` (same survives-logout requirement already documented for `com.lifeos.pulse`), and `enable --now`s the timer. `--uninstall` and `--status` get matching linux branches.

**Update:** a code review caught that the `loginctl enable-linger` call originally read the username from `process.env.USER`, which isn't guaranteed to be set in every invoking environment (some non-interactive shells don't inherit it) — an empty string would make the call fail. Fixed by adding a `username()` helper (`id -un`, mirroring the existing `uid()` helper's `id -u`) and using that instead. Re-verified end to end after the fix (see testing evidence below).

## Testing evidence

Verified end-to-end on a real Linux host (Ubuntu, systemd):

```
$ bun ~/.claude/LIFEOS/TOOLS/InstallWorkSweep.ts
[InstallWorkSweep] detected bun at /home/.../bun
[InstallWorkSweep] wrote /home/.../.config/systemd/user/com.lifeos.worksweep.service
[InstallWorkSweep] wrote /home/.../.config/systemd/user/com.lifeos.worksweep.timer
[InstallWorkSweep] systemd timer enabled — com.lifeos.worksweep.timer active
NEXT                        LEFT     LAST  PASSED UNIT                       ACTIVATES
Wed 2026-07-29 00:55:48 UTC 59min... -      -      com.lifeos.worksweep.timer com.lifeos.worksweep.service

$ systemctl --user status com.lifeos.worksweep.service
● com.lifeos.worksweep.service ...
   Process: ... ExecStart=.../bun .../WorkSweep.ts (code=exited, status=0/SUCCESS)

$ bun ~/.claude/LIFEOS/TOOLS/WorkSweep.ts
[WorkSweep] repo=<redacted> since=24h max-create=50 APPLY
+ created #17 [Goal] ... [goal:G0]
+ created #18 [Goal] ... [goal:G1]
+ created #19 [BPE] Subtraction pass due — never run [bpe-audit-due]
[WorkSweep] done in 5092ms — sessions=0 new=0 stale=0 project-checks=0 goals=2 bpe=1
```

Also confirmed `--uninstall` cleanly disables + removes both unit files and `--status` reports the timer's next-run time. The macOS/launchd path is unchanged — only new `if (IS_LINUX)` branches were added, so no risk of regression for existing darwin installs.

After the `username()` fix: re-ran `install` and `--status`, both succeeded, and `loginctl show-user` confirmed `Linger=yes`.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/InstallWorkSweep.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/InstallWorkSweep.ts
@@ -56,6 +56,14 @@ async function uid(): Promise<string> {
   return r.out.trim();
 }
 
+async function username(): Promise<string> {
+  // Prefer `id -un` over process.env.USER — USER isn't guaranteed to be set
+  // in every invoking environment (e.g. some non-interactive shells), and an
+  // empty string would make `loginctl enable-linger ""` fail silently.
+  const r = await run(["id", "-un"]);
+  return r.out.trim();
+}
+
 async function detectBun(): Promise<string> {
   // Detect bun via `which bun` — survives different install paths (homebrew, ~/.bun, asdf).
   const r = await run(["which", "bun"]);
@@ -153,7 +161,7 @@ async function installLinux(): Promise<void> {
   await run(["systemctl", "--user", "daemon-reload"]);
 
   // Survive logout/reboot, same requirement PULSE/manage.sh documents for com.lifeos.pulse.
-  await run(["loginctl", "enable-linger", process.env.USER || ""]);
+  await run(["loginctl", "enable-linger", await username()]);
 
   const r = await run(["systemctl", "--user", "enable", "--now", TIMER_UNIT]);
   if (!r.ok) {
--- a/LifeOS/install/LIFEOS/TOOLS/InstallWorkSweep.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/InstallWorkSweep.ts
@@ -1,16 +1,24 @@
 #!/usr/bin/env bun
 /**
- * InstallWorkSweep.ts — Materialize com.lifeos.worksweep.plist.template and bootstrap it.
+ * InstallWorkSweep.ts — Materialize the WorkSweep scheduler unit(s) and bootstrap them.
  *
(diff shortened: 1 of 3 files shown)


danielmiessler:
Ported into source with credit — thank you, and welcome.

The public repo is generated from a private tree at release time, so a merge here would be reverted at the next cut; the intent landed in source instead, with your handle and this PR number in the code comment at the fix site.

This is exactly the shape platform support should take here: a second backend beside the existing one, selected by `process.platform`, with the macOS launchd path left alone. I verified that before landing it — every deletion in the diff is a doc comment, a moved const, or a dispatch line that now ternaries into the same original functions, and `--status` still reports live launchd state on macOS.

Closing as ported.

PR #1696, opened 2026-07-29
Fix Books.md template: restore 'Daniel Kahneman' as the book author

anikinsasha:
The `{{PRINCIPAL_NAME}}` placeholder substitution caught the author's first name in the Books starter template, so every install renders line 17 as `<principal name> Kahneman` (e.g. "Alice Kahneman") in the suggested-reading list.

**Before:** `- **Thinking, Fast and Slow** — {{PRINCIPAL_NAME}} Kahneman`
**After:** `- **Thinking, Fast and Slow** — Daniel Kahneman`

This is the only content-level placeholder hit in the payload — `grep -rP '\{\{PRINCIPAL_NAME\}\} [A-Z][a-z]+' LifeOS/install` returns just this line ({{DA_NAME}} has no equivalent hits).


danielmiessler:
Thanks @anikinsasha — Kahneman is already restored in the current source, so this one is in hand. Closing here since the public repo is generated from that source; it'll be in the next release.

PR #1704, opened 2026-07-30
feat(linux): add systemd --user backend to 4 scheduler installers

takanorinishida:
## What's broken

`DerivedSync.ts` / `CommitmentSweep.ts` / `UsageAggregator.ts` / `CodexUpdate.ts` have no macOS dependency in their service bodies (they're plain bun scripts), but their installers only know how to write a launchd plist — so none of the 4 can be scheduled on Linux.

`DerivedSync` in particular has real user-visible impact: it's what keeps `PRINCIPAL_TELOS.md` (imported by `CLAUDE.md`) and `LIFEOS_STATE.json` (read by the statusline rings) in sync with hand-edited TELOS files. Without it, those two files silently go stale after every TELOS edit.

## Fix

Follows the `process.platform` dispatch pattern `InstallWorkSweep.ts` established (#1692, closed/ported): darwin path untouched, linux path materializes a systemd `--user` unit pair instead of a plist.

- **derivedsync**: `WatchPaths` (7 dirs) + `RunAtLoad` → `.path` unit (`PathModified` x7) + `.service` (oneshot); the service is started once at install time to match `RunAtLoad`'s "fire once when loaded" behavior
- **commitmentsweep**: `StartCalendarInterval` 07:00 → `.timer` (`OnCalendar`, `Persistent=true`) + `.service`
- **usage-aggregator**: `StartCalendarInterval` 03:30 + `RunAtLoad` → `.timer` (`OnCalendar` + `OnBootSec=2min`) + `.service`
- **codexupdate**: `StartCalendarInterval` 04:00, no `RunAtLoad` → `.timer` (`OnCalendar` only) + `.service`

`ThrottleInterval` becomes `StartLimitIntervalSec`/`StartLimitBurst` on the `.service` (not on derivedsync's `.path` — `TriggerLimitIntervalSec` needs systemd 255+, unavailable on e.g. Debian 12's 252).

`InstallCommitmentSweep.ts`'s darwin plist hardcodes `/opt/homebrew/bin/bun` (no `detectBun()`) — left untouched. Only the new linux path calls a newly-added `detectBun()` to resolve `{{BUN}}` in the systemd templates.

Also fixes `BackgroundServices.md`'s now-inaccurate "macOS-only" claim in the Fresh Install section.

## Why this doesn't change macOS behavior

Every deletion in this diff is either a `const` moved a few lines up (`LABEL`, `STATE_DIR`, `COMMAND_TIMEOUT_MS` — same value, same file, just relocated next to the other platform-specific paths) or a dispatch line that now ternaries into the same original function, renamed `*Darwin` with its body byte-for-byte unchanged. No darwin behavior is touched.

## Verification

Tested end-to-end on Ubuntu 24.04 / systemd 255:

- `systemd-analyze --user verify` passes clean on all 8 new units
- **DerivedSync**: install → `.path` unit active → touching `TELOS.md` fired a targeted re-run (confirmed via the tool's own `derived-sync.jsonl` log — only the actions tied to the changed file ran) → `PRINCIPAL_TELOS.md` regenerated. Re-running install is idempotent. `--uninstall` removes both units cleanly.
- **UsageAggregator**: install → the boot-equivalent run fired immediately, exited 0, wrote `usage-daily.jsonl`. Timer correctly queued for the next 03:30.
- **CommitmentSweep**: install → timer correctly queued for the next 07:00; service not run (matches `RunAtLoad false`).
- **CodexUpdate**: timer registered for 04:00; service deliberately left un-started in this test (it has global-install side effects) — `systemctl --user is-active` confirms `inactive`.

Unrelated pre-existing bug found during verification, not fixed here (out of scope — it's in `DerivedSync.ts`'s own logic, not the installer/scheduler): `DeriveDenyHashes.ts` fails with `ENOENT` because `~/.claude/skills/_LIFEOS/` doesn't exist on this host. Would fail identically under launchd.

Related: #1692 (WorkSweep systemd backend, same pattern, ported).

### DIFF
--- a/LifeOS/install/LIFEOS/DOCUMENTATION/Services/BackgroundServices.md
+++ b/LifeOS/install/LIFEOS/DOCUMENTATION/Services/BackgroundServices.md
@@ -60,7 +60,7 @@ bun ~/.claude/LIFEOS/TOOLS/Services.ts doc                 # regenerate the tabl
 
 ## Fresh install
 
-The LifeOS install skill (`skills/LifeOS/`, `DeployComponents.ts`) stands up the core set during Setup. `Services.ts install --all` is the complete-coverage path: point a fresh macOS install at it and every service comes up, opt-ins included. macOS-only — `launchd` services skip cleanly on Linux/Windows. **Integration status:** `DeployComponents.ts` currently covers ~9 components; wiring it to read `Services.ts`'s registry so a single Setup step covers all 16 is the tracked follow-up.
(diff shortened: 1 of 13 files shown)


danielmiessler:
Thanks @takanorinishida — this is clean, well-verified platform work. I'm closing it as superseded rather than declined: #1698 covers these same four installers plus seven more behind a single shared typed spec, and I took that one, so landing both would put two different systemd idioms in the tree. Your point about DerivedSync silently going stale on Linux is correct and is preserved by the path that landed. Appreciate the Ubuntu verification you included.

PR #1705, opened 2026-07-30
feat(linux): add systemd --user discovery to Services.ts status/doc

takanorinishida:
## What's broken

`Services.ts`'s own doc comment says `status` "reports reality, not a hand-maintained guess" — but `loadedLabels()`/`findPlist()`/`cadenceOf()` only ever look at `launchctl`/`~/Library/LaunchAgents`, so on Linux every one of the 16 registered services reports "✗ missing" regardless of what's actually installed. After the systemd backends for WorkSweep (#1692) and DerivedSync/CommitmentSweep/UsageAggregator/CodexUpdate (#1704) land, that's no longer true — but `Services.ts` has no way to know it.

## Fix

Adds a `process.platform` branch to the three discovery functions:

- **`loadedLabels()`**: darwin parses `launchctl list`; linux parses `systemctl --user list-units --all --type=service,timer,path`, stripping the `.service`/`.timer`/`.path` suffix so a label counts as loaded from any of its unit files (same "loaded, not necessarily running" semantics `launchctl list` already had).
- **`findPlist()`**: darwin looks for an installed plist then a template in `TOOLS`/`PULSE`; linux looks for an installed unit (preferring `.timer` or `.path` over `.service`, since that's where the schedule lives) then a template, falling back to a bare `.service` for persistent daemons with no timer pairing (pulse — the systemd analog of a RunAtLoad-only launchd job) or a bare non-`.template` unit (pulse's own `manage.sh` keeps its systemd source that way, mirroring the plist side's existing dual lookup).
- **`cadenceOf()`**: darwin parses `StartInterval`/`StartCalendarInterval`/`WatchPaths`/`RunAtLoad` from XML; linux parses `PathModified`/`OnCalendar`/`OnUnitActiveSec`/`Type=simple` from the ini-style unit file.

`install`/`uninstall` need no branch — they just run each service's `install:` command, which is itself platform-dispatching inside the `Install*.ts` script; `Services.ts` never had to know how a service installs itself, only how to tell whether it did.

Services with no Linux port yet (`com.lifeos.pulse-menubar`, `com.lifeos.deriver` — both launchd-only by design, deriver explicitly per its own `manage-deriver.sh`: "Linux/systemd analog is intentionally not implemented") correctly continue to report "✗ missing" on Linux. That's fact, not a regression.

## Why this doesn't change macOS behavior

Every deletion in this diff is a function signature line whose body moved unchanged into a renamed `*Darwin` variant, with the original name now the platform-dispatching wrapper — same pattern used across `InstallWorkSweep.ts` and the four `Install*.ts` scripts this depends on.

## Verification

Tested on Ubuntu 24.04 / systemd 255 with all 6 currently-installed systemd services live (pulse, derivedsync, worksweep, commitmentsweep, usage-aggregator, codexupdate):

```
bun Services.ts status
  ● running    at load         Pulse (dashboard server)
  ● running    on file-change  Derived-file sync
  ● running    every 1h        Work sweep
  ● running    daily/scheduled Commitment sweep
  ● running    daily/scheduled Codex update
  ● running    daily/scheduled Usage aggregator
```

Every cadence string matches the actual installed unit content (`OnUnitActiveSec=60min` → "every 1h", `OnCalendar` daily entries → "daily/scheduled", derivedsync's `.path` `PathModified` → "on file-change", pulse's `Type=simple` with no timer/path → "at load"). The 10 not-yet-ported services (Conduit, healthsync, blogdiscovery, etc.) correctly show "✗ missing". `bun Services.ts doc` also verified — produces the same markdown table shape as the darwin-generated one, reflecting Linux's currently-narrower coverage.

darwin path unaffected: `bunx tsc --noEmit` shows only pre-existing `@types/node`-absence noise, no new errors; the 3 renamed `*Darwin` functions are byte-for-byte identical to their pre-change bodies.

Depends on: #1704 (the 4 installers whose systemd units this reads).

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/Services.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/Services.ts
@@ -11,17 +11,28 @@
  *   bun Services.ts doc                 # emit the canonical markdown table (for the doc)
  *
  * launchctl install/uninstall are the privileged steps; `status`/`doc` are read-only.
+ *
+ * Linux: `status`/`doc` branch on `process.platform` to read systemd --user
+ * units instead of launchd plists (same pattern as InstallWorkSweep.ts and
+ * the sibling Install*.ts scripts, which each materialize a systemd unit
+ * pair on Linux). `install`/`uninstall` need no branch here — they just run
+ * each service's `install:` command, which is itself platform-dispatching
+ * inside the Install*.ts script. Services with no systemd unit yet (macOS
+ * menu-bar app, deriver — both launchd-only by design) correctly report
+ * "✗ missing" on Linux; that's fact, not a bug.
  */
 import { existsSync, readFileSync } from "node:fs";
 import { homedir } from "node:os";
 import { join } from "node:path";
 
+const IS_LINUX = process.platform === "linux";
 const HOME = homedir();
 const CLAUDE = join(HOME, ".claude");
 const LIFEOS = join(CLAUDE, "LIFEOS");
 const TOOLS = join(LIFEOS, "TOOLS");
 const PULSE = join(LIFEOS, "PULSE");
 const LAUNCH_AGENTS = join(HOME, "Library", "LaunchAgents");
+const SYSTEMD_USER = join(HOME, ".config", "systemd", "user");
 const AMBER = join(LIFEOS, "USER/CUSTOMIZATIONS/ARBOL/Workers/_A_AMBER_LEDGER");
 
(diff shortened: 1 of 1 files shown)


danielmiessler:
Thanks @takanorinishida — the Services.ts Linux discovery. Ported into the source tree (the public repo is generated from it, so I'm closing this here), credited by number at the fix site. Ships in the next release. I took most of it but declined two hunks as regressions — the grep-lifeos filter and a seconds-vs-minutes cadence change.

PR #1706, opened 2026-07-30
feat(i18n): add hooks/lib/locale.ts, wire praise-vocabulary union into SatisfactionCapture

takanorinishida:
## What this does

Generalizes `hooks/lib/work-strings.ts`'s design contract (#1695: key-level fallback, fail-open, `paiUserDir()`-resolved locale files) into a reusable lib any surface can adopt.

## Design

Two lookup shapes, because prose and vocabulary need opposite fallback semantics — this is the actual design contribution, not just a rename of work-strings' machinery:

- **`t(lang, key, fallback, vars)` — OVERRIDE.** A locale bundle replaces the caller's default for a key it translates. No baked-in EN dictionary (unlike `work-strings.ts`, which legitimately owns one scoped to Work System issue bodies) — the caller supplies its own default, since a generic lib has no single key-space to own.
- **`tList(lang, key)` — UNION.** A locale bundle's array values are ADDED to the caller's own list, never replacing it. A Japanese locale should still recognize "perfect" as praise, not lose English vocabulary by opting in — this is what makes "locale unset → output unchanged" provable by construction rather than by convention: with no bundle, `tList()` returns `[]` and the caller's list is untouched.

`resolveLang()` priority: explicit arg > `LIFEOS_CONFIG.toml` `[principal].language` (new optional field) > `WORK.ISSUE_LANGUAGE` (read independently, not by importing `work-config.ts`'s unexported `loadIssueLanguage()` — keeps this lib standalone rather than reaching into Work System internals for one value) > `"en"`. Every step is try/catch-guarded and never throws, same fail-open contract `work-strings.ts` uses for locale bundles.

## First consumer

`SatisfactionCapture.hook.ts`'s positive-praise fast-path. `POSITIVE_PRAISE_WORDS`/`POSITIVE_PHRASES` stay exactly as they are — `EFFECTIVE_PRAISE_WORDS`/`EFFECTIVE_PRAISE_PHRASES` union in `tList(lang, "praise.words"/"praise.phrases")` only when non-empty, so with no locale configured they're the literal same `Set` object, not a copy. Also widens the fast-path's punctuation-stripping regex to full-width CJK punctuation (`。、！？…「」`) alongside the existing ASCII set — English prompts never contain these characters, so this is a no-op for them; it's what lets "完璧！" normalize to "完璧" and match locale vocabulary.

## Why this doesn't change English behavior

Every deletion in this diff is either a regex line replaced by a wider version of itself (same normalization purpose) or a Set-reference read that now points at a value which, absent locale configuration, IS the original Set (not a copy) — same "dispatch line branching to the original" pattern I used in #1704/#1705, applied to a value reference instead of a function call. `LifeosConfig.ts`'s changes are purely additive (one optional interface field, one line threading it through `validateAndNormalize`) — zero deletions.

## Verification

Isolated `LIFEOS_CONFIG_PATH`/`LIFEOS_DIR` probe harness (same style as #1703):

- **(a) locale unset** (no `[principal].language`, no `WORK/config.yaml`, no locale file) — `perfect`/`great job`/`nice`/`now do X`/`8` produce byte-identical output to pre-change; Japanese praise ("完璧", "ありがとう") correctly does NOT fire (no vocabulary configured, matches English-only baseline)
- **(b) `[principal].language = "ja"` + `ja.json` with `praise.words`/`praise.phrases`** — English praise still fires (union, not replacement); "完璧！", "ありがとう", "助かった" now fire rating 8; non-praise Japanese ("次はどうする", "これを直して") correctly produces no rating
- **(c) `WORK.ISSUE_LANGUAGE: ja`** (no `[principal].language` set) — confirms the fallback chain's third tier resolves and activates the same vocabulary independently of the toml field
- **(d) malformed locale JSON** — no throw, exit 0, English behavior preserved (bundle load fails closed, `tList()` returns `[]`)

Also verified on `develop` after merging alongside #1703 (the ASCII-guard fix, which touches a disjoint region of the same file): both changes coexist correctly — "2はあとで" is now rejected (per #1703) and "ありがとう" now fires as praise (per this PR) in the same merged file, confirmed via the same probe harness.

Related: #1695 (Work System locale mechanism, the design contract this generalizes), #1703 (the other fix touching this file).

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/LifeosConfig.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/LifeosConfig.ts
@@ -38,6 +38,10 @@ export interface LifeosPrincipal {
   timezone: string;
   hometown?: string;
   voiceCloneId?: string;
+  // ISO 639-1 code (e.g. "ja"). Optional — most consumers (hooks/lib/locale.ts,
+  // hooks/lib/work-strings.ts) fall back to WORK.ISSUE_LANGUAGE or "en" when
+  // this is unset, so existing TOML files need no edit.
+  language?: string;
 }
 
 export interface LifeosVoiceSettings {
@@ -169,6 +173,7 @@ function validateAndNormalize(raw: unknown, path: string): LifeosConfig {
       timezone: principal.timezone,
       hometown: principal.hometown,
       voiceCloneId: principal.voice_clone_id ?? principal.voiceCloneId,
+      language: principal.language,
     },
     da: {
       name: da.name,
--- a/LifeOS/install/hooks/SatisfactionCapture.hook.ts
+++ b/LifeOS/install/hooks/SatisfactionCapture.hook.ts
@@ -38,6 +38,7 @@ import { getLearningCategory } from './lib/learning-utils';
 import { getISOTimestamp, getPSTComponents } from './lib/time';
 import { captureFailure } from '../LIFEOS/TOOLS/FailureCapture';
 import { addRatingPulse } from './lib/isa-utils';
+import { resolveLang, tList } from './lib/locale';
 
 // Normalize env path vars that Claude Code injects without shell expansion (LifeOS#1404)
 for (const k of ["LIFEOS_DIR", "LIFEOS_CONFIG_DIR", "PROJECTS_DIR"]) {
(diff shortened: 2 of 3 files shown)


danielmiessler:
Thanks @takanorinishida — the engineering here is careful, and the union-not-override choice for vocabulary (with the same Set reference when locale is unset, so "unset means unchanged" is provable) is exactly right. I'm passing on it, though: it adds i18n configuration surface to a system built around a single English-speaking principal, to translate one praise-detection word list, and it depends on #1695 which I'm also not taking as a whole. The `tList()` union/override distinction is the idea I'd reach for if a locale layer ever becomes wanted. Closing.

PR #1714, opened 2026-07-31
feat(isa): persistent ISA backlog index — work stays findable after it leaves the push window

anikinsasha:
## The problem

An ISA is the memory of a piece of work. Today it goes invisible in two steps: it drops off the SessionStart block at 48h, then out of `work.json` at 7d, where the TTL **deletes** the row. After that the work exists only as bytes on disk that nothing tells you about.

That is freshness governing what *exists*, not just what is *pushed*. Those are different things, and only one of them should have a window.

## What this adds

`MEMORY/STATE/isa-index.json` — an append-only index of every ISA the instance owns, across both WORK trees and the persistent tool ISAs under `LIFEOS/TOOLS/` — plus a **Stalled ISAs** block in `LoadContext` that surfaces the unfinished ones long after they leave the push window.

The index keeps everything forever. What the *block* shows is bounded by `stalledMaxAgeDays` (30 days by default, `0` for genuinely any age), and the heading states whichever window is in force. Bounding the display is fine; bounding the memory is the thing this exists to prevent.

Three properties are load-bearing:

- **No entry is ever removed.** Not on TTL, not on cap eviction, not when the artifact leaves disk. A deleted artifact is tombstoned (`missing: true`) and un-marks itself if the file returns.
- **There is deliberately no retention knob.** Every *display* window is configurable; the index itself is not, because a retention knob on memory is the defect this removes.
- **`work.json` rows expire INTO the index**, never into the void, and the delete is **deferred** if the archive write fails. A row never leaves memory because a write lost a race.

## Knob-gated, conservative by default

Everything is governed by an `isaPickup` block in `settings.system.json`, so you get defaults and each principal decides whether to mandate more:

| Knob | Default |
|---|---|
| `enabled` | `true`. Switches off this index — every write path and the Stalled ISAs block. It does **not** govern the Recent Sessions block, which predates the index. With no index to archive into, work.json rows are then dropped on expiry rather than kept. |
| `stalledMaxAgeDays` | `30` (`0` = index-forever) |
| `stalledDisplayLimit` | `5` (a **display** cap, never a data cap) |
| `recentWorkWindowHours` / `recentWorkLimit` | `48` / `8` |
| `strandedAfterDays` / `reconcileMaxAgeDays` | `7` / `30` |
| `workJson.{nativeStartingHours,completeHours,defaultDays,capRows}` | `4` / `24` / `7` / `50` |

Verified end to end against a synthetic root: with stock defaults a 7-month-old ISA is held back; at `stalledMaxAgeDays: 0` it surfaces; at `enabled: false` the block disappears entirely.

## Three defects in `main` this also fixes

These are independent of the feature and worth reviewing on their own:

1. **`LoadContext` scanned only one WORK tree.** Sessions live in both `~/.claude/MEMORY/WORK` and `<LIFEOS_DIR>/MEMORY/WORK`; scanning one starved the block of exactly the work it exists to surface. It now scans both, keyed so identical directory names in the two trees cannot collide.
2. **"Done" meant two different things on one screen.** The Recent Sessions filter hardcoded `status === 'COMPLETED'` while the terminal vocabulary elsewhere is fuller (`complete`/`completed`/`closed`/`done`/`abandoned`/`archived`). A session marked `CLOSED` kept occupying a Recent slot forever. Both paths now route through `TERMINAL_PHASES`.
3. **`IsaReconcile --abandon-old-verify` rewrote stranded ISAs to `phase: complete`.** That erased the exact backlog signal this index exists to carry — a reconciler silently marking unfinished work as finished. Retired: it now prints a notice and proceeds, and stranded ISAs are flagged in the index instead. It was wired to nothing, so a hard failure would only punish someone typing the old command.

## The sweep never writes an ISA file

Stated precisely, because a vaguer version of this claim was wrong once already. Three independent legs:

1. **File.** `IsaReconcile.ts` contains no write call — no `writeFileSync`, no `writeFrontmatterField`.
2. **Call graph, at the write.** The only ISA write reachable from the sweep is `syncToWorkJson`'s v6.9.0 Resume After Complete write-back, and the new `SyncOptions.artifactReadOnly` suppresses it *beside the write itself*, not from a gate in another file. It suppresses the whole resume, not just the write: no phase flip, no iteration bump, no rework event. An observer observes.
3. **Call graph, by construction.** That branch is in fact unreachable from `--fix` even without the flag: the resume arms only when the registry phase and the frontmatter phase are both exactly `complete`, and that is precisely the state `classify()` calls `in-sync`, which the sync gate skips. Verified empirically against a flag-stripped copy across the whole reachable classify space (in-sync / drift / orphan, matching and mismatching case): zero ISA bytes changed in every case, including the three that did call `syncToWorkJson`.
4. **Destination.** The index writer's output path is operator-configurable, so the claim was refutable by pointing `LIFEOS_ISA_INDEX_PATH` at an `ISA.md` and letting the publish rename over it. The destination is now validated rather than merely described: it must be a `.json` file and must not be an artifact filename.

Leg 3 is why the flag is defense in depth rather than the load-bearing guard, and exactly why it is worth keeping — leg 3 is an emergent property of a classifier, leg 2 is a property of the callee that owns the write.

## Design notes

- **Indexing is not gated on `--fix`.** `--audit` means "no writes to `work.json` or ISA frontmatter"; append-only memory capture is not a mutation of state anyone reasons about. `--no-index` opts out.
- **Concurrency — one writer at a time, because nobody ever removes an existing lock.** A mkdir lock plus tmp/rename publish. `mkdir` is the exclusive claim, and it is *confirmed* by reading our own owner token back out of the directory — a claim that does not read back is not a claim, and proceeding on one would be the only remaining way to end up with two writers. If the lock already exists we **defer, unconditionally**: whoever owns it, however old it is, alive or dead or unidentifiable. No age check, no liveness check as a decision input. A lock is removed by exactly two actors — the owner that created it, and a human running `rm`. There is no third, so no interleaving of this module against itself can produce a second writer. Deferral is safe by construction: callers read `false` as keep-my-rows, and the one caller that cannot tolerate a skip (TTL eviction) checks that boolean before deleting. Readers never see a partial file. A corrupt index is quarantined aside rather than silently wiped.
- **Tolerant read path.** A large share of real ISAs put the H1 above the frontmatter block, which the strict parser reads as `null`. Skipping those would rebuild the very invisibility this removes, so the index uses a *separate* tolerant reader with derived fallbacks (directory name, H1, mtime, ISC counts). The strict parser is unchanged for every writer.
- **Tool ISAs are discovered by presence,** not registration: any `LIFEOS/TOOLS/<tool>/ISA.md`. They are index-only and never become `work.json` rows, since `work.json` is a per-session view and a tool ISA is not a session.
- **Keys are the artifact's directory** relative to `~/.claude`, so a `PRD.md` → `ISA.md` rename stays one entry rather than splitting into two.

## Notes for reviewers

- The sweep is registered **async** at SessionStart (30s timeout) so it stays off the blocking path. A session may therefore render the previous sweep's index; the live upsert in `ISASync` keeps actively-worked ISAs current regardless. This is memory, not telemetry.
- **Cost, stated plainly rather than as "cheap":** the blocking read is one JSON parse of a file that grows without bound by design, and the display cap is applied *after* the parse, so it bounds output rather than work. The async sweep walks every owned ISA. Both are inexpensive at realistic sizes — order 10ms at a few thousand entries, ~1.3 MB after five years at ~350 ISAs/year — but neither is constant-time. The index is written compact rather than pretty-printed, which roughly halves both. If it ever does matter, the fix is a compact push-view beside the full index, never a retention window.
- No tests are included: there is no hook-test harness in `LifeOS/install/` to add them to. A suite exists on my side and I am happy to contribute it separately if you want one.

## Review pass (second commit)

The published diff was reviewed and one **critical** correctness bug was found and fixed, along with eight places where code and prose disagreed. Detail is in the commit message; the headline:

`archiveWorkRows()` skipped any non-placeholder row whose `session.isa` did not resolve, then returned `true` once no entries remained — and the caller reads `true` as permission to delete *every* row in the batch. Pathless rows were therefore deleted unarchived, and a mixed batch archived the path-bearing subset while dropping the rest. That directly contradicted this PR's central promise. The function is now total: a row's memory lives on the row, not the file, so a pathless row is archived under a namespaced key and tombstoned, and a shortfall between memory-carrying rows and produced entries returns `false` (defer) rather than `true`.

The remaining eight were either fixed in code (quarantine-failure overwrite, entries-array validation, index destination aliasing, `phase:`/`status:` precedence, `enabled` not gating every write path, the `any age` heading, the `--quiet` stderr leak) or made honest in the text (what `enabled` actually governs, what the SessionStart read actually costs). One unrelated `getSessionAgents` change that had ridden in on the branch was reverted to upstream rather than fixed here — it belongs to its own change.

## Concurrency: how the lock ended up here

Several review rounds went at the write path, and the shape of the mistake is more useful than any individual fix. Every round found the same organ misbehaving in a new way: **this module removing a lock it did not create.** First the rule was "take it if it looks old", which cannot tell a writer that is stopped — SIGSTOP, a suspended laptop, a stalled network mount, a long GC pause — from one that is dead; the machinery built to survive that mistake grew to generation-named archives, a regression-repair pass and a pruner. Then the rule became "take it if the owner is provably dead", which is better and still wrong: two contenders can prove the same corpse and displace each other, putting a wrongly-captured lock back resurrects a directory another writer had moved on from, and `kill(pid, 0)` simply lies across pid namespaces and a relocated index path. Each answer was a narrower version of the same idea, and each one produced its own criticals.

So the mechanism is gone rather than refined. The rule is now total: **if the lock exists, defer.** The cost is real and is stated rather than hidden — a writer that dies holding the lock wedges every later index write until a person clears it. Nothing is lost while it is wedged (writers defer, and `work.json` rows that should have expired are kept rather than archived), but the surface silently stops updating, and a designed trade with a silent failure mode is just a silent failure mode. So the condition is surfaced where the human who must act will see it: the SessionStart **ACTIVE WORK** banner renders a notice naming how long the lock has been held, who holds it, what is accumulating meanwhile, and the exact command to clear it. It renders even when there is nothing else to show, because an empty banner is exactly what a wedged index looks like from outside. Nothing auto-escalates and nothing auto-clears.

```
📋 ACTIVE WORK:

  ── ⚠️  ISA INDEX: WRITES ARE BLOCKED ──

  🔒 The index lock has been held for 3h by pid 1 (appears alive).
     Nothing is lost while it is held: every writer defers, and work.json
     rows that should have expired are KEPT instead of archived. What does
     degrade is that the index stops gaining entries, so the backlog above
     goes stale and newly-finished work stops appearing in it.
     This will NOT clear itself — no code here removes a lock it does not
     own. If that process is gone or is not an ISA index writer, clear it:
       rm -rf <path>/isa-index.json.lock
```

What remains is a residual-window list re-derived for this protocol rather than edited: three windows (quarantine, commit, release), each reachable concurrently only if a human removes a live lock by hand. Acquire contributes none, because a failed claim leaves the directory exactly as it found it. The claim is exactly that and nowhere stronger: **no lock is ever removed except by its owner or a human; this code cannot create a second writer.** That is not "races are impossible" — an operator with `rm` can still make one, and the three windows say what it would cost.


danielmiessler:
Thanks @anikinsasha — the diagnosis here is genuinely sharp, and the framing that freshness should govern what *exists* rather than only what's *pushed* is the right way to see it. I took the three concrete fixes embedded in this into the source tree already (scanning both WORK trees in LoadContext, routing "done" through one terminal-phase set, and removing `--abandon-old-verify`), credited to this PR. What I'm not taking is the 2,100-line append-only index on the SessionStart path — a full-tree scan over the session directory at every session start is more cost in that hot path than the problem needs. Closing this one, but the parts that landed ship in the next release.

PR #1734, opened 2026-08-03
fix(pulse/observability): sample-entry regex, fd leak, unbounded caches, gendered string

elhoim:
**4 commits, one per fix**, so each can be cherry-picked independently. Each was verified to apply cleanly onto `47df8ee` by itself.

| Commit | Fix |
|---|---|
| `f2f01fd` | close the fd when readJsonlByteTail throws |
| `6ef670c` | evict stale series keys from the activity and climb caches |
| `6eb8d90` | drop the gendered pronoun from a shipped API response |
| `5d71e64` | match the shipped "(sample — …)" entries in SAMPLE_ENTRY_RE |

---

Four defects in `LIFEOS/PULSE/Observability/observability.ts`, all in code added or changed by 7.28.3.

- **`SAMPLE_ENTRY_RE` (~4216)** — `/^\**\s*\(sample\)/i` requires a literal `)`, so it misses the shipped `(sample — …)` entries. `MISSION.md:M2` alone made `realEntries(mission) > 0`, so `telosPersonalized()` returned true on a fresh install, template mode never armed, and Pulse showed the sample mission and goals to a new user as their own. Now `\(sample\b`, matching the sibling filter in `LIFEOS/TOOLS/GenerateTelosSummary.ts:164`.
- **`readJsonlByteTail` (~147)** — no `try/finally` around the fd, so a throw between open and close leaks it. Both sibling readers added in the same change already have the guard; this matches them.
- **`activityCache` / `climbCache` (~405)** — each series is capped, but the number of KEYS was unbounded: every slug and session_id ever folded stayed resident for the daemon's life. Adds key eviction on the fold path only.
- **`buildTeamFromUserFiles` (~3941)** — a hardcoded gendered pronoun in a shipped API response, in a function whose own comment says names never live in this file. Now phrased without one.

Verified: a scratch harness against the real shipped `USER/TELOS` scaffold — 28 sample bullets, 0 missed by the new regex, real entries still pass, eviction keeps live keys and drops stale.

One thing for the porter: the cache eviction is a real behaviour change for `climbCache`, not only memory. `buildIscDeltas` computes all-time totals from a slug's points, so an ISA quiet for >24h then resuming re-seeds from `prevDone = 0` and shows a one-off spike. 24h was chosen as 6x the widest consumer window (`RUN_ACTIVITY.nativeStaleMs`, 4h). Bounding key count instead of age is the alternative if all-time totals are meant to be durable.

---

### DIFF
--- a/LifeOS/install/LIFEOS/PULSE/Observability/observability.ts
+++ b/LifeOS/install/LIFEOS/PULSE/Observability/observability.ts
@@ -143,10 +143,14 @@ function readJsonlByteTail(filePath: string, maxBytes: number): any[] {
     const size = statSync(filePath).size
     const start = Math.max(0, size - maxBytes)
     const fd = openSync(filePath, "r")
-    const buf = Buffer.alloc(size - start)
-    readSync(fd, buf, 0, buf.length, start)
-    closeSync(fd)
-    let text = buf.toString("utf-8")
+    let text: string
+    try {
+      const buf = Buffer.alloc(size - start)
+      readSync(fd, buf, 0, buf.length, start)
+      text = buf.toString("utf-8")
+    } finally {
+      closeSync(fd)
+    }
     if (start > 0) {
       const nl = text.indexOf("\n")
       text = nl === -1 ? "" : text.slice(nl + 1)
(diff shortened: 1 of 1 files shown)


danielmiessler:
Thanks @elhoim — the Pulse observability fixes: the sample-entry regex, the fd leak, cache bounds, and the gendered string. Ported into the source tree (the public repo is generated from it, so I'm closing this here), credited by number at the fix site. Ships in the next release.

PR #1735, opened 2026-08-03
fix(observability): five React defects incl. Memory Graph crash on fresh installs

elhoim:

| Commit | Fix |
|---|---|
| `bbe709a` | stop the Memory Graph route crashing on fresh installs |
| `52552da` | clear the Ledger error on a successful poll |
| `d730b30` | clear the Algorithm regenerate interval on unmount |
| `9e04106` | make Live-sessions rows actually expand |
| `524028c` | abort in-flight CapabilityStrip fetches on window switch |

---

Five React defects under `LIFEOS/PULSE/Observability/src/`.

- **`app/memory/graph/page.tsx:257`** — `data.themes.map` crashes the whole Memory Graph route on a fresh install. `handleMemoryGraphApi` returns a default **200** carrying `communities` and no `themes` when the graph file is absent, and the only guard is `!data`, which a truthy themeless object passes. `themes` is now optional on the interface and the call site defaults it.
- **`app/ledger/page.tsx:90`** — the success path never cleared `error`, and `if (error)` replaces the whole page, so one transient fetch failure pinned the Ledger to its error state permanently.
- **`app/algorithm/page.tsx:275`** — `setInterval` created inside a click handler rather than an effect, leaking up to 15 minutes of `setState` on an unmounted component. Adds a ref plus an unmount effect, matching the file's existing idiom.
- **`components/activity/WorkBoard.tsx:950`** — the Live-sessions section hard-coded `expanded={false}` and a no-op `onToggle` while `BoardRow` still rendered the chevron, `cursor-pointer` and hover background, so the row advertised itself as clickable and did nothing. Wired to `expandedId`/`setExpandedId` exactly as the sibling sections are.
- **`components/activity/CapabilityStrip.tsx:153`** — no AbortController or generation guard, so after a window switch the slower 24h response overwrote the newer 1h one.

Verified: `tsc --noEmit` output is byte-identical to the clean tree (confirmed by stashing and re-running) — the one remaining error is pre-existing in `app/telos/_v7/app.tsx`, untouched here. `next build` was deliberately not run: `generateBuildId` uses a timestamp and `distDir` is the tracked `out/`, so it would rewrite committed artifacts, and `typescript.ignoreBuildErrors` is set so it adds no checking.

---

### DIFF
--- a/LifeOS/install/LIFEOS/PULSE/Observability/src/app/algorithm/page.tsx
+++ b/LifeOS/install/LIFEOS/PULSE/Observability/src/app/algorithm/page.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useCallback, useEffect, useMemo, useState } from "react";
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import {
   Workflow,
   ArrowRight,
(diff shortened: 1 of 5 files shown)


danielmiessler:
Thanks @elhoim — the five React fixes, including the Memory Graph fresh-install crash. Ported into the source tree (the public repo is generated from it, so I'm closing this here), credited by number at the fix site. Ships in the next release.

PR #1736, opened 2026-08-03
fix(pulse): unblock the event loop, bound two child spawns, hoist the cron parse

elhoim:

| Commit | Fix |
|---|---|
| `d3d7f20` | stop hypotheses blocking the event loop |
| `2ce4dc2` | drain stderr and bound the atlas insights spawn |
| `44815ef` | parse the cron expression once per scan, not once per minute |
| `c7dbe94` | fail loud on an output target this build no longer handles |
| `8d92c6b` | stop the menubar rendering "NaNd" for a down sidecar |

---

Five defects in the Pulse server modules. Two are regressions of bugs this same release fixed elsewhere.

- **`modules/hypotheses.ts:180`** — `execFileSync` with a 120s timeout inside an HTTP handler blocks Bun's single-threaded event loop; `/healthz`, the dashboard and hook validation all stall behind it. Matches the async shape this release already applied to `modules/bunker.ts`.
- **`modules/atlas.ts:48`** — `stderr: "pipe"` never drained and no timeout, so >64KB on stderr deadlocks the request and wedged children accumulate on every uncached poll. Matches the `modules/work.ts` fix from this same release: drain both streams concurrently under a kill timer.
- **`lib.ts:348`** — `mostRecentOccurrence` re-parsed the cron expression on every one of up to 10,080 minute-steps. The parse, not the field test, was the whole cost. Hoisted out of the loop; `matchesCron`'s public signature is unchanged.
- **`lib.ts:468`** — `telegram` was removed from `OutputTarget` with no `default:` branch, so a config upgraded from an older version keeps those jobs, sends nothing, and still reports success. Now logs an error.
- **`modules/menubar.ts:352`** — a clamped value was used for `tsMs` but the raw `NaN` reached `agoFrom`, rendering "Hermes sidecar down — NaNd". Clamps once and uses it for both.

`modules/upgrades.ts` is in the diff for one reason only: making `graduateHypothesis` async requires its caller to `await`. Not a drive-by.

---

### DIFF
--- a/LifeOS/install/LIFEOS/PULSE/lib.ts
+++ b/LifeOS/install/LIFEOS/PULSE/lib.ts
@@ -478,6 +478,14 @@ async function dispatchSingle(output: string, target: OutputTarget, jobName: str
 
       case "log":
         break
+
+      // Fail loud on a target this build no longer handles. `output` is cast
+      // unchecked at config load, so a config carrying an older target (e.g.
+      // `telegram`, since removed from OutputTarget) fell through the switch,
+      // sent nothing, and still reported the dispatch as a success.
+      default:
+        log("error", `Dispatch skipped for ${jobName}: unknown output target "${String(target)}" — valid targets are voice, ntfy, email, log`)
+        break
     }
   } catch (err) {
     log("error", `Dispatch to ${target} failed for ${jobName}`, { error: String(err) })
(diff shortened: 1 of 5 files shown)


danielmiessler:
Thanks @elhoim — the Pulse server fixes: unblocking the event loop, bounding child spawns, and hoisting the cron parse. Ported into the source tree (the public repo is generated from it, so I'm closing this here), credited by number at the fix site. Ships in the next release.

PR #1737, opened 2026-08-03
fix(tools): seven checks that reported the wrong result

elhoim:
**8 commits, one per fix**, so each can be cherry-picked independently. Each was verified to apply cleanly onto `47df8ee` by itself.

| Commit | Fix |
|---|---|
| `8012f9c` | skip the replay corpus check when the corpus is not installed |
| `65ff8cc` | match every writing-gate block reason in the recurrence ledger |
| `84d46b5` | reject a NaN iteration instead of writing null to the corpus |
| `25ca9a2` | create the reflections directory before appending |
| `31c1bfe` | stop ISARender clamping off-bracket phases to "Marking" |
| `8b6a303` | make the config-eval lock atomic |
| `cd00168` | notify when the behavioural suite cannot run |
| `3f9a521` | exit non-zero when CreateUpdate fails |

---

Seven defects under `LIFEOS/TOOLS/`. The common shape is a check that reports the wrong result rather than one that fails.

- **`IntegrityCheck.ts:863`** — `checkReplayCorpus` shells `bun test <CLAUDE_DIR>/test/regression` with no `existsSync` guard, but the payload ships no `test/` dir. It exits non-zero with no `(fail)` lines, so it raises a **blocking** finding: `/ic` is permanently red on every fresh public install. Now skips with a note, exactly like its sibling `checkRetirementRegistry`.
- **`RecurrenceLedger.ts:147`** — filtered `r.decision === "block"`, a value `hooks/WritingGate.hook.ts` never writes. The hook logs `block-strong-no-run`; `"block"` is its response to the harness, not its telemetry vocabulary. Every writing-gate block was dropped from the ledger and report. Prefix-matches `block*` so new reasons land without a third place to edit.
- **`Reflect.ts:200`** — `Number("x")` is `NaN`, which is `typeof "number"`, and `NaN < 0` is false, so a malformed `--iteration` passed `validate` and `JSON.stringify` wrote it as `null` into the corpus this gate exists to protect. Adds `Number.isFinite`.
- **`Reflect.ts:224`** — `appendFileSync` with no `mkdirSync`, where every sibling appender in this release does the mkdir first.
- **`ISARender.ts:41`** — `ASCENT_BRACKETS` is a three-slot subset of the six run states, so `indexOf` returns -1 for `verify`/`traverse`/`idle` and `Math.max(0, …)` sent all three to index 0. A `phase: verify` ISA rendered "Marking" in the bar while the badge on the same page said ANCHORING. Folds to the nearest earlier bracket by `ASCENT` order.
- **`ConfigEvalOnChange.ts:65`** — `existsSync`-then-write is not a lock; two sentinel files edited in the same instant both ran the suite and both notified. Uses the `wx` flag with stale reclaim and a lost-race fallback. It also swallowed a suite error into a JSONL line, so a suite that cannot **run** looked exactly like a clean pass on every config edit; that now notifies, same as a regression does.
- **`CreateUpdate.ts:678`** — `main().catch(console.error)` exits 0 on failure, and `IntegrityMaintenance.ts` checks only `code === 0`, so it recorded a ledger entry for work that never happened.

---

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/ConfigEvalOnChange.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/ConfigEvalOnChange.ts
@@ -97,7 +97,14 @@ async function main(): Promise<void> {
       );
     }
   } catch (e) {
-    log({ event: 'error', trigger, error: (e as Error)?.message ?? String(e) });
+    // Notify, don't just log. A suite that cannot RUN is the same operational
+    // fact as a suite that fails: the config change went unverified. Logging to
+    // a JSONL nobody tails made a permanently-broken suite — the shipped one is
+    // still legacy v1 `tasks:`, which the runner cannot execute — look exactly
+    // like a clean pass on every config edit.
+    const msg = (e as Error)?.message ?? String(e);
+    log({ event: 'error', trigger, error: msg });
+    await notify(`Behavioural eval could not run after editing ${trigger}: ${SUITE} errored (${msg}). The change is unverified.`);
   } finally {
     try { rmSync(LOCK, { force: true }); } catch { /* best-effort */ }
   }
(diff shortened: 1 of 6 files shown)


danielmiessler:
Thanks @elhoim — the seven checks that were reporting the wrong result. Ported into the source tree (the public repo is generated from it, so I'm closing this here), credited by number at the fix site. Ships in the next release.

PR #1738, opened 2026-08-03
fix(hooks): six defects in ModelRungGuard, VerificationGate, ISAFoldGate, EventLogger

elhoim:

| Commit | Fix |
|---|---|
| `f06dec4` | read only the transcript tail and skip sidechain rows |
| `2f04673` | let the verification gate split units on a newline between digits |
| `aee48b3` | decide hard-fail from is_error, not from words in the output |
| `382ac46` | stop ISAFoldGate reading a read-only command as a prod mutation |
| `0a7551a` | resolve the bun binary instead of spawning it bare |

---

Six defects under `hooks/`. These are guardrails, so each change is deliberately conservative: nothing previously blocked starts passing.

- **`ModelRungGuard.hook.ts:125`** — `liveModel()` does not filter `isSidechain`, so a subagent's model reads as the main loop's. Dispatching to a lower rung injected a false "MODEL RUNG" warning and wrote a bogus `off-pin` telemetry row.
- **`ModelRungGuard.hook.ts:118`** — read the entire transcript into memory before slicing `TAIL_BYTES`, contradicting the file's own "<20ms hot path" comment. Reads only the tail.
- **`VerificationGate.hook.ts:76`** — the digit-lookaround split refuses to break on a newline between two digits, so unrelated lines merged into one "unit" that could satisfy both halves of a T5 block, defeating the gate.
- **`VerificationGate.hook.ts:222`** — hard-fail was decided by a regex over `resultText` and never consulted `is_error`, so a **successful** `rg -n "exit 1"` tripped it via the new `"command"` event fallback.
- **`ISAFoldGate.hook.ts:48`** — `PROD_MUTATION_RES` matched raw command text, so a read-only `rg -n "Tools/Release.ts"` counted as a production mutation and blocked Stop. The 120-char truncation separately cut real `wrangler … INSERT` commands past the match point.
- **`EventLogger.hook.ts:235`** — spawned bare `bun` detached, the exact class `hooks/lib/resolve-bin.ts` was added this release to fix. ENOENT surfaced asynchronously past the sync try/catch, and the resulting missing state file made the self-heal fire on every tool call. Uses the helper.

---

### DIFF
--- a/LifeOS/install/hooks/EventLogger.hook.ts
+++ b/LifeOS/install/hooks/EventLogger.hook.ts
@@ -49,6 +49,7 @@ import {
 import { dirname, join } from 'path';
 import { execFileSync, spawn } from 'child_process';
 import { paiPath, getSettingsPath } from './lib/paths';
+import { resolveBun } from './lib/resolve-bin';
 import { getISOTimestamp, getPSTDate, getYearMonth } from './lib/time';
 import { bumpLastToolActivity, bumpLastToolActivityByUUID } from './lib/isa-utils';
 import {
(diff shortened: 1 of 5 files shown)


danielmiessler:
Thanks @elhoim — the six hook defects. Ported into the source tree (the public repo is generated from it, so I'm closing this here), credited by number at the fix site. Ships in the next release. I tightened the ISAFoldGate exemption a bit further on the way in — dropped less/bat and excluded rg --pre.

PR #1739, opened 2026-08-03
fix(skills,install): alias quoting, binding cascade, silent CLEAN, two eval defects

elhoim:
**7 commits, one per fix**, so each can be cherry-picked independently. Each was verified to apply cleanly onto `47df8ee` by itself.

| Commit | Fix |
|---|---|
| `383d20a` | quote the launcher paths written into the shell alias |
| `59e6627` | stop DeployCore running bun install across every user skill |
| `2cacf0f` | cascade binding removal to a fixpoint |
| `d4cbd2f` | never print CLEAN off a decode that failed |
| `911c78e` | survive a malformed user sources.json |
| `c6d1c0b` | diagnose a legacy v1 suite instead of throwing a TypeError |
| `23f7df8` | compute pass^k as all-k-passed, not as the mean |

---

Seven defects across the skills payload and the installer.

- **`install.sh:314`** (and its mirror at `skills/LifeOS/install/install.sh`) — the alias body interpolated `$LAUNCHER`/`$SYS_PROMPT` unquoted. A `$HOME` with a space breaks the launcher; one with a quote corrupts the rc file. By that point the awk pass has already commented out the working old alias, so the failure leaves the user with **no** alias. Adds `shq()` and quotes both parse levels. Tested with paths containing a space, a single quote, and a `$(…)` substitution: all four round-trip identically and nothing executes.
- **`skills/Tldraw/Tools/Tldr.ts:272`** — a single forward pass over bindings leaves the `start` binding dangling when the target shape is removed. Reproduced with the tool's own `validate`: `INVALID: binding:shape-a-start: dangling binding` before, `valid` after (3 records removed vs 4). Iterates to a fixpoint.
- **`skills/AudioEditor/Tools/GateRepair.ts:185`** — the verification decode's exit status was unchecked, so on decode failure it printed `CLEAN` and exited 0, certifying a file it never scanned.
- **`skills/LocalIntelligence/Tools/UserSources.ts:88`** — unguarded `JSON.parse`; a trailing comma in the user's `sources.json` aborted the run *after* all eight fetchers had succeeded, so nothing was persisted.
- **`skills/Evals/Tools/EvalRunner.ts:133`** — `for (const c of suite.cases)` throws a bare `TypeError`, because the only shipped suite (`Suites/Regression/core-behaviors.yaml`) is still the legacy v1 `tasks:` format. Callers that catch broadly turned that into a regression eval that silently never ran. I did **not** write a v1→v2 translation layer: `UseCases/*.yaml` define weighted graders and carry no `prompt`, so there is nothing for this single-shot runner to send, and guessing at the intended mapping seemed worse than a precise error. Before: `TypeError: undefined is not an object (evaluating 'suite.cases')`. After: a message naming the format, the 4 tasks, and the two ways forward.
- **`skills/Evals/Tools/EvalRunner.ts:149`** — `pass_to_k = passed/trials` is a mean, not pass^k: 2-of-3 reported 67% where the true pass^k is 0, making a flaky case look mostly-passing.
- **`skills/LifeOS/Tools/DeployCore.ts:277`** (and its mirror at `LifeOS/Tools/DeployCore.ts`) — walked all of `configRoot/skills` running `bun install` in every dir with a `package.json`, touching pre-existing user skills and double-installing the payload copy; failures land in `blockers`, so the documented offline install exits 1.

`DeployCore.ts` and `install.sh` are each shipped in two copies. Both copies were identical at `47df8ee` and both are updated here — a payload file fixed in one copy and left broken in the other is its own defect.

---

### DIFF
--- a/LifeOS/Tools/DeployCore.ts
+++ b/LifeOS/Tools/DeployCore.ts
@@ -24,7 +24,7 @@
 
 import { existsSync, mkdirSync, readdirSync } from "node:fs";
 import { homedir } from "node:os";
-import { join } from "node:path";
+import { basename, dirname, join } from "node:path";
 import { copyMissing, detectDevTree } from "./InstallEngine";
 
 // Runtime top-level entries this tool does NOT deploy:
@@ -242,10 +242,10 @@ function deployDependencies(payloadInstall: string, configRoot: string, apply: b
  * pulse.ts otherwise reports as a copy-paste fix command, so a fresh Pulse
  * doesn't 503 until a human intervenes.
  */
-function findNestedDependencyDirs(runtimeDst: string): string[] {
+function findNestedDependencyDirs(runtimeDst: string, skip: Set<string> = new Set()): string[] {
   const found: string[] = [];
   const walk = (dir: string): void => {
-    if (!existsSync(dir)) return;
+    if (!existsSync(dir) || skip.has(dir)) return;
     for (const entry of readdirSync(dir, { withFileTypes: true })) {
       if (entry.name === "node_modules" || entry.name === ".git") continue;
       const p = join(dir, entry.name);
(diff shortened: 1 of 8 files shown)


danielmiessler:
Thanks @elhoim — the alias quoting, binding cascade, silent CLEAN, and two eval defects. Ported into the source tree (the public repo is generated from it, so I'm closing this here), credited by number at the fix site. Ships in the next release.

PR #1744, opened 2026-08-03
Cloudflare collector: degrade per-call, don't discard the whole run

schmetti-dev:
## Summary

`collect()` threw on the first non-OK response from any account-level `cf()` listing call (`workers/routes`, `dns_records`, `storage/kv/namespaces`, `r2/buckets`, `d1/database`, `workers/scripts`, `workers/domains`) — discarding everything already gathered in that run (zones, DNS, workers, KV, R2, D1, all of it) the moment any ONE of those hit a permission gap.

Found running this against a real Cloudflare account with a narrowly-scoped API token (zone/DNS read access, but not the account-level Workers/KV/R2/D1 APIs — an entirely ordinary least-privilege setup, e.g. a token created for a specific deploy pipeline). The very first sync threw on the per-zone `workers/routes` call and produced **zero** Cloudflare assets, even though account + zone + DNS enumeration had already succeeded moments before.

This contradicts the collector's own doctrine ("a partial page, an API error, a rate limit → `PARTIAL` run, no sweep, prior state intact" — `Store.ts`'s sweep gate exists precisely to make this safe), which the bare `cf()` throw never actually implemented for this failure class.

## Fix

Added a `cfSoft()` wrapper: catches a failure on a single listing call, logs which one and why, marks the run `partial` (so `Store.applyRun`'s sweep gate correctly withholds sweeping instead of expiring real assets over incomplete data), and returns an empty array so the rest of the enumeration proceeds. Applied to every account/zone-scoped listing call **except** `/accounts` and `/zones` themselves, which stay hard failures — without an account or the zone list, nothing else in the collector is meaningful anyway.

## Testing evidence

Verified against the real account described above:

```
# Before the fix — one 403 on the first zone's routes call kills the whole run:
$ bun ~/.claude/LIFEOS/ATLAS/Atlas.ts sync cloudflare
✗ cloudflare: CF /zones/<id>/workers/routes -> 403

# After the fix — degrades per-call, keeps everything else:
$ bun ~/.claude/LIFEOS/ATLAS/Atlas.ts sync cloudflare
[cloudflare] routes lookup for zone <zone> failed: CF /zones/<id>/workers/routes -> 403
[cloudflare] KV namespaces failed: CF /accounts/<id>/storage/kv/namespaces -> 401
[cloudflare] R2 buckets failed: CF /accounts/<id>/r2/buckets -> 403
[cloudflare] D1 databases failed: CF /accounts/<id>/d1/database -> 401
[cloudflare] worker scripts failed: CF /accounts/<id>/workers/scripts -> 403
[cloudflare] worker custom domains failed: CF /accounts/<id>/workers/domains -> 403
✓ cloudflare: 22 assets, 20 edges (run N, PARTIAL)
```

Zones + DNS records (the calls the token *does* have permission for) now correctly land in the graph instead of being discarded, and the run is honestly marked `PARTIAL` rather than either crashing or silently claiming `complete`.

### DIFF
--- a/LifeOS/install/LIFEOS/ATLAS/collectors/Cloudflare.ts
+++ b/LifeOS/install/LIFEOS/ATLAS/collectors/Cloudflare.ts
@@ -52,6 +52,28 @@ async function cfOne(path: string, tok: string): Promise<any | null> {
   return body.success ? body.result : null;
 }
 
+/**
+ * Degrading wrapper around cf() for listing calls that can legitimately hit a
+ * token-scope gap (found 2026-08-03: a narrowly-scoped CLOUDFLARE_API_TOKEN
+ * threw on workers/routes 403, then storage/kv/namespaces 401, discarding
+ * everything already gathered each time). One missing permission on one
+ * resource type must degrade that signal only, never the whole collector —
+ * matches Atlas's own doctrine ("a partial page, an API error → PARTIAL run,
+ * no sweep, prior state intact"), which the bare cf() calls didn't actually
+ * implement for this failure class. Sets `partial` via the caller-owned flag
+ * so the run reports incomplete (never sweeps) instead of claiming success
+ * over data it couldn't fully see.
+ */
+async function cfSoft(path: string, tok: string, label: string, onFail: () => void): Promise<unknown[]> {
+  try {
+    return await cf(path, tok);
+  } catch (err) {
+    onFail();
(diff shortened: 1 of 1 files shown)


danielmiessler:
Thanks @schmetti-dev — having the Cloudflare collector degrade per-call instead of discarding the whole run. Ported into the source tree (the public repo is generated from it, so I'm closing this here), credited by number at the fix site. Ships in the next release.

PR #1745, opened 2026-08-03
fix(lifeos): print the current 'lifeos' command name instead of the old 'k'

elhoim:
## What

`install/LIFEOS/TOOLS/lifeos.ts` still prints and documents the launcher's old command name, `k`. Every install path wires the alias as `lifeos` (`install/install.sh:316`, `INSTALL.md:103`, `Workflows/Setup.md:56`), so the usage lines the CLI prints are not runnable as shown on a current install.

## Change (one file)

`install/LIFEOS/TOOLS/lifeos.ts` — renames `k` → `lifeos` in 15 places: the wallpaper usage/example lines, the update-available hint, the `mcp set` usage lines (2), the profiles header, the three `Examples:` lines in `mcp list`, the `mcp list | mcp set` and `prompt` error strings, the `cmdDoctor` docblock (3 mentions), the bun `--`-swallowing comment, and the `-r` session-ID comment.

Strings only — no behaviour change, no control flow touched. The three `Examples:` lines got their inline `#` comments re-aligned since the command grew five characters.

Deliberately **not** renamed: `.map(([k]) => k)` in `getMcpProfiles()` — that `k` is a destructured entry key, unrelated to the command name.

## Verification

- `bun build --target=bun --no-bundle install/LIFEOS/TOOLS/lifeos.ts` — parses clean.
- Ran the equivalent change against my own install and executed `lifeos mcp list`; output now reads `Profiles (use with 'lifeos mcp set')` and the three examples print runnable commands.
- `rg '\bk\b' install/LIFEOS/TOOLS/lifeos.ts` → only the destructured-variable line remains.
- Not run: a fresh-system install verification (step 3 of the contributing process).

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/lifeos.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/lifeos.ts
@@ -382,11 +382,11 @@ function setWallpaper(filename: string): boolean {
 }
 
 /**
- * `k doctor` — the capability check. Doctor.ts and the statusline both advise
+ * `lifeos doctor` — the capability check. Doctor.ts and the statusline both advise
  * running "lifeos doctor" on a capability regression, but there was no such
  * subcommand: the only working invocation was the more obscure
  * `bun LIFEOS/TOOLS/Doctor.ts`. Extra args are forwarded verbatim, so
- * `k doctor --network` and `k doctor decline <name>` work too.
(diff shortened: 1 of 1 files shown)


danielmiessler:
Thanks @elhoim — printing the current 'lifeos' command name. Ported into the source tree (the public repo is generated from it, so I'm closing this here), credited by number at the fix site. Ships in the next release.

PR #1748, opened 2026-08-03
feat(pulse): make every optional module switchable, and fix three flags that never worked

elhoim:
## The bug underneath this

`loadModules()` gates three modules on config that `loadPulseConfig()` never supplies:

```ts
if (config.local_intelligence?.enabled !== false) { ... }   // config.local_intelligence is always undefined
if (config.hypotheses?.enabled !== false) { ... }           // same
if (config.upgrades?.enabled !== false) { ... }             // same
```

`loadPulseConfig()` builds its return object key by key and copies through `voice`, `imessage`, `observability`, `performance`, `syslog`, `work`, `bunker`, `content`, `telos`, `hooks`, `da` — but not those three. They're declared on `PulseConfig`, so it typechecks. At runtime the gate reads `undefined`, `undefined !== false` is `true`, and the module loads every time.

Net effect: **`[local_intelligence] enabled = false` in PULSE.toml has never done anything.** Same for `[hypotheses]` and `[upgrades]`.

## The wider gap

Fourteen more user-facing modules had no switch at all — `wiki`, `memory`, `conduit`, `books`, `synapse`, `ledger`, `projects`, `assets`, `atlas`, `threatmodel`, `usage`, `evals`, `algorithm-tab`, `hermes` are unconditional `await import(...)`. And HEALTH / FINANCES / BUSINESS / GROWTH have no module of their own — `observability.ts` serves them at `/api/life/*` — so there was nothing to switch even in principle. GROWTH is the sharpest case: it's backed by an optional USER customization that isn't shipped, so out of the box its API returns `note: "Growth is an optional customization; not installed."` and the tab renders an empty page with no way to remove it.

## Change

- **`lib/modules.ts`** (new) — `MODULE_DEFAULTS` plus `resolveModules()`, layering defaults ← legacy `[section].enabled` flags ← the `[modules]` table. Pure and importable; `pulse.ts` calls `main()` at import time so nothing in it can be unit-tested.
- **`lib.ts`** — re-export.
- **`pulse.ts`** — `PulseConfig.modules`; `loadPulseConfig()` returns `resolveModules(parsed)` (this is the passthrough fix); 12 existing gates repointed at the map; 14 previously ungateable modules wrapped; `GET /api/config/modules`; a route guard for the four `/api/life/*` surfaces that have no module.
- **`PULSE.toml`** — documented `[modules]` table.

Infrastructure Pulse needs to serve anything at all — observability, hooks, tab-freshness, menubar, siri, doctor — is deliberately **not** switchable; a disabled one leaves a dashboard that can't render.

## Compatibility

Existing configs need no changes. Legacy `[section].enabled` flags still work and still win over the defaults. Shipped defaults leave everything on except the already-opt-in `imessage` and `syslog`.

## Verification

`resolveModules` against the shipped PULSE.toml and four synthetic configs:

| Case | Result |
|---|---|
| Shipped PULSE.toml | only `imessage`, `syslog` off — no behaviour change |
| `[modules]` local/growth = false | both false |
| Legacy `[local_intelligence].enabled = false` | `local` false — the switch that previously did nothing |
| Junk keys / wrong types (`telos = "yes"`) | ignored, default preserved |
| Empty config | deep-equals `MODULE_DEFAULTS` |

`bun build --target=bun --no-bundle` parses clean on `pulse.ts` and `lib/modules.ts`.

Ported to my own install (which predates the synapse/ledger/atlas/hermes split, so the key set differs) and ran it live with `local = false`: Pulse boots, `LocalIntelligence module loaded` is absent from the current boot's journal, `/api/local-intelligence/latest` → 404, `/api/config/modules` reports `local` off among 26 keys.

### DIFF
--- a/LifeOS/install/LIFEOS/PULSE/PULSE.toml
+++ b/LifeOS/install/LIFEOS/PULSE/PULSE.toml
@@ -10,6 +10,54 @@
 
 # ── Module Configuration ──
 
+# [modules] is the one place to switch a Pulse surface off. Every key defaults
+# to the value shown; delete a line and you get that default back. Disabling a
+# module stops it loading, stops its API routes answering, and (via
+# /api/config/modules) hides its tab in the dashboard instead of leaving a nav
+# entry that opens an empty page.
+#
+# Infrastructure Pulse needs to serve anything at all — observability, hooks,
+# tab-freshness, menubar, siri, doctor — is deliberately not listed: those are
+# how Pulse runs, not modules you use.
+#
+# The older `[section].enabled` flags below still work and still win over these
+# defaults, so an existing config needs no changes.
+[modules]
+# Life surfaces (the top nav)
+telos = true
+work = true
+content = true
+health = true
+finances = true
+business = true
+growth = true       # audience metrics; needs USER/CUSTOMIZATIONS/TOOLS/Growth.ts
+local = true        # LocalIntelligence civic digest; needs a Hometown set
+gear = true
+atlas = true
+memory = true
+synapse = true
+projects = true
+books = true
+# System surfaces
+algorithm = true
+bunker = true
+conduit = true
+upgrades = true
+hypotheses = true
+ledger = true
+usage = true
+performance = true
+evals = true
+threatmodel = true
+hermes = true
+docs = true
+da = true
+# Opt-in surfaces (off unless you turn them on)
+voice = true
+imessage = false
+syslog = false
+
+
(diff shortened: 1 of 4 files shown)


danielmiessler:
Thanks @elhoim — the Pulse module switches, which fix three flags that never worked. Ported into the source tree (the public repo is generated from it, so I'm closing this here), credited by number at the fix site. Ships in the next release.

PR #1749, opened 2026-08-03
feat(pulse): hide nav tabs whose module is switched off

elhoim:
> **Stacked on #1748.** This branch contains that commit as its parent; the diff below is the second commit only. Merge or port #1748 first — without `/api/config/modules` the hook fetch 404s and every tab stays visible (which is the designed fallback, so nothing breaks, but nothing hides either).

## Problem

Switching a module off stops its backend but leaves its tab in both nav rows, the mobile menu and the command palette. Clicking one opens a page with nothing answering it — an empty view rather than an honest absence.

GROWTH shows this without any config at all: it's backed by an optional USER customization that isn't shipped, so on a fresh install the tab is always there and always empty.

## Change

- **`nav-manifest.ts`** — `NavItem.module?: string`, set on 21 entries. Entries without one (Home, Agents, Skills, Hooks, Arbol, Security) are infrastructure and always render.
- **`lib/use-enabled-modules.ts`** (new) — fetches `/api/config/modules` once, returns a `useCallback`-stable predicate so callers can memoise on it.
- **`AppHeader.tsx`** — filters `tier1Nav` and `systemNav` at all four render sites (desktop tier-1, desktop system row, mobile sections, mobile system).
- **`CommandPalette.tsx`** — searches the filtered set, so `Cmd-K` can't jump to a page the nav is hiding.

## Fails open, deliberately

Until the fetch lands — and permanently if it errors, or the daemon predates the endpoint — every entry stays visible. A dashboard briefly showing a tab it then hides is a much smaller problem than a nav that empties itself because one request failed.

## Verification

Ported to my own install, built (`next build`, 40 static routes), restarted Pulse with `local = false`.

**Backend:** `/api/config/modules` reports `local` off; `/api/local-intelligence/latest` → 404; `LocalIntelligence module loaded` absent from the current boot's journal.

**Rendered nav** — the real `AppHeader` at 1440px, read out of the live DOM:

```
/memory/graph /telos /work /content /projects /health /finances
/business /growth /assets /knowledge /amber /agents /assistant
```

14 links. `/local` sits between `/growth` and `/assets` in the manifest and is absent from the render; the sequence jumps straight from `/growth` to `/assets`. Every other entry is untouched.

**Visually** — captured the desktop header in a real Chrome and zoomed in on the nav strip. Row 1 reads `TELOS WORK CONTENT PROJECTS HEALTH FINANCES BUSINESS`, row 2 reads `GROWTH ASSETS KNOWLEDGE AMBER`. No LOCAL.

`tsc --noEmit`: 26 pre-existing errors in unrelated files, none in the four files this touches.

### DIFF
--- a/LifeOS/install/LIFEOS/PULSE/Observability/src/components/AppHeader.tsx
+++ b/LifeOS/install/LIFEOS/PULSE/Observability/src/components/AppHeader.tsx
@@ -13,11 +13,17 @@ import { TabFreshnessPill } from "@/components/TabFreshnessPill";
 // meta view of the system working on itself, never part of the scrolling row.
 // SYSTEM is the mode-switch into the machine plane (lands on systemHome).
 import { tier1Nav, systemNav, metaNav, systemHome } from "@/lib/palette/nav-manifest";
+import { useEnabledModules } from "@/lib/use-enabled-modules";
 
 const systemPaths = [...systemNav.map((i) => i.href), "/system"];
 
 export default function AppHeader() {
   const pathname = usePathname();
+  // A tab whose module is switched off in PULSE.toml would open a page with no
+  // backend answering it, so drop it from both rows and the mobile menu.
+  const isEnabled = useEnabledModules();
+  const tier1 = tier1Nav.filter(isEnabled);
+  const system = systemNav.filter(isEnabled);
   const [mobileMenuOpen, setMobileMenuOpen] = useState(false);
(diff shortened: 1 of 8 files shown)


danielmiessler:
Thanks @elhoim — hiding nav tabs whose module is off. Ported into the source tree (the public repo is generated from it, so I'm closing this here), credited by number at the fix site. Ships in the next release.

Issue #1756, opened 2026-08-03
The Unified Event System has two writers and no reader, and the docs describe an emitter that has never shipped


anikinsasha:
## What I found

`MEMORY/STATE/events.jsonl` is documented as the Unified Event System. Two
shipped handlers write it:

- `hooks/handlers/MemoryDirIntegrity.ts` — path constant `:36`, append `:42`
- `hooks/handlers/KnowledgeConformance.ts` — path constant `:47`, append `:53`
  (new this release)

Programmatic readers: **zero**, in every release including this one. The write
side gained an emitter at 7.28.3; the read side is still empty.

Around that gap the documentation describes a system that does not exist.

**`appendEvent()` — the hook-facing emitter — has never shipped.**
`HookSystem.md:1555` says "Hooks call `appendEvent()` as a secondary write
alongside their existing state writes", `:1559-1560` shows the call, `:1566`
describes a field it auto-injects. No file under `hooks/` imports or defines it.
The one shipped function with that exact name is `LIFEOS/PULSE/Conduit/store.ts:20`,
which writes a per-day Conduit log; `LIFEOS/TOOLS/Conveyor/Ledger.ts:142` exports
`appendEvents` (plural) and writes `STATE/content-pipeline/events.jsonl`. Neither
is this file.

**`HookSystem.md:1550` `### Components` is a heading with an empty body** — two
blank lines, then the next heading.

**The event-type table names emitters that do not emit.**
`HookSystem.md:1575-1589` maps 13 event categories to 11 named hooks. Checked
against the payload:

- `RebuildSkill (DocRebuild handler)` (`:1584`) has **no file anywhere in the
  tree**. `git grep RebuildSkill` returns exactly one hit: that table row.
- `VoiceNotification` (`:1581`) is a handler with no registration of its own; it
  is reached through `VoiceCompletion.hook.ts`, and it writes
  `MEMORY/VOICE/voice-events.jsonl`.
- **None of the 11 writes `STATE/events.jsonl`.** The two handlers that actually
  write it appear nowhere in the table.

**Two documents contradict each other about the channel, and one contradicts
itself.**

- `HookSystem.md:1616` still says "`EventLogger.hook.ts` is now the single
  observability writer" — EventLogger writes `MEMORY/OBSERVABILITY/*.jsonl`, a
  different store — while `:1555`, in the same file, describes hooks calling
  `appendEvent()` into this one.
[...]

Issue #1596, opened 2026-07-24
hooks.json never registers PostToolObserver/DriftReminder, but HookSystem.md says they run — LoopDetector has no fire path on fresh installs


anikinsasha:
## The contradiction

`LifeOS/install/hooks/hooks.json` (the manifest `InstallHooks.ts` installs from) never registers three hooks that ship in `LifeOS/install/hooks/`:

- `PostToolObserver.hook.ts` — the consolidated sync catch-all PostToolUse dispatcher (its header: merges the two sync catch-all modules that emit additionalContext)
- `LoopDetector.hook.ts` — exact-repeat / oscillation / hammering detection; per its dispatcher's design it is ONLY reachable through PostToolObserver
- `DriftReminder.hook.ts` — voice/format drift nudges

But `LIFEOS/DOCUMENTATION/Hooks/HookSystem.md` describes them as live:

- Its post-consolidation summary lists `LoopDetector → PostToolObserver` among the dispatcher imports and counts PostToolObserver among the consolidation targets.
- Its PostToolUse settings listing shows `PostToolObserver.hook.ts` registered (catch-all, timeout 5).
- Its DriftReminder section says it is registered on UserPromptSubmit (user settings, async) + PostToolUseFailure (system settings).

Neither name has ever appeared in `hooks.json` history (`git log -S` comes back empty for both), and the shipped `settings.system.json` template carries no hooks block — so a fresh install gets no fire path for LoopDetector or DriftReminder at all. The consolidation's live registrations appear to have never been carried into the public install manifest.

Two smaller HookSystem.md nits in the same area:
- Its "8 files on disk but NOT registered directly" list omits `FormatGate.hook.ts`, which `StopGates.hook.ts` imports first in its gate chain (FormatGate's header describes it as OutputFormatGate's successor with teeth, 2026-07-11) — so the dispatched-file count is 9, not 8.
- The "30 registered / 38 on disk" counts don't reconcile against the public payload (28 distinct files registered by hooks.json; 39 hook files shipped counting `ContextReduction.hook.sh`).

## Possible resolutions (maintainer's call on intent)

1. Registration is intended → add `PostToolObserver` (PostToolUse catch-all) and `DriftReminder` to `hooks.json` so installs match HookSystem.md.
2. The lean manifest is intended → update HookSystem.md to mark these as shipped-but-optional, and note LoopDetector/DriftReminder are inactive by default.

Happy to PR either direction.

## Context

[...]

PR #1758, opened 2026-08-03
feat(hooks): give the advisory event log a reader

anikinsasha:
Implements #1756.

Closes the loop the Unified Event System describes but has never had a reader
for. Emission stays where it is; delivery moves to SessionStart, which is the
only end of that pair where context injection works.

## What changes

| File | Change |
|---|---|
| `hooks/lib/events.ts` | **NEW** — `appendEvent()` (the API `HookSystem.md:1555` documents), `emitFindingSet()`, and a backwards per-type-complete reader |
| `hooks/lib/advisory-readback.ts` | **NEW** — changed-set digest with a slow re-announce |
| `hooks/handlers/MemoryDirIntegrity.ts` | inline emit → the lib; findings gain stable keys |
| `hooks/handlers/DocCrossRefIntegrity.ts` | the summary emission its docstring has claimed since v5.0.0 — **emission only**, apply semantics untouched |
| `hooks/LoadContext.hook.ts` | fourth loader + a `notification-channel` guard at `main()` |
| `settings.system.json` (both copies) | `dynamicContext.advisoryReadback`, ships `true` |
| `LIFEOS/PULSE/Observability/observability.ts` | `STATE/events.jsonl` constant + merged into `/api/events/recent` |
| `HookSystem.md`, `MemorySystem.md`, `NotificationSystem.md`, `hooks/README.md` | doc repair |

## Three decisions worth your attention

**The emission contract is normative, not illustrative.** One event per
*completed* check carrying the full finding set including the empty set; no event
on a *skipped* check. Both directions matter: an emitter that speaks only when it
has findings can never announce that a problem was fixed, so the digest shows it
forever; an emitter that emits an empty set from a skip path claims a clean bill
of health it never checked. `DocCrossRefIntegrity`'s early return is commented for
exactly this reason.

**Finding keys carry no counts or timestamps.** The readback compares key *sets*.
A count-bearing key re-fires the digest every run; a timestamp-bearing one never
settles.

**The channel guard is at `main()`, not per loader.** Putting it inside the new
loader only would leave the existing loaders — including `activeWorkSummary`, the
one routinely non-empty on a real install — injecting the principal's context into
a remote turn. One check covers all of them and every future one. It sits
immediately after your `isSubagentContext()` check, which is the same class of
gate.

## What the doc repair covers

The event-type table at `HookSystem.md:1575-1589` mapped 13 categories to 11
named emitters. One of them, `RebuildSkill (DocRebuild handler)`, has no file
anywhere in the payload — `git grep` returns exactly that one table row. None of
the 11 writes `STATE/events.jsonl`, and the two handlers that do
(`MemoryDirIntegrity`, and `KnowledgeConformance` new this release) were not in
the table. It is now a table of what writes this log, with a note listing the
other observability streams that write their own files.

`### Components` at `:1550` was a heading followed by two blank lines; it now has
the component table it was going to have. `MemorySystem.md:647` asserted that
every hook in its integration table emits here; it now names the three that do.

## Wire-format note

`doc.integrity.memory_dir` events change shape: `drift` → `findings` (each with a
stable `key`), `drift_count` → `finding_count`. There are no programmatic readers
of this file in any release, so the only consumers are humans tailing it, but it
is a change and I would rather name it.

`timestamp` stays UTC via `Date#toISOString`, matching every event ever written to
this log. `lib/time.ts`'s `getISOTimestamp()` is local-with-offset; adopting it
would have silently changed the format, and it drags `identity.ts` and a YAML
parse onto a path inside every hook run.

## Shared files and rebase expectations

Every row below is the result of merging the two branches against this base, not
a reading of the diffs.

| File | Also touched by | Expectation |
|---|---|---|
| `hooks/handlers/MemoryDirIntegrity.ts` | PR-B | **Conflicts, and the resolution decides whether this PR still works. See below.** |
| `hooks/LoadContext.hook.ts` | PR-D, PR-E | **Conflicts with each of them.** The regions differ in intent but sit close together: this PR's loader and `main()` guard against D's edit inside the learning block and E's deletion of the relationship block. Small, and both sides are kept each time. D and E merge clean with each other. |
| `hooks/handlers/DocCrossRefIntegrity.ts` | PR-C | **Conflicts.** Both touch the SIDE EFFECTS docstring and the emit block — C's proposal event and this PR's typed emitter want the same lines. Small; take both sides. |
| `settings.system.json` ×2 | PR-C, PR-E | Merges clean in every combination, including with E's removal of a sibling key inside `dynamicContext`. |
| `hooks/README.md` | PR-C, PR-E | Different rows of the same tables; merges clean. |
| `LIFEOS/PULSE/Observability/observability.ts` | — | Sole toucher in this set. |
| `LIFEOS/DOCUMENTATION/Memory/MemorySystem.md` | PR-B | Merges clean — this PR edits the note at `:647` in the Hook Integration section, B rewrites the inventory table ~380 lines above it. |

**Resolving the PR-B conflict.** Two hunks, and neither is either/or. In the type
block, this PR adds the `DriftItem` docstring and `key` field while B adds the
status vocabulary sets: keep both. In the handler, B extracts the drift loops into
a pure `computeDrift()` and adds a third direction: keep B's structure, and carry
this PR's `key` fields into its pushes — `unknown_on_disk:<dir>`,
`missing_active:<NAME>`, `unrecognised_status:<NAME>`,
`inventory_unparseable:<class>`.

Taking B's side verbatim silently removes every key. Keys are the finding identity
the SessionStart readback compares to decide whether anything changed, so without
them the digest re-announces on every run or falls silent depending on ordering,
and the change-detection this PR is about stops working. Both test suites stay
green either way — B's does not know about keys, and mine tests the readback
rather than this handler — so a typechecker is what catches it, not `bun test`.

## Tests

`hooks/lib/events.test.ts` and `hooks/lib/advisory-readback.test.ts` — 30 cases,
`bun test`, green against this base. They cover the contract claims specifically:
an empty set is still emitted and flagged `ok`; a chatty emitter cannot evict a
quiet one from the reader's window; multi-byte characters survive a read-window
boundary; a torn trailing line is skipped; the digest is silent on an unchanged
set, re-announces after 20 sessions, never re-announces an empty one, and renders
one finding as exactly one line whatever the finding's text contains.

Two boundaries stated rather than left implied. The anti-eviction property holds
within the reader's `maxBytes` window and not outside it: on an append-only log a
quiet type whose last emission is further back than 8 MiB is not found, which
costs an absent advisory line, never a wrong one. And `detail` is flattened before
it is rendered, because `MemoryDirIntegrity` puts on-disk directory names in it
and the digest lands inside the SessionStart block.

This repository ships no tests today, so these are additive and entirely
separable — the fix stands without them if you would rather not carry them. They
are colocated `*.test.ts` because that is the convention `hooks/lib/work-events.ts:9`
already names.

## Deliberately not included

- No migration of the family off SessionEnd. The naming sweep landed this
  release and moved things the other way.
- No rotation for `events.jsonl`. The reader is bounded from the EOF end so growth
  costs it nothing, but the file still grows — asked as Q-b in the issue.
- No change to `DocCrossRefIntegrity`'s apply behavior. That is PR-C.
- No hook registration. #1596 is your in-flight commitment and no PR of mine
  touches `hooks.json`.

### DIFF
--- a/LifeOS/install/LIFEOS/DOCUMENTATION/Hooks/HookSystem.md
+++ b/LifeOS/install/LIFEOS/DOCUMENTATION/Hooks/HookSystem.md
@@ -34,7 +34,7 @@ The LifeOS hook system is an event-driven automation infrastructure built on the
 - **Security Validation** - Active (v5+, consolidated 2026-05-14) — Single `Safety.hook.ts` dispatching by `hook_event_name`: PermissionRequest gates outgoing tool calls via the shape classifier in `lib/safety-classifier.ts` (auto-allows safe shapes, neutral on dangerous/credential/injection); PostToolUse tags WebFetch/WebSearch responses with the "treat as data" warning + injection marker. Replaces the prior split between `SmartApprover.hook.ts` and `PromptInjection.hook.ts`. The v4.0 Inspector Pipeline was deleted 2026-05-06. See `LIFEOS/DOCUMENTATION/Security/README.md`.
 - **Multi-Agent Support** - Agent-specific hooks with voice routing
 - **Tab Titles** - Dynamic terminal tab updates with task context
-- **Unified Event Stream** - All hooks emit structured events to `events.jsonl` for real-time observability
+- **Unified Event Stream** - The advisory checkers emit structured events to `events.jsonl`, read back at SessionStart and by Pulse (see § Unified Event System for the emitters that participate)
 
 **Key Principle:** Most hooks run asynchronously and fail gracefully. Security hooks (e.g. `hooks/Safety.hook.ts`) are synchronous — the PermissionRequest path emits `decision: allow` JSON when safe (otherwise stdout is empty and the native engine prompts). All `.ts` hooks have `#!/usr/bin/env bun` shebangs and `+x` permissions — settings.json references them directly (e.g., `$HOME/.claude/hooks/Safety.hook.ts`) without a `bun` prefix. HTTP hooks (SkillGuard, AgentGuard) run via Pulse routes on `localhost:31337`.
 
@@ -1549,6 +1549,13 @@ Alongside existing filesystem state writes (algorithm-state JSON, ISAs, session-
 
 ### Components
 
+| Component | Path | Role |
+|-----------|------|------|
+| Emitter | `hooks/lib/events.ts` | `appendEvent()` / `emitFindingSet()` — the shared writer every emitter uses |
+| Log | `LIFEOS/MEMORY/STATE/events.jsonl` | Append-only JSONL, never rotated |
+| Model-facing reader | `hooks/lib/advisory-readback.ts` | SessionStart digest of the current finding set, via `LoadContext.hook.ts` |
+| Human-facing reader | `LIFEOS/PULSE/Observability/observability.ts` | Merged into `/api/events/recent`, rendered by the Pulse activity dashboard (`src/components/activity/ObservabilityDashboard.tsx`) |
+| Shell reader | — | `tail -f` / `jq` (see § Consuming Events) |
 
 ### Usage in Hooks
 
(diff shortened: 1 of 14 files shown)


danielmiessler:
Thanks @anikinsasha — giving the advisory event log a reader. Ported into the source tree (the public repo is generated from it, so I'm closing this here), credited by number at the fix site. Ships in the next release. While in there I also closed a remote-channel context-injection leak the change surfaced.

Issue #1728, opened 2026-08-02
Fresh install: ActivateImports returns ok:true with zero activations when the install page skips the CLAUDE.md/InstallSettings overlay step


rpriven:
*Related to #1724, which covers the same underlying fragility on the **update** path (Update.md step 3 has no re-overlay tool). This is the **fresh-install** sibling, via the AI-native install page, with a distinct sharp edge: the silent `ok:true`.*

The dangerous part first: **`ActivateImports` returns `ok: true` with an empty activation list.** A success code on zero work means an install can leave every identity import dead while reporting success — the assistant then runs on `{{PRINCIPAL_NAME}}`/`{{DA_NAME}}` placeholders and nothing signals a problem.

Verified in source: `ActivateImports.ts:59` hardcodes `ok: true` on the `--apply` path regardless of how many imports activated; `written` is the only zero-work signal. `InstallEngine.activateImports` (`:717`) returns an empty `{activated, skipped}` when `CLAUDE.md` is absent (no throw). So `--apply` with no `CLAUDE.md` present yields `ok: true` silently. **Note the asymmetry:** the *dry-run* path (`ActivateImports.ts:39-42`) *does* error on missing `CLAUDE.md` — only `--apply` is silent, which is backwards from what you'd want.

How it was reached on a fresh Linux user: on the AI-native path, `CLAUDE.md` was not yet in place when `ActivateImports --apply` ran, so it activated nothing and still reported success. (I can't quote the served `ourlifeos.ai/install` page's exact steps — it isn't in the release payload — so I'm not asserting the page's wording; what's certain is the fresh flow reached `ActivateImports` before `CLAUDE.md` existed.) The installing assistant caught it by cross-reading `Workflows/Setup.md`, which *is* correct (places `CLAUDE.template.md`→`CLAUDE.md` at `:30-31`, runs InstallSettings `:33`, then ActivateImports `:55`), placed the file, and re-ran — all imports then activated. So the Setup workflow is right; the gap is the fresh path reaching activation too early. (Consistent with #1724's part 2 — unsubstituted `{{PRINCIPAL_NAME}}` in functional strings — same placeholder-survives-silently mode from a different entry point.)

**Suggested fix (the first is the load-bearing one):**
1. Make `ActivateImports` fail loudly — or at minimum warn — when its import target doesn't exist. `ok: true` on zero activations is the trap regardless of how you got there.
[...]

PR #1771, opened 2026-08-04
feat(install): add OverlaySystem — the missing Update.md step-3 tool, so an update can refresh system-owned files

### LINKED ISSUE #1770: Update path cannot update any existing system-owned file: `copyMissing` is the only deploy mechanism, so a by-the-book 7.1.1→7.28.3 update leaves 31/50 hooks, 61/143 tools, 44/51 docs and the entire Algorithm at the old version

## TL;DR

`InstallEngine.copyMissing` writes a file only when the destination does not exist. It is the only mechanism any deploy tool uses. Consequently **the documented update path can add new files but can never update a changed one** — and `Workflows/Update.md` step 3, the step that would compensate, names no implementing tool.

Following the update path exactly on a 7.1.1 → 7.28.3 upgrade produced an install whose **constitution is 7.28.3 and whose machinery is 7.1.1**. Nothing errors. There is no warning. `LIFEOS/VERSION` can be bumped to 7.28.3 while almost none of the release actually landed.

This generalises #1724 (which reports the missing step-3 tool and the unwired hooks as separate symptoms). Both are the same root cause, and the blast radius is much wider than the constitutional files.

## Evidence

**1. The mechanism.** `Tools/InstallEngine.ts:364` — the file branch of `copyMissing`:

```ts
if (stat.isFile()) {
  if (!existsSync(d)) {
    mkdirSync(dirname(d), { recursive: true });
    cpSync(s, d);
    copied++;
  }
}
```

There is no overwrite path. `Tools/DeployCore.ts:119` deploys the runtime as `install/LIFEOS/<entry> → <configRoot>/LIFEOS/<entry>` through it, and its own header states the contract: *"NEVER overwrites a populated target."* Correct and desirable for a fresh install; fatal for an update.

**2. `DeployCore` is not in the update path at all.** `Workflows/Update.md` steps 3–7 cover system templates, `InstallHooks.ts --apply`, `ScaffoldUser.ts --apply`, `ActivateImports.ts --apply`, verify. Nothing deploys `install/LIFEOS/**`. On our run, after completing steps 3–7 as written, `LIFEOS/DOCUMENTATION/{Atlas,Ledger,Synapse,Hermes,Upgrades}` were **absent** and `LIFEOS/RULES/Verification.md` was **absent** — while the freshly-installed 7.28.3 system prompt references `RULES/Verification.md` by name in its On-Demand Rules Index. The constitution pointed at files that did not exist.

**3. Measured staleness after a correctly-followed update** (byte comparison of every payload file against its installed counterpart, `install/` vs `<configRoot>/`):

| Surface | Identical | **Still at 7.1.1** |
|---|---|---|
| `hooks/*.hook.ts` | 19 | **31 of 50** |
| `LIFEOS/TOOLS/*.ts` | 82 | **61 of 143** |
| `LIFEOS/DOCUMENTATION/**/*.md` | 7 | **44 of 51** |
| `LIFEOS/PULSE` (ts/tsx/json) | 163 | **64** |
| **Algorithm** | — | **`LATEST` = 8.4.0 vs payload 8.17.3** |

Stale hooks included `VerificationGate`, `Safety`, `SystemFileGuard`, `MemoryDeltaSurface`, `AlgorithmNudge`, `IntegrityCheck`, `WritingGate`, `FormatGate`, `StopGates`, `PreToolGuard`.

**4. Why this silently disables the release's headline features.** Several new hooks are composed inside dispatchers rather than registered standalone — by their own headers: `ISAGate`, `ISACloseGate`, `ISAFoldGate` are *"dispatched by / evaluated inside `StopGates.hook.ts`"*; `ISAStaleWriteGuard` is *"dispatched by `PreToolGuard`"*; `SystemChangeSurface` is *"composed inside `PostToolObserver`"* with `clearLedger` called from `MemoryTurnStart`.

The payload's dispatchers import them — `StopGates.hook.ts:32-34`, `PreToolGuard.hook.ts:50`, `PostToolObserver.hook.ts:24`, `MemoryTurnStart.hook.ts:34`. **The installed dispatchers import none of them**, because all four already existed and were skipped. So the ISA close teeth and the ⚙️ SYSTEM line cannot fire, while the system prompt that mandates them is fully installed.

**5. Residual, confirming #1724 part 1.** `install/hooks/hooks.json` ships **zero** entries for `TimeContext`, `VersionDrift`, `ModelRungGuard`, `KnowledgeWriteGuard`, `AtlasEventCapture`, `SpendAuditor`, so `InstallHooks.ts --apply` copies the files and reports them unwired with no manifest-sanctioned trigger to install against. (The other five from that issue's list are the composed ones above and need no manifest entry — only a current dispatcher.)

**6. Also confirming #1724 part 2, plus two new sites.** `hooks/ModelRungGuard.hook.ts:168` still has `{{PRINCIPAL_NAME}}` inside an `emit()` argument in 7.28.3. Two further unsubstituted placeholders appear in prompt text at `hooks/PromptProcessing.hook.ts:708` and `:721`.

## Impact

An operator who follows the documented update path gets a version marker that says 7.28.3 and an install that is mostly 7.1.1, with doctrine and machinery at different versions and no signal that anything is wrong. The Algorithm's runtime enforcement layer — the release's headline — does not arrive at all.

## Reproduction

1. Install 7.1.1 and populate it (any real use).
2. Run `bash install/install.sh`, then `Workflows/Update.md` steps 3–7 exactly as written.
3. Byte-compare `<skillRoot>/install/LIFEOS/**` against `<configRoot>/LIFEOS/**`, and `<skillRoot>/install/hooks/*.hook.ts` against `<configRoot>/hooks/*.hook.ts`.
4. Compare `<configRoot>/LIFEOS/ALGORITHM/LATEST` with `<skillRoot>/install/LIFEOS/ALGORITHM/LATEST`.

## Suggested fix

1. Add the re-overlay tool `Update.md` step 3 describes: overwrite **system-owned** paths from the payload — `hooks/`, `skills/`, `agents/`, `LIFEOS/{TOOLS,DOCUMENTATION,ALGORITHM,RULES,PULSE}`, plus `CLAUDE.md` and `LIFEOS_SYSTEM_PROMPT.md` — while excluding `USER/`, `LIFEOS/MEMORY/` and `settings.json`. A same-relative-path-only rule is sufficient and never deletes operator-authored files.
2. Put that tool into the `Update.md` sequence at step 3, and write `LIFEOS/VERSION` **last**, only on a fully successful apply — so a partial update can never present as complete.
3. Add `hooks.json` entries for the six standalone hooks in §5.
4. Add a release gate grepping for `{{...}}` outside comments (§6).

**A PR implementing (1) and (2) is open — linked below.** It adds `Tools/OverlaySystem.ts` in `DeployCore`'s style (dry-run by default, `--apply` to mutate, refuses a dev tree, JSON output, exit 0/1) and updates `Workflows/Update.md` step 3 to call it.

Validated by rolling a VM snapshot back to the pre-overlay state and re-running the whole update with the tool in place:

- dry-run planned 425 updates and left `VERSION` at 7.1.1
- `--apply` performed 425 updates + 1 create, then wrote `VERSION` 7.1.1 → 7.28.3
- a second `--apply` reported `updated=0, created=0, alreadyCurrent=1556` (idempotent)
- staleness re-measured to **0** on hooks, TOOLS and DOCUMENTATION; Algorithm `LATEST` 8.4.0 → 8.17.3
- `USER/` (282 files), `LIFEOS/MEMORY/` (57 KNOWLEDGE notes, 9 ISAs) and operator-authored hooks and skills all untouched
- full verification gate green afterwards: VERSION 7.28.3, 5/5 identity imports resolving, Pulse `/healthz` `/work` `/telos` all 200

One ordering note the tool made obvious and worth keeping in the docs: because it restores `CLAUDE.md` from the template, the identity imports return to their commented state, so `ActivateImports.ts --apply` must run after it — which is exactly where `Update.md` step 6 already sits.

## Environment

- Upgrade path: LifeOS 7.1.1 → 7.28.3 (release tarball via `ourlifeos.ai/install.sh`, sha256 `d938f099eed686ab240eb2e0d75b690a22e5ca786ca4e4fdbb5f4e787617f62a`)
- macOS 26.6 (25G72), Claude Code 2.1.221, bun 1.3.14
- Performed in an isolated Parallels VM cloned from a production install, never on the live system.

### How this was tested

Every finding above comes from a repeatable harness rather than a one-off attempt: the install is cloned into a snapshotted VM, the upgrade is rehearsed there, and a red gate rolls back in seconds. The evidence here was produced that way, including one deliberate rollback to re-run the entire update through the proposed fix from a clean pre-upgrade state.

We've written the whole method up rather than offer to explain it — **discussion #1769**. It also covers the two habits that made these findings trustworthy: claims close on probes rather than exit codes, and the assistant stops for a human decision before anything irreversible.

bnkath2o:

Implements `Workflows/Update.md` step 3, which describes a re-overlay of the system templates but has no tool behind it.

Fixes #1770 · methodology behind the validation: discussion #1769

## Why

`InstallEngine.copyMissing` writes a file only when the destination is absent (`InstallEngine.ts:364`), and it is the only mechanism the deploy tools use. That is the correct contract for a fresh install and means an update can add new files but never refresh a changed one.

Measured on a 7.1.1 → 7.28.3 update performed exactly as documented, before this change: **31 of 50 hooks, 61 of 143 tools, 44 of 51 docs and the whole Algorithm (`LATEST` 8.4.0 vs payload 8.17.3) were left at the old version**, while `LIFEOS/VERSION` could be bumped to the new one. Because `StopGates`, `PreToolGuard`, `PostToolObserver` and `MemoryTurnStart` were among the stale files, the new ISA close gates and the ⚙️ SYSTEM line could not fire even though the constitution mandating them installed fine. Full detail in #1770.

## What this adds

**`Tools/OverlaySystem.ts`** — follows `DeployCore`'s shape: dry-run by default, `--apply` to mutate, `--config-root` / `--skill-root`, refuses a dev tree with exit 2, JSON result, `process.exit(ok ? 0 : 1)`.

Scope is a fixed allowlist of system-owned paths:

```
hooks/ · skills/ · agents/ · LIFEOS/{TOOLS,DOCUMENTATION,ALGORITHM,RULES,PULSE}/
CLAUDE.md · LIFEOS/LIFEOS_SYSTEM_PROMPT.md
```

Never touched: `USER/`, `LIFEOS/MEMORY/`, `settings.json` (InstallHooks owns that merge), `node_modules`, `.git`, `out`, `.next`. Symlinks are skipped rather than followed or replaced, so the `USER` symlink contract is safe.

**Nothing is ever deleted.** A file is written only where the payload ships one at the same relative path, so operator-authored hooks, skills and tools survive untouched.

**`LIFEOS/VERSION` is written last**, and only when every copy succeeded — so a partial update cannot leave a marker claiming the new version.

**`Workflows/Update.md` step 3** updated to call it.

## Validation

Rolled a VM snapshot back to the pre-overlay state and re-ran the entire update with the tool in place:

| Check | Result |
|---|---|
| Dry run | planned 425 updates, `VERSION` left at 7.1.1 |
| `--apply` | 425 updated, 1 created, `VERSION` 7.1.1 → 7.28.3 |
| Second `--apply` | `updated=0, created=0, alreadyCurrent=1556` — idempotent |
| Staleness after | **0** across hooks, TOOLS, DOCUMENTATION |
| Algorithm | `LATEST` 8.4.0 → 8.17.3, spec file present |
| `USER/` | 282 files, unchanged |
| `LIFEOS/MEMORY/` | 57 KNOWLEDGE notes, 9 ISAs, unchanged |
| Operator-authored skills/hooks | present |
| Verification gate | green — VERSION 7.28.3, 5/5 identity imports resolve, Pulse `/healthz` `/work` `/telos` 200 |

## One ordering note

Because the tool restores `CLAUDE.md` from the template, the identity imports return to their commented state, so `ActivateImports.ts --apply` must run after it. That is already `Update.md` step 6, so the documented order is correct as written — worth stating explicitly since the failure would be silent, and `ActivateImports` returns `ok: true` even when it activates nothing (#1728). Verifying by reading `CLAUDE.md` back is the reliable check.

## Not included

No change to `copyMissing` or `DeployCore` — their additive contract is right for fresh installs and other callers depend on it. This is a separate tool for the update path.

Happy to adjust scope, naming, or the allowlist to taste.

### DIFF
--- a/LifeOS/Tools/OverlaySystem.ts
+++ b/LifeOS/Tools/OverlaySystem.ts
@@ -1,0 +1,247 @@
+#!/usr/bin/env bun
+/**
+ * OverlaySystem.ts — the update-path counterpart to DeployCore: refresh SYSTEM-OWNED files
+ * that already exist, which copyMissing by contract cannot do.
+ *
+ * DeployCore is additive: `copyMissing` writes only when the destination is absent, so a
+ * populated install can receive NEW files but never an UPDATED one. That is the right
+ * contract for a fresh install and the wrong one for an update — `Workflows/Update.md`
+ * step 3 ("Re-overlay system") describes this tool, which did not previously exist.
+ *
+ * Scope is deliberately narrow. Only paths LifeOS owns are overwritten:
+ *
+ *   hooks/ · skills/ · agents/ · LIFEOS/{TOOLS,DOCUMENTATION,ALGORITHM,RULES,PULSE}/
+ *   plus CLAUDE.md and LIFEOS/LIFEOS_SYSTEM_PROMPT.md
+ *
+ * Never touched: USER/ (the principal's tree), LIFEOS/MEMORY/ (per-install state),
+ * settings.json (hook registrations — InstallHooks owns that merge), node_modules, .git.
+ *
+ * Nothing is ever DELETED. A file is written only where the payload ships one at the same
+ * relative path, so operator-authored hooks, skills and tools survive untouched.
+ *
+ * VERSION is written LAST and only on a fully successful --apply, so a partial update can
+ * never present as complete (the failure mode where the marker says the new version while
+ * the machinery is still the old one).
+ *
+ * Dry-run by default (`--apply` to mutate); REFUSES the author's live source tree
+ * (`--allow-dev` to override; exit 2), matching DeployCore's failure contract.
+ *
+ * Usage:
+ *   bun OverlaySystem.ts [--config-root <dir>] [--skill-root <dir>] [--apply] [--allow-dev]
+ */
+
+import { existsSync, lstatSync, mkdirSync, readdirSync, readFileSync, writeFileSync, cpSync } from "node:fs";
+import { homedir } from "node:os";
+import { dirname, join } from "node:path";
+import { detectDevTree } from "./InstallEngine";
+
+/** Directories inside a system tree that are never overlaid, at any depth. */
+const SKIP_DIRS = new Set(["node_modules", ".git", "MEMORY", "USER", "out", ".next"]);
+
+/**
+ * System-owned trees: [payload-relative, configRoot-relative].
+ * Anything not listed here is left alone — absence from this list is the safety property.
+ */
+const SYSTEM_TREES: Array<[string, string]> = [
+  ["hooks", "hooks"],
+  ["skills", "skills"],
+  ["agents", "agents"],
+  ["LIFEOS/TOOLS", "LIFEOS/TOOLS"],
+  ["LIFEOS/DOCUMENTATION", "LIFEOS/DOCUMENTATION"],
+  ["LIFEOS/ALGORITHM", "LIFEOS/ALGORITHM"],
+  ["LIFEOS/RULES", "LIFEOS/RULES"],
+  ["LIFEOS/PULSE", "LIFEOS/PULSE"],
+];
+
+/** System-owned single files: [payload-relative, configRoot-relative]. */
+const SYSTEM_FILES: Array<[string, string]> = [
+  ["LIFEOS/LIFEOS_SYSTEM_PROMPT.md", "LIFEOS/LIFEOS_SYSTEM_PROMPT.md"],
+  ["skills/LifeOS/install/CLAUDE.template.md", "CLAUDE.md"],
+];
+
+function arg(a: string[], flag: string): string | undefined {
+  const i = a.indexOf(flag);
+  return i >= 0 && a[i + 1] && !a[i + 1].startsWith("--") ? a[i + 1] : undefined;
+}
+
+function sameBytes(a: string, b: string): boolean {
+  try {
+    const x = readFileSync(a);
+    const y = readFileSync(b);
+    return x.length === y.length && x.equals(y);
+  } catch {
+    return false;
+  }
+}
+
+interface OverlayResult {
+  what: string;
+  src: string;
+  dst: string;
+  present: boolean;
+  /** existed and differed → rewritten */
+  updated: number;
+  /** already byte-identical → untouched */
+  current: number;
+  /** absent → created */
+  created: number;
+  failures: string[];
+}
+
+/**
+ * Walk the payload tree and write each file to the matching relative path under dst.
+ * Updates and creations are both performed; deletions never are.
+ */
+function overlayTree(src: string, dst: string, what: string, apply: boolean): OverlayResult {
+  const r: OverlayResult = { what, src, dst, present: existsSync(src), updated: 0, current: 0, created: 0, failures: [] };
+  if (!r.present) return r;
+
+  const walk = (s: string, d: string): void => {
+    let entries: string[];
+    try {
+      entries = readdirSync(s);
+    } catch (err) {
+      r.failures.push(`readdir ${s}: ${String(err)}`);
+      return;
+    }
+    for (const name of entries) {
+      const sp = join(s, name);
+      const dp = join(d, name);
+      let st;
+      try {
+        st = lstatSync(sp);
+      } catch {
+        continue;
+      }
+      if (st.isSymbolicLink()) continue; // never follow or replace symlinks (USER is one)
+      if (st.isDirectory()) {
+        if (SKIP_DIRS.has(name)) continue;
+        walk(sp, dp);
+        continue;
+      }
+      if (!st.isFile()) continue;
+      if (existsSync(dp)) {
+        if (sameBytes(sp, dp)) {
+          r.current++;
+          continue;
+        }
+        if (apply) {
+          try {
+            cpSync(sp, dp);
+          } catch (err) {
+            r.failures.push(`copy ${sp}: ${String(err)}`);
+            continue;
+          }
+        }
+        r.updated++;
+      } else {
+        if (apply) {
+          try {
+            mkdirSync(dirname(dp), { recursive: true });
+            cpSync(sp, dp);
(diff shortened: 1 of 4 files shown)


danielmiessler:
Thanks @bnkath2o — and to Abe for the writeup. This is the fix for the single most-reported problem in the repo: `copyMissing` only writes files that are absent, so a documented update adds new files but silently leaves most of the machinery frozen at the old version while `VERSION` bumps. Your VM measurement of it (31/50 hooks, 61/143 tools, the whole Algorithm stale) is exactly the failure.

Ported into the source tree as `skills/LifeOS/Tools/OverlaySystem.ts`, credited to this PR, wired into `Update.md` step 3. One adaptation on the way in: before it overlays `CLAUDE.md` and the system prompt from the template, it now backs each up to a timestamped `.pre-overlay-*.bak` and surfaces the paths in its output — those two are the files an operator is most likely to have hand-edited, and restoring them from template would otherwise clobber local edits silently. Everything else is yours: dry-run default, `--apply`, the system-owned allowlist, never touching USER/MEMORY/settings, symlinks skipped, VERSION written last only on a clean apply. I verified it end-to-end on a scratch install (stale files refreshed, backups written with the hand-edits intact, USER/MEMORY untouched, idempotent second run).

One known gap I left for a follow-up: `out/` is skipped, so the prebuilt Pulse dashboard doesn't refresh on update — that needs a rebuild step, tracked separately.

Closing here since the public repo is generated from the source. Ships in the next release.

PR #1778, opened 2026-08-05
fix(pulse): fmtHero renders negative amounts as unabridged digits

takanorinishida:
`fmtHero()`'s magnitude thresholds (`>= 1_000_000`, `>= 10_000`, `>= 1_000`) are checked against the signed value, so any negative number fails every branch and falls through to the last one, which was written assuming `n` was already known-small. Net (Overall tab) commonly goes negative when spending exceeds income, so a K/M-abbreviated sibling KPI like "$2.1M" ends up next to "$-2,120,000" instead of "-$2.1M".

## Fix

Threshold on `Math.abs(n)` instead and prepend the sign once, before the currency symbol.

## Non-destructive

Positive inputs are byte-identical (sign is the empty string); this only changes output for values that were already rendering as long unformatted digit strings.

## Verification

Standalone checks: `fmtHero(2120000)` → `$2.1M` (unchanged), `fmtHero(-2120000)` → `-$2.1M` (was `$-2,120,000`), plus the K/thousands/zero boundary cases, all matching.

### DIFF
--- a/LifeOS/install/LIFEOS/PULSE/Observability/src/app/finances/page.tsx
+++ b/LifeOS/install/LIFEOS/PULSE/Observability/src/app/finances/page.tsx
@@ -185,13 +185,20 @@ interface FinancesDataV2 {
 
 function fmtHero(dollars: number | null | undefined): string {
   const n = Number(dollars) || 0;
-  if (n >= 1_000_000) return `$${(n / 1_000_000).toFixed(1)}M`;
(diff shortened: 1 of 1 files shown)


danielmiessler:
Thanks @takanorinishida — the fmtHero negative-amount abridging fix. Ported into the source tree (the public repo is generated from it, so I'm closing this here), credited by number at the fix site. Ships in the next release.

PR #1779, opened 2026-08-05
fix(pulse): Finances tab reads 0 income/expenses off LifeOS's own shipped templates

takanorinishida:
Two independent bugs in `/api/life/finances`, both hit by installs that follow the shipped `USER/TELOS/FINANCES` templates as-is (not specific to any one deployment's data):

- `parseCurrencyTable()` always read the amount from column 1. LifeOS's own `EXPENSES.md` ships as `| Category | Vendor | Amount | Frequency | Notes |` — column 1 is Vendor (never numeric), column 2 is Amount. Every install using the shipped layout silently got 0 for every expense row.
- `INCOME.md` ships as bulleted prose with no pipe-table at all (the code comment claiming it "leads with 'Annual Income Estimate'" doesn't match the shipped file), so `parseCurrencyTable(incomeRaw)` always returned `[]` and every install's income read as $0/¥0 — which then propagates into the Sankey, the 12-month trend, and the post-tax net.

## Fix

- `parseCurrencyTable()` now picks the amount column by header name ("amount", "annual", "monthly", "total", "cost", "value"), falling back to column 1 when no header matches so tables that already worked are unaffected. When a Frequency/Cadence/Period column is present, its value now annualizes a per-period amount via the existing `cadenceToMonthly()` instead of taking the cell at face value.
- Per-source income bullets use inconsistent ad hoc keys ("Gross per period", "Current MRR", "Per-engagement amount", ...), too heterogeneous to sum reliably, so this doesn't attempt that. Instead a new `parseCurrencyBullet()` reads the one number both the shipped template and hand-written files anchor on — the "## Expected Monthly Total" section's headline bullet — only as a fallback when the table is empty.

## Non-destructive

Both changes only add coverage for inputs that previously parsed to 0/`[]` — nothing that already extracted a nonzero number changes.

## Verification

Ran both parsers against the shipped `USER/TELOS/FINANCES/{INCOME,EXPENSES}.md` templates (still correctly yield nothing — placeholder `$X,XXX` text has no digits) and against a filled-in real install: income moved from ¥0 to the correct ¥4,800,000/year, expenses correctly read ¥1,524,000/year off the Amount column instead of the Vendor column.

### DIFF
--- a/LifeOS/install/LIFEOS/PULSE/Observability/observability.ts
+++ b/LifeOS/install/LIFEOS/PULSE/Observability/observability.ts
@@ -1501,16 +1501,42 @@ function computeFreshness(entries: Array<{
   return { dataDate: oldest, label, daysOld, tier, perFile }
 }
 
+// Cadence label (as it appears in a "Frequency" column) → the ObligationYaml
+// cadence values cadenceToMonthly() understands. Same mapping vendors.yaml's
+// legacy `frequency` field uses (see FREQUENCY_TO_CADENCE in handleLifeFinances) —
+// kept as a second copy here rather than shared, since this one keys off
+// free-text table cells (needs its own synonyms: "annually", "yearly", "one time")
+// while that one keys off a fixed YAML enum.
+const TABLE_FREQUENCY_TO_CADENCE: Record<string, "monthly" | "annual" | "quarterly" | "one_time"> = {
+  monthly: "monthly",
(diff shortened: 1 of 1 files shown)


danielmiessler:
Thanks @takanorinishida — header-aware column parsing on the Finances tab, which fixes the 0 income/expenses off the shipped templates. Ported into the source tree (the public repo is generated from it, so I'm closing this here), credited by number at the fix site. Ships in the next release.

Issue #1762, opened 2026-08-04
healthsync hardcodes America/Los_Angeles in 3 places: every reading near local midnight is filed under the wrong day for non-Pacific operators


jacobo-ortiz:
`LIFEOS/TOOLS/healthsync/` pins every day-boundary calculation to `America/Los_Angeles`:

| File | What it pins |
|---|---|
| `store.ts` (`dayKeyLA()`) | which day-file a reading lands in |
| `store.ts` (`isoNowLA()`) | the `fetched_at` / `lastSuccess` stamp |
| `eightsleep.ts` | the `tz=` query param sent to the Eight Sleep trends API |

For any operator outside Pacific time the day key is wrong near local midnight, so readings are silently filed under the previous or next day. Nothing errors. The data is just in the wrong bucket, and `HealthSync.ts` then builds trend windows off those mis-keyed files.

**Reproduced on a UTC-5 host on 2026-07-21:** an export at **01:13 local** was written to the **previous day** file, because 01:13 at UTC-5 is 23:13 the day before in Los Angeles. Two hours of every day land in the wrong file; for UTC+ operators the window is larger.

`eightsleep.ts` compounds it: the provider is asked to bucket by LA while the files are keyed by LA too, so the two agree only for LA users and the `from`/`to` window silently covers the wrong span for everyone else.

## Fix

A `resolveTimeZone()` in `store.ts` taking the first usable IANA zone from:

1. `LIFEOS_HEALTH_TZ` (explicit override for this subsystem)
2. `TZ` (the POSIX standard, which launchd/cron operators already set)
3. the host zone via `Intl.DateTimeFormat().resolvedOptions().timeZone`
4. `America/Los_Angeles` (the current value, so behaviour is unchanged when nothing else resolves)

Candidates validated by constructing an `Intl.DateTimeFormat` and falling through on throw, so a typo in `TZ` degrades instead of crashing the sync. Memoised, with a reset seam for tests.

## Compatibility: no call site changes

`dayKeyLA` and `isoNowLA` stay as `@deprecated` aliases of a renamed `dayKey` / `isoNow`. **There are 12 call sites** across `HealthSync.ts`, `apple.ts`, `oura.ts`, `function.ts` and `eightsleep.ts`; **none need to change.** They can be migrated whenever, or never.

[...]

PR #1783, opened 2026-08-06
fix(healthsync): resolve the operator's timezone instead of hardcoding America/Los_Angeles

jacobo-ortiz:
## Problem

`LIFEOS/TOOLS/healthsync/` pins every day-boundary calculation to `America/Los_Angeles` in three places:

| File | Line | What it pins |
|---|---|---|
| `store.ts` | 45 | `dayKeyLA()`: decides which day-file a reading lands in |
| `store.ts` | 67 | `isoNowLA()`: the `fetched_at` / `lastSuccess` stamp |
| `eightsleep.ts` | 200 | the `tz=` query param sent to Eight Sleep's trends API |

For any operator outside Pacific time the day key is wrong near local midnight, so readings are silently filed under the previous or next day. Nothing errors; the data is just in the wrong bucket, and `HealthSync.ts:253` then builds trend windows off those mis-keyed files.

Reproduced on a UTC-5 host on 2026-07-21: an export at **01:13 local** was written to `DATA/apple/2026-07-20.json`, because 01:13 UTC-5 is 23:13 the previous day in Los Angeles. Two hours of every day land in the wrong file; for UTC+ operators the window is larger.

`eightsleep.ts:200` compounds it: the provider is asked to bucket by LA while the files are keyed by LA too, so the two agree only for LA users, and the `from`/`to` window silently covers the wrong span for everyone else.

## Fix

Add `resolveTimeZone()` to `store.ts`, resolving in order and taking the first usable IANA zone:

1. `LIFEOS_HEALTH_TZ`: explicit override for this subsystem
2. `TZ`: the POSIX standard, which launchd/cron operators already set
3. the host zone via `Intl.DateTimeFormat().resolvedOptions().timeZone`
4. `America/Los_Angeles`: the previous hardcoded value, so behaviour is unchanged when nothing else resolves

Candidates are validated by constructing an `Intl.DateTimeFormat` and falling through on throw, so a typo'd `TZ` degrades instead of crashing the sync. The result is memoised with a `resetTimeZoneCache()` test seam.

The three hardcodes now call it. Error strings that said "LA day key" now name the resolved zone, which makes the next timezone bug self-describing.

## Compatibility

`dayKeyLA` and `isoNowLA` are kept as `@deprecated` aliases of the renamed `dayKey` / `isoNow`, so **no call site changes**. There are 12 call sites across `HealthSync.ts`, `apple.ts`, `oura.ts`, `function.ts` and `eightsleep.ts`; none were touched. Upstream can migrate them whenever, or never.

Default behaviour for an operator who sets nothing changes from "always LA" to "the machine's own zone", which is the intended semantics of a personal health log. Operators who genuinely want LA can set `TZ` or `LIFEOS_HEALTH_TZ` explicitly.

## Verification

Same instant (`2026-07-21T06:13:00Z`) through `dayKey()` under each zone:

| `LIFEOS_HEALTH_TZ` | day key |
|---|---|
| `America/Bogota` | `2026-07-21` |
| `America/Los_Angeles` | `2026-07-20` (old behaviour preserved) |
| `Europe/Madrid` | `2026-07-21` |
| `Asia/Tokyo` | `2026-07-21` |
| unset | `2026-07-21` (host zone) |
| `Zona/Inventada` | `2026-07-21` (invalid, falls through, no throw) |

`bun LIFEOS/TOOLS/HealthSync.ts pull` runs clean against all four providers after the change, and the fix has been running in production on a UTC-5 install since 2026-07-21.

## Files

- `LIFEOS/TOOLS/healthsync/store.ts`: +54/−6
- `LIFEOS/TOOLS/healthsync/eightsleep.ts`: +4/−1

## Related

Same class as the timezone work in PRs #762 / #780 / #806, which fixed `hooks/lib/time.ts` and `statusline-command.sh` but did not reach `LIFEOS/TOOLS/`. `Tools/FailureCapture.ts` and `Tools/RelationshipReflect.ts` still carry the same hardcode and are **not** covered by this PR.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/healthsync/eightsleep.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/healthsync/eightsleep.ts
@@ -16,6 +16,7 @@ import {
   isoNowLA,
   loadState,
   loadTokens,
+  resolveTimeZone,
   saveTokens,
   timedFetch,
   writeDayFile,
@@ -195,9 +196,11 @@ export async function pull(ctx: Ctx): Promise<SourceResult> {
 
   const to = dayKeyLA(ctx.now.getTime());
   const from = dayKeyLA(ctx.now.getTime() - WINDOW_DAYS * 24 * 60 * 60 * 1000);
+  // Ask Eight Sleep to bucket by the same zone the day keys use, or the window
+  // boundaries and the files they land in disagree for non-LA users.
   const url =
     `${CLIENT_API}/users/${auth.userId}/trends` +
-    `?tz=America/Los_Angeles&from=${from}&to=${to}` +
+    `?tz=${encodeURIComponent(resolveTimeZone())}&from=${from}&to=${to}` +
     `&include-main=false&include-all-sessions=true&model-version=v2`;
 
   let response: Response;
--- a/LifeOS/install/LIFEOS/TOOLS/healthsync/store.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/healthsync/store.ts
@@ -36,13 +36,58 @@ function stripQuotes(value: string): string {
   return trimmed;
 }
 
-export function dayKeyLA(epochMs: number): string {
+/** Last-resort zone, kept so behaviour is unchanged when nothing else resolves. */
+const FALLBACK_TIME_ZONE = "America/Los_Angeles";
+
+let cachedTimeZone: string | null = null;
+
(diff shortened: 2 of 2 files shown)


danielmiessler:
This landed from your own issue #1762 in yesterday's adoption pass — `resolveTimeZone()` with the same env-var order you proposed, credited to you at the fix site (store.ts header). Closing the PR as superseded by that port; nothing of yours was lost, it just arrived through the issue instead of the diff.

PR #1786, opened 2026-08-06
fix(ModelRungGuard): seek the transcript tail instead of reading the whole file

AntSentry:
`liveModel()` in `ModelRungGuard.hook.ts` computes a 256 KB tail offset and then `readFileSync`s the **entire** transcript to reach it — O(session length) work for a fixed-size read.

**Nobody is currently paying for this**, and I want to be upfront about that: the hook ships in `install/hooks/` but is registered in neither `settings.system.json` nor `settings.enhancements.json`, and the hooks README has no row for it. So this is a latent defect in shipped-but-unwired code, not a live regression. It bites whoever wires the hook up themselves — which is how I found it.

## The measurement

On a 152 MB transcript, each implementation run in its own process so the numbers don't share cache effects:

| | per call (warm, 5-run avg) | process RSS |
|---|---|---|
| current (`readFileSync`) | 22.1 ms | 183 MB |
| this patch (`openSync`/`readSync` seek) | **0.1 ms** | **30 MB** |

The hook runs on every `UserPromptSubmit`, so the cost scales with session length. In practice it doesn't look like a performance bug — it looks like an intermittent hook timeout that only appears late in long sessions, which is a considerably more annoying thing to debug.

## Behaviour preservation

I ran the old and new implementations side by side over **every transcript larger than 100 KB on one machine — 7763 files, 1.1 MB to 152 MB**:

```
differential: 7763/7763 agree, 0 disagree
```

That covers the null path (no assistant message yet), partial trailing lines at the tail boundary, and files smaller than the tail window.

## Scope

One function. The import list gains `openSync`/`readSync`/`closeSync`; `readFileSync` stays because `settingsPin()` still uses it. Nothing else in the file changes, and no other file is touched.

## One question for you, separate from the patch

**Is this hook meant to be registered?** It ships but nothing wires it, which is presumably why the defect has gone unnoticed. Two coherent answers and I have no stake in which:

- **Wire it** — then this fix matters before anyone turns it on, and the hooks README wants a `UserPromptSubmit` row.
- **Drop the file** — then close this PR; a shipped hook that nothing registers is dead weight either way, and that's a cleaner outcome than a fixed hook nobody runs.

Happy to follow up with either. If you'd rather not carry a fix for code you're about to delete, say so and I'll close it myself.

### DIFF
--- a/LifeOS/install/hooks/ModelRungGuard.hook.ts
+++ b/LifeOS/install/hooks/ModelRungGuard.hook.ts
@@ -35,7 +35,16 @@ for (const __k of ["LIFEOS_DIR", "LIFEOS_CONFIG_DIR", "PROJECTS_DIR"]) {
  * Failure mode: any error logs to stderr and exits 0, never blocking prompts.
  */
 
-import { appendFileSync, existsSync, mkdirSync, readFileSync, statSync } from "node:fs";
+import {
+  appendFileSync,
+  closeSync,
+  existsSync,
+  mkdirSync,
+  openSync,
+  readFileSync,
+  readSync,
+  statSync,
+} from "node:fs";
 import { join } from "node:path";
 
 const STDIN_TIMEOUT_MS = 300;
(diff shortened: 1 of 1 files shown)


danielmiessler:
Good fix, and your differential run over 7,763 transcripts is the kind of verification we wish every PR carried. The same seek-the-tail implementation landed in source a few hours before this PR via @elhoim's #1738 (credited at the fix site), so closing this one as already-landed rather than porting a second copy. The latent-defect framing (shipped-but-unwired) was accurate and appreciated — you're in the README credits for the report.

Issue #1766, opened 2026-08-04
ISACloseGate passes unitIsClaimable to .filter(), so its opts parameter receives the array index


xmasyx:
## Summary

`hooks/ISACloseGate.hook.ts:59` passes `unitIsClaimable` directly to `Array.prototype.filter`. `filter` invokes its callback with `(value, index, array)`, so the function's second parameter — declared as an options object — receives the **array index** at runtime.

Verified in `v7.28.3`:

```ts
// hooks/ISACloseGate.hook.ts:59
for (const u of splitIntoUnits(stripNoise(message)).filter(unitIsClaimable)) {

// hooks/VerificationGate.hook.ts:110
export function unitIsClaimable(u: string, opts?: { allowNarration?: boolean }): boolean {
```

## Impact

It does not currently throw: `opts` is a number, and `opts?.allowNarration` on a number is `undefined`, which happens to coincide with the intended default. So the behaviour is accidentally correct today — but the signature is being violated on every call, and the moment `unitIsClaimable` starts reading `opts` in any way that a number can satisfy (`typeof opts === "object"`, `Object.keys(opts)`, a truthiness check on `opts` itself), the gate changes behaviour silently.

It is also a hard type error, so the file cannot pass a strict typecheck:

```
error TS2769: No overload matches this call.
  Argument of type '(u: string, opts?: { allowNarration?: boolean | undefined; } | undefined) => boolean'
  is not assignable to parameter of type '(value: string, index: number, array: string[]) => unknown'.
    Types of parameters 'opts' and 'index' are incompatible.
      Type 'number' has no properties in common with type '{ allowNarration?: boolean | undefined; }'.
```

## Reproduction

```bash
bunx tsc --noEmit --strict   # over hooks/ISACloseGate.hook.ts
```

## Proposed fix

Wrap the callback so only the value is forwarded, which also makes the intended `opts` default explicit at the call site:

```diff
-  for (const u of splitIntoUnits(stripNoise(message)).filter(unitIsClaimable)) {
+  for (const u of splitIntoUnits(stripNoise(message)).filter((u) => unitIsClaimable(u))) {
```

Worth a grep for other bare `.filter(fn)` / `.map(fn)` call sites where `fn` takes an optional second parameter — this failure mode is invisible until the second parameter is actually read.

danielmiessler:
[...]

Issue #1765, opened 2026-08-04
hooksHealth() returns a flattened object that does not match its declared { status, stats } type


xmasyx:
## Summary

`hooksHealth()` in `LIFEOS/PULSE/modules/hooks.ts` declares a return type of `{ status: string; stats: HookStats }` but returns `{ status: "ok", ...stats }`. The spread flattens `HookStats` into the top level, so the returned object has **no `stats` key at all**. The declared type is a lie, and any consumer reading `.stats` gets `undefined`.

Verified in `v7.28.3` at `LifeOS/install/LIFEOS/PULSE/modules/hooks.ts:141-143`.

```ts
interface HookStats {
  requests: number
  skillGuard: { total: number; blocked: number; passed: number }
  agentGuard: { total: number; warned: number; passed: number }
}

export function hooksHealth(): { status: string; stats: HookStats } {
  return { status: "ok", ...stats }   // <- flattened, no `stats` key
}
```

## Impact

`pulse.ts:493` assigns the result straight into the health payload:

```ts
subsystems.hooks = hooksHealth()
```

So the health endpoint serves `{ status, requests, skillGuard, agentGuard }` instead of the documented `{ status, stats }`. Anything reading `subsystems.hooks.stats.requests` silently reads `undefined` — no throw, no log, just a metric that is always empty.

## Reproduction

```bash
bunx tsc --noEmit --strict   # over the hooks module
```

Minimal repro:

```ts
interface HookStats { requests: number }
const stats: HookStats = { requests: 0 }
export function hooksHealth(): { status: string; stats: HookStats } {
  return { status: "ok", ...stats }
}
```

```
error TS2741: Property 'stats' is missing in type
'{ requests: number; status: string; }' but required in type
'{ status: string; stats: HookStats; }'.
```

## Proposed fix

Nest the stats instead of spreading them, which matches the declared type and the shape the health consumers expect:

```diff
 export function hooksHealth(): { status: string; stats: HookStats } {
-  return { status: "ok", ...stats }
+  return { status: "ok", stats: { ...stats } }
 }
```

(The inner spread keeps the returned snapshot decoupled from the live mutable `stats` object, so a later `stats.requests++` cannot retroactively change a payload already handed out.)

If the flattened shape is the intended wire format instead, the fix is the mirror image — change the signature to `HookStats & { status: string }` — but then `pulse.ts` should be checked for readers of `.stats`.

danielmiessler:
[...]

Issue #1614, opened 2026-07-25
Offering a few small tools back: typecheck step, document-fidelity gate, PDF/docx renderers, archive purge


xmasyx:
## Offering some tools back, if any are wanted

Running LifeOS daily produced a handful of small tools that fill gaps in the shipped set. Rather than open a pile of unsolicited PRs, here they are as a list — happy to send a PR for any that look useful, one at a time, and equally happy to hear that a given one belongs in a user's own install rather than in the project.

Ordered roughly by how general the value is.

**1. A real typecheck step.** `bun build` does not typecheck. A field that is read but never declared bundles cleanly, so it "compiles" — and stays broken for days. This runs `tsc --noEmit` over two roots (the root config, and the Pulse server, which needs its own config because its dependencies differ and the file cannot be named `tsconfig.json` without Bun picking it up). Roughly 60 lines.

**2. A document-fidelity checker.** For any pipeline that renders markdown to PDF: extract the text back out with an independent extractor (poppler, so the verifier does not share the renderer's assumptions) and diff it against the source, failing if a word or number was dropped. A PDF that opens is not a PDF that is correct — the first run of ours found a command silently deleted by the renderer. This also caught a defect *in itself*: the vocabulary rule required tokens of pure letters and digits, so every word followed by punctuation was never checked. On a real document it was watching 400 tokens out of 508.

**3. A markdown → PDF renderer for decision documents.** Pure JS (pdfmake 0.2.x — 0.3 removed the server-side stream), base-14 fonts, no browser, no native binary, native image embedding. Exists because decisions read in a terminal are reflexes; the same document on a phone or on paper is a decision. Pairs with (2), which gates it.

**4. A markdown → .docx builder.** Builds OOXML directly rather than converting, so no word processor is in the path. The gate that matters: after writing, open the package and assert images are **embedded**, not linked — a linked image looks identical until the recipient opens the file somewhere else and sees an empty rectangle.

**5. An archive-purge tool with a grace window.** Deletes files older than N days **only** inside a directory whose basename matches exactly, refuses any other path, spares files marked with a `KEEP-` prefix, and touches nothing under `--dry-run`. Small, but the fail-closed directory check is what makes it safe to point at anything.

**6. A local-delta verifier.** Described separately in a discussion — a registry of local changes plus behavioral probes, so an upgrade can report what it reverted instead of reverting it silently.

[...]

PR #1793, opened 2026-08-07
feat(tools): add a real typecheck step (bun build does not typecheck)

xmasyx:
Offered in #1614; you said yes to (1).

`bun build` bundles without typechecking, so a field that is read but never declared compiles cleanly and stays broken silently.

**What it does.** `bun LIFEOS/TOOLS/Typecheck.ts` runs `tsc --noEmit` over two roots: `hooks/` + `LIFEOS/TOOLS/`, and `LIFEOS/PULSE/**`. Two passes rather than one wider `include`, because Pulse resolves against its own `node_modules`.

**The configs are generated and deleted.** No file named `tsconfig.json` ever lands at the install root. Bun reads that name at runtime, and one `paths`/`baseUrl`/`jsx` field in it would move the behaviour of every live hook and of the running Pulse server. The generated configs use a non-standard name and are removed in a `finally`, so runtime invariance is true by construction rather than by discipline. It also means no installer change: one file, nothing to deploy.

**Exclusions are printed loudly.** `PULSE/Observability/**` is a standalone Next.js app with its own deps and deserves its own pass; every run says so. Excluded does not mean healthy, it means not looked at. A missing Pulse dir prints `skipped` and counts as neither pass nor fail.

**Verified both ways:** exit 1 against this tree, exit 0 against a clean one, generated configs removed in both cases.

## What it finds on current `main`

Reporting rather than fixing, to keep this PR to one file. Happy to send any of them separately.

Two of the four are ones you have already closed at my report, which is the useful part: the step re-derives them from the code alone, with no issue tracker involved.

Already reported and already fixed in your source — listing them because they are still in the published payload, and because they are what the step catches on a cold checkout:

- `hooks/ISACloseGate.hook.ts:59` — `.filter(unitIsClaimable)` hands the array index to the optional second parameter. Filed as #1766, closed 6 Aug.
- `LIFEOS/PULSE/modules/hooks.ts:142` — the returned object is missing `stats`, which its own declared type requires. Filed as #1765, closed 6 Aug.

Not previously reported:

- `LIFEOS/TOOLS/SecretScan.ts:228` — `error.message` on `unknown`; throws on a non-Error throw.
- `LIFEOS/TOOLS/ExtractTranscript.ts` and `SplitAndTranscribe.ts` both `import OpenAI from "openai"` at top level, and `openai` is declared in no package.json reachable from `LIFEOS/TOOLS/` (only `skills/Art/Tools/`). Those two fail at runtime on a fresh install, not just here.

Deliberate, and worth a decision rather than a silencer:

- `LearningPatternSynthesis.ts:632` → `../../test/lib/hook-replay`, and `pulse.ts:146` → `./Assistant/module`. Both are optional imports guarded by `existsSync` with a comment saying so (#1419). Correct code that simply cannot resolve in the public payload. If you want this green as a gate, those two need ambient declarations or an explicit allowlist. I left the call to you rather than shipping a suppression mechanism.


danielmiessler:
Ported: Typecheck.ts ships in v7.40.4. You were right that `bun build` doesn't typecheck; the tool caught eight real errors tree-wide on adoption. It runs standalone today, and wiring it into the release gates is queued. Closing rather than merging: this repo is generated from private source at release, so a merge here would be overwritten.

PR #1794, opened 2026-08-07
feat(tools): add PurgeArchives, an age-based purge with a fail-closed directory guard

xmasyx:
Offered in #1614; you said yes to (5).

Deletes files older than N days (default 14) inside an archive directory. A manual archive is a grace window, not forever: silence means it goes after N days, a `KEEP-` prefix means it stays. The trigger is lazy, called right after archiving something by hand, so the directory prunes itself as a side effect of being used and there is no scheduler.

**The guard is the reason to take it.** It refuses any directory whose basename is not exactly the sentinel (`--dir-name`, default `_archive`). A tool that deletes by age is one bad argument away from deleting live work, so the check is not "this path looks plausible": the directory has to be *named* as an archive, which is a property no working directory acquires by accident. Files only, never subdirectories; dotfiles skipped; every action logged; `--dry-run` touches nothing.

```
bun LIFEOS/TOOLS/PurgeArchives.ts <dir> [--days N] [--dir-name NAME] [--dry-run]
```

**Verified** against a fixture tree, case by case: pointed at a live directory it refuses with exit 1 and leaves the contents intact; `--dry-run` prints and deletes nothing; the real run removes only the expired non-`KEEP-` file while `KEEP-`, fresh files, dotfiles and subdirectories all survive; the sentinel override works; `--days abc` is rejected with exit 1.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/PurgeArchives.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/PurgeArchives.ts
@@ -1,0 +1,119 @@
+#!/usr/bin/env bun
+/**
+ * PurgeArchives.ts — delete expired archived versions inside an archive directory.
+ *
+ * POLICY. A manual archive is a grace window, not forever. A file older than N days
+ * (default 14) is deleted, UNLESS its name starts with `KEEP-`, which is an explicit
+ * "hold this" signal. Silence means it goes after N days; `KEEP-` means it stays.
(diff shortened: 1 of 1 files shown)


danielmiessler:
Ported: PurgeArchives.ts ships in v7.40.4, fail-closed as you built it, with a symlink-safe realpath guard added during a later audit. Closing rather than merging: this repo is generated from private source at release, so a merge here would be overwritten.

PR #1798, opened 2026-08-08
Add SuggestSkills: decide which skills to build from your work history

rathko:
## What

Adds SuggestSkills, a read-only skill that tells you which new skills are worth building, from your own work history and satisfaction signals. It proposes a ranked shortlist and routes accepted items to CreateSkill. It never creates or edits a skill itself.

## Why

CreateSkill covers building, validating, testing, and improving a skill. Nothing covers the step before it: deciding what to build. A plain topic scan of your sessions gives a false "nothing here", because the real gaps hide in two places a topic scan cannot see. First, frustration: a topic can look covered while you keep hitting the same wall inside it. Second, discipline gaps under a covered topic: a build skill owns "app work" while the recurring pain is an unowned discipline the build skill never touches. SuggestSkills reads the frustration signal LifeOS already captures via SatisfactionCapture, alongside work sessions, so it catches both.

## Shape

Read-only and proposal-only. It has no capability to create a skill, which is what makes "never auto-create" a boundary rather than a promise in prose.

Deterministic gathering. Tools/CollectSignals.ts emits a normalized corpus (sessions, low-rating frustrations, the skill/loop/workflow registry, and warnings for any missing or malformed store). The model only clusters and judges what the tool returns, so two runs see the same evidence. Store paths are discovered via flag, env, or root, so it works across installs.

Two verification passes, reported as a union with an agreement tag. Requiring both passes to agree would hide the subtle gaps this exists to find.

## Evidence

Blind test on real history: a topic-only pass missed a recurring state-modeling gap and reported nothing. The frustration-aware pass, run over the same window with the answer hidden, surfaced it in both passes independently. It also files behavior corrections such as verbosity or scope misreads as steering, not skills.

## Contents

SKILL.md, Workflows/Scan.md, Tools/CollectSignals.ts. Generic, no external dependencies.

## Credit

The idea comes from Hermes Agent by Nous Research, whose learning loop creates skills from experience. Hermes creates and grooms them autonomously. This takes the discovery half and inverts the risk: it is read-only, proposes only, and a human approves every skill before CreateSkill builds it.

### DIFF
--- a/LifeOS/install/skills/SuggestSkills/SKILL.md
+++ b/LifeOS/install/skills/SuggestSkills/SKILL.md
@@ -1,0 +1,35 @@
+---
+name: SuggestSkills
+description: "Discover WHICH new skills you should create, from your own work history plus your satisfaction/frustration signals. Read-only and proposal-only: it surfaces recurring pain that no existing skill, loop, or workflow covers, then hands you a ranked shortlist to build with CreateSkill. It never creates or edits a skill itself. Frustration is a first-class signal (a topic can look 'covered' while you keep hitting the same wall inside it), so it reads low ratings and recurrence markers, not just session topics. USE WHEN should I create a skill, what skills do I need, suggest skills, skill gap, based on my recent work, am I missing a skill, what should I build. NOT FOR creating/validating/testing/optimizing an individual skill (use CreateSkill) — this only decides WHAT to build, not how."
+---
+
+# SuggestSkills — what should I build next?
+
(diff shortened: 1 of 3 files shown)


danielmiessler:
Ported via CreateSkill: SuggestSkills ships in v7.40.4 with your MIT footer preserved verbatim and CollectSignals proven read-only. Nothing in the system did this before. Closing rather than merging: this repo is generated from private source at release, so a merge here would be overwritten.

Issue #1730, opened 2026-08-02
Windows: setupUserSeparation() symlinkSync fails with EPERM; the copy fallback promised in INSTALL.md never runs (directory junctions work unelevated)


umair-a11y:
## Summary

`setupUserSeparation()` in `Tools/InstallEngine.ts` creates the system/user separation link with plain `symlinkSync(dataUserDir, liveUserDir)` (lines 604 and 614, no `type` argument). On Windows, creating a real symlink requires elevation or Developer Mode, so on a stock machine `symlinkSync` throws `EPERM`; the engine reports `symlink creation failed: EPERM ...`, the separation link is never established, and `checkSymlinkContract` then fails with `... is not a symlink (system/user separation broken)`.

INSTALL.md line 163 promises otherwise:

> `| **Claude Code — Windows** | ✅ (copy fallback where symlinks need admin) | ✅ full |`

No copy fallback exists in the v7.28.3 code path. Code and docs currently disagree about what happens on Windows.

## Environment

| | |
|---|---|
| LifeOS | v7.28.3 (release tarball) |
| OS | Windows 11 Home 10.0.26200 |
| Harness | Claude Code v2.1.220 |
| bun | 1.3.14 |
| git | 2.51.0.windows.1 |

## Repro

1. Windows 11, Developer Mode off, non-elevated PowerShell.
2. Run the install through the user-separation step (with `HOME` set manually to get past the home-resolution bug, #1729).
3. `symlinkSync` fails with `EPERM: operation not permitted, symlink ...`. No fallback runs, and the symlink contract check fails.

## Suggested fix: directory junctions

```ts
symlinkSync(dataUserDir, liveUserDir, "junction");
```

Why this works:

- Directory junctions need no elevation and no Developer Mode.
- Node and Bun ignore the `type` argument on non-Windows platforms, so no platform branch is needed.
- `dataUserDir` is already absolute, which junctions require.
- The existing contract survives unchanged: `lstatSync().isSymbolicLink()` returns true for junctions and `readlinkSync()` returns the target, which is all `checkSymlinkContract()` (`Tools/InstallEngine.ts:625-639`) relies on.

Verified on this install: after creating the junction manually,

```powershell
New-Item -ItemType Junction -Path "$env:USERPROFILE\.claude\LIFEOS\USER" -Target "$env:USERPROFILE\.config\LIFEOS\USER"
```

`checkSymlinkContract` passed as written, and the installed system has been running against the junction since.

[...]

Issue #1119, opened 2026-05-01
PRDSync hook silently fails on Windows — backslash file_path breaks MEMORY/WORK substring check


Stevenloo911:
## Bug

`PRDSync.hook.ts` silently fails on Windows installs because the `MEMORY/WORK/` substring check uses forward slashes, but Claude Code on Windows passes `tool_input.file_path` with native backslash separators (`C:\Users\…\MEMORY\WORK\…`). The check returns false, the hook returns early, and `work.json` is never updated. The Activity Dashboard / SessionProgress CLI then show stale data indefinitely.

## Repro

Windows 11, Git Bash, Bun 1.3.9, PAI v4.0.3 (also present on `main` as of `9fa02cb00`).

```bash
# This is approximately what Claude Code spawns on Windows:
echo '{"tool_input":{"file_path":"C:\\Users\\me\\.claude\\MEMORY\\WORK\\20260501-120000_test\\PRD.md"}}' \
  | bun ~/.claude/hooks/PRDSync.hook.ts
```

Result: exit 0, **no `{"continue":true}` output**, `work.json` mtime unchanged. Same input with forward slashes runs the full hook and updates `work.json` correctly.

## Root cause

[`hooks/PRDSync.hook.ts` line 35](https://github.com/danielmiessler/Personal_AI_Infrastructure/blob/main/Releases/v5.0.0/.claude/hooks/PRDSync.hook.ts):

```typescript
const filePath = toolInput.file_path || '';
if (!filePath.includes('MEMORY/WORK/') || !filePath.endsWith('PRD.md')) return;
```

`filePath.includes('MEMORY/WORK/')` is false when `filePath` is `C:\Users\…\MEMORY\WORK\…\PRD.md`.

## Fix (one line, separator-normalize before the check)

```diff
-  const filePath = toolInput.file_path || '';
-  if (!filePath.includes('MEMORY/WORK/') || !filePath.endsWith('PRD.md')) return;
+  // Normalize separators — Claude Code on Windows passes native backslash paths
+  const filePath = (toolInput.file_path || '').replace(/\\/g, '/');
+  if (!filePath.includes('MEMORY/WORK/') || !filePath.endsWith('PRD.md')) return;
```

After applying locally on v4.0.3: backslash test now prints `{"continue":true}`, `work.json` is written, and previously-stale sessions (registered manually for one) are picked up on the next PRD edit.

## Impact

[...]

PR #1800, opened 2026-08-09
IntegrityCheck: tolerate UTF-8 BOM and CRLF in frontmatter parsing (#1732)

### LINKED ISSUE #1732: v7.28.3: IntegrityCheck opens red on a stock install: 239-ref payload debt (#1541 regrown), unshipped replay corpus, BOM/CRLF-blind frontmatter probes
## Summary

On a fresh v7.28.3 Windows install, `LIFEOS/TOOLS/IntegrityCheck.ts` reports 281 blocking findings. After separating out local/derived-config classes, three vendor-side classes remain that a stock install can never clear. This is the same shipped-payload debt class as the closed #1541 (223 missing refs at the time) and #1504; the count has grown to 244.

## Environment

| | |
|---|---|
| LifeOS | v7.28.3 (release tarball) |
| OS | Windows 11 Home 10.0.26200 |
| Harness | Claude Code v2.1.220 |
| bun | 1.3.14 |

## Finding 1: 244 missing references in the shipped payload

`references` check: 2648 refs across 1619 files, 244 missing. Top clusters:

| Referencing file | Missing refs |
|---|---|
| `LIFEOS/ALGORITHM/changelog.md` | 42 |
| `LIFEOS/DOCUMENTATION/LifeosSystemArchitecture.md` | 16 |
| `LIFEOS/DOCUMENTATION/Work/WorkSystem.md` | 14 |
| `LIFEOS/DOCUMENTATION/Testing/TestingDoctrine.md` | 8 |
| `LIFEOS/DOCUMENTATION/Observability/ObservabilitySystem.md` | 8 |
| `LIFEOS/DOCUMENTATION/Memory/MemorySystem.md` | 8 |
| `skills/Remotion/Workflows/GeneratedContentVideo.md` | 8 |
| `LIFEOS/DOCUMENTATION/Pulse/PulseSystem.md` | 7 |
| `LIFEOS/DOCUMENTATION/Ledger/LedgerSystem.md` | 7 |
| `LIFEOS/DOCUMENTATION/Tools/Cli.md` | 6 |

Docs also reference hooks that do not ship, e.g. `hooks/ULWorkSync.hook.ts` (6 refs). Additionally, the checker scans runtime artifacts under `LIFEOS/MEMORY/` (e.g. `OBSERVABILITY/reviewer-runs/<ts>/prompt.user.md` picked up 7 "missing refs" from generated content); excluding runtime dirs from the reference scan would keep the signal clean.

## Finding 2: replay-corpus check demands a corpus the tarball does not ship

The `replay-corpus` check runs `bun test <configRoot>/test/regression`. The public release contains no `test/regression`, so the check fails on every stock install with `Test filter ... had no matches`. Suggested: skip-with-note when the corpus directory is absent (the `retirement-registry` and `gate_counts` checks already model this: "skipped — release tooling not installed").

## Finding 3: frontmatter probes are CRLF- and BOM-intolerant (Windows false positives)

The frontmatter checks match `/^---\n([\s\S]*?)\n---/` against raw file bytes (`IntegrityCheck.ts:658` for commands/agents, `:742` for skills). Two byte-level realities of Windows-authored files defeat the anchor:

- CRLF line endings: the file starts `---\r\n`, the regex fails (reproduced with a Windows-authored skill; converting to LF cleared the finding with zero content change).
- UTF-8 BOM: `EF BB BF` before the opening `---` breaks the `^` anchor even with LF endings (reproduced with two command files; stripping the BOM cleared them).

Both report as `no frontmatter block` on files whose frontmatter is complete. Suggested fix: strip a leading BOM and use `/^---\r?\n([\s\S]*?)\r?\n---/` (or normalize before matching). Same Windows-probe family as the detection notes in #1730.

## Additional notes

- `hook_registration` counts only directly-registered hooks, while `hook_wiring` credits dispatcher-invoked hooks; dispatcher-called hooks therefore inflate the registered-vs-on-disk delta on installs that are actually correctly wired.
- `CarrierProbe.ts` on Windows always returns INCONCLUSIVE with `sidechain: []` (it does not find the subagent transcript in the session dir), so the carrier fact can never be established on this platform.

## Impact

A fresh install's integrity verdict opens at ~250+ blocking findings the user cannot fix, which buries the findings they can. Once local classes were repaired here, the vendor classes above are the entire remaining verdict.

## Workarounds used locally

None available for Findings 1-2 (vendor payload). Finding 3: converted affected files to LF.

PotatoFarmerBoat:
Fixes #1732 (the CRLF/BOM false-positive class).

## Problem

Both frontmatter anchor regexes in `IntegrityCheck.ts` (`/^---\n([\s\S]*?)\n---/`, used by the commands/agents check and the skills check) fail on Windows checkouts:

- a UTF-8 BOM defeats the `^` anchor
- CRLF line endings defeat the `---\n` match

Every affected file then produces a false blocking `no frontmatter block` finding.

## Fix

Normalize the file read (strip `﻿`, convert CRLF to LF) before matching, at both parse sites. Four lines, no behavior change for LF files.

## Verification

A/B on Windows 11 (bun 1.3.14): a fixture `.md` written with a real BOM + CRLF (`EF BB BF` leading bytes) is flagged `BLOCK: no frontmatter block` by the current build and produces zero findings after this change.

Found while validating LifeOS native on Windows — the full hook pipeline runs there (verified against live `claude -p` sessions, plus a `windows-latest` CI probe suite). If Windows support is welcome, I have a small stack of similarly-scoped fixes ready to submit one at a time: `Bun.which()` for tool detection (`command -v` is absent from the Windows spawn shell), a directory-junction fallback in `setupUserSeparation` (the sibling of #1730's fix), a backslash-path normalization sweep for the hook layer (the #1119 class, with unit tests), and launcher spawn support for the `.cmd`/`.ps1` shims npm actually installs. Happy to open them in whatever order/shape you prefer.

Developed with Claude (co-author trailer on the commit); every change was probed on a real Windows machine before submission.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/IntegrityCheck.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/IntegrityCheck.ts
@@ -655,6 +655,9 @@ function checkFrontmatterUnits(name: string, dir: string, requireName: boolean):
   for (const f of files) {
     let content = '';
     try { content = readFileSync(join(dir, f), 'utf8'); } catch { continue; }
+    // Tolerate UTF-8 BOM and CRLF before anchoring — both defeat ^---\n and
+    // produced false "no frontmatter block" findings on Windows checkouts (#1732).
+    content = content.replace(/^\uFEFF/, '').replace(/\r\n/g, '\n');
     const fm = content.match(/^---\n([\s\S]*?)\n---/);
     if (!fm) { findings.push({ detail: `${name}/${f}: no frontmatter block`, blocking: true }); continue; }
     if (requireName && !/^name:\s*\S/m.test(fm[1])) findings.push({ detail: `${name}/${f}: missing frontmatter 'name:'`, blocking: true });
@@ -739,6 +742,7 @@ function checkSkills(): void {
     const rel = skillMd.replace(CLAUDE_DIR + '/', '');
     let content = '';
     try { content = readFileSync(skillMd, 'utf8'); } catch { continue; }
+    content = content.replace(/^\uFEFF/, '').replace(/\r\n/g, '\n'); // BOM/CRLF tolerance (#1732)
     const fmMatch = content.match(/^---\n([\s\S]*?)\n---/);
     if (!fmMatch) { findings.push({ detail: `${rel}: no frontmatter block`, blocking: true }); continue; }
     const fm = fmMatch[1];


danielmiessler:
Ported: `matchFrontmatter` now strips a leading BOM and tolerates CRLF (LIFEOS/TOOLS/IntegrityCheck.ts), shipped in v7.40.4. Your credit was missing at the fix site because the change landed via the #1732 thread; that's corrected in source now. Closing rather than merging: this repo is generated from private source at release, so a merge here would be overwritten.

PR #1808, opened 2026-08-10
fix(telos): read Context Filter from TELOS.md instead of a hardcoded value

0bsolescence:
## Problem

`GenerateTelosSummary.ts` emits a hardcoded string as the Context Filter section of `PRINCIPAL_TELOS.md`:

```
'When steering work, bias toward: human flourishing, Human 3.0 transition,
 AI augmentation strategies, becoming one's full self, correct framing.'
```

`PRINCIPAL_TELOS.md` is `@`-imported at every session start, and the Context Filter is the field that steers prioritisation. Because the literal is unconditional, every install renders the same filter regardless of what the user wrote in their own `TELOS.md`, silently replacing their steering criteria with someone else's.

## Fix

Read the `## Context Filter` section from unified `TELOS.md`, falling back to a neutral prompt when the section is absent.

## Verification

On a live 7.28.3 install with a populated Context Filter, `PRINCIPAL_TELOS.md` now renders the user's own text. With the section absent, it renders the fallback prompt rather than an unrelated value.

Scope: one function plus its call site. No behaviour change for any other section.

---
*Found and fixed with AI assistance while installing 7.28.3; verified on a live install rather than by inspection.*

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/GenerateTelosSummary.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/GenerateTelosSummary.ts
@@ -188,6 +188,24 @@ function paragraphItems(content: string, prefix: string): ParsedItem[] {
 /**
  * Parse mission items from MISSION.md
  */
+/** Read the "## Context Filter" H2 body from the unified TELOS.md. */
+function readContextFilter(): string {
+  try {
+    const raw = readFileSync(join(TELOS_DIR, 'TELOS.md'), 'utf-8');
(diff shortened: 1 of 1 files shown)


danielmiessler:
Ported: the Context Filter is read from TELOS.md's own section, with the hardcoded line demoted to fallback (LIFEOS/TOOLS/GenerateTelosSummary.ts). Ships in v7.40.4, credited in source and README. Closing rather than merging: this repo is generated from private source at release, so a merge here would be overwritten.

PR #1809, opened 2026-08-10
fix(telos): unified TELOS.md must take precedence over legacy per-topic files

0bsolescence:
## Problem

`readTelosFile()` checks the legacy per-topic path first and only falls through to the unified `TELOS.md` section when that file is absent:

```ts
const path = join(TELOS_DIR, filename);
if (existsSync(path)) { return readFileSync(path, 'utf-8')... }
```

`ScaffoldUser` copies the legacy template stubs into every fresh install, so those files always exist and always win. The documented contract that `LIFEOS/USER/TELOS/TELOS.md` is the single source of truth as of 2026-05-01 is therefore never true on a new install.

Observed on 7.28.3: after writing a real `TELOS.md`, `PRINCIPAL_TELOS.md` rendered `(sample)` template content, and `MISSION.md` tripped the zero-items guard so the summary refused to generate at all.

## Fix

Prefer the unified section whenever it carries content; fall back to the per-file path otherwise, preserving back-compat for pre-unification installs.

## Verification

Three cases on a live install:

- stub present + `TELOS.md` section present → `TELOS.md` wins (previously the stub won)
- `TELOS.md` section absent, stub present → output byte-identical to unpatched upstream, confirming the legacy path is not regressed
- neither present → unchanged

---
*Found and fixed with AI assistance while installing 7.28.3; verified on a live install rather than by inspection.*

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/GenerateTelosSummary.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/GenerateTelosSummary.ts
@@ -89,14 +89,13 @@ function loadTelosSections(): Record<string, string> {
 }
 
 function readTelosFile(filename: string): string {
-  // Legacy per-file path first (back-compat), then unified-TELOS section.
-  // Strip leading YAML frontmatter (public issue #1496): in files with no
-  // ID-form items, paragraphItems() otherwise swallows the frontmatter block
-  // as a garbage first entry (e.g. "M0: --- provenance: interview ...").
-  const path = join(TELOS_DIR, filename);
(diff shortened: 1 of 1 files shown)


danielmiessler:
Ported: unified TELOS.md now takes precedence, with legacy per-topic files as fallback (LIFEOS/TOOLS/GenerateTelosSummary.ts). Ships in v7.40.4, credited in source and README. Closing rather than merging: this repo is generated from private source at release, so a merge here would be overwritten.

PR #1821, opened 2026-08-11
fix: register six orphaned hooks, render runtime {{PRINCIPAL_NAME}}, add orphan check lane to HookHealer (#1817)  

### LINKED ISSUE #1817: Six hook scripts ship registered nowhere, four documented as live; unrendered {{PRINCIPAL_NAME}} reaches the model from two hooks
Verified against `danielmiessler/LifeOS` main at `58381b3` (2026-08-07), clean clone fetched 2026-08-10.

## Summary

Six files in `LifeOS/install/hooks/` are registered under no event in `hooks/hooks.json` and are imported by no sibling hook. `settings.system.json` and `settings.enhancements.json` carry no `hooks` key at all on main, so the manifest is the only place a registration could live. These six are dead from the moment they land: the code ships, the docs describe the behavior as running, and no event ever reaches them.

Separately, a raw template placeholder reaches model-facing context from two hooks, one of which is registered and firing today.

## The six

Computed by listing every `*.hook.{ts,sh}` in `LifeOS/install/hooks/` (51 files), subtracting every basename named in `hooks/hooks.json`, then dropping anything imported by a sibling. Exactly six remain:

| File | Header declares | In manifest | Imported by |
|---|---|---|---|
| `ModelRungGuard.hook.ts` | `TRIGGER: UserPromptSubmit` | no | nothing |
| `AtlasEventCapture.hook.ts` | `PostToolUse (Bash, Write, Edit, MultiEdit)` | no | nothing |
| `KnowledgeWriteGuard.hook.ts` | `PostToolUse (Write, Edit, MultiEdit)` | no | nothing |
| `SpendAuditor.hook.ts` | its own trigger set | no | nothing |
| `TimeContext.hook.ts` | per-turn wall-clock line | no | nothing |
| `VersionDrift.hook.ts` | `UserPromptSubmit` nag | no | nothing |

Dispatcher-imported hooks are excluded on purpose and are **not** part of this report. `FormatGate`, `VerificationGate`, `ISACloseGate`, `ISAGate`, `ISAFoldGate` and `WritingGate` are unregistered but reached through `StopGates.hook.ts` lines 30-35, which is deliberate design. `LIFEOS/DOCUMENTATION/Hooks/HookSystem.md` line 22 already documents that whole dispatcher-imported set.

## Why this is not just dead code

**Four of the six are documented as registered and running**, in tables and in the ASCII wiring diagram:

- `hooks/README.md` line 210 lists `AtlasEventCapture.hook.ts` with matchers and timeout; line 80 draws it.
- `hooks/README.md` line 182 lists `VersionDrift.hook.ts`; line 60 draws it.
- `hooks/README.md` line 237 lists `SpendAuditor.hook.ts`; lines 93 and 346 draw it.
- `hooks/README.md` line 186 lists `TimeContext.hook.ts`; line 64 draws it, and line 636 names it as one of only four hooks that should write to stdout.

**One is documented with a registration date it never got.** `HookSystem.md` line 22 states `KnowledgeWriteGuard` was "registered 2026-07-29", line 321 shows it inside a `settings.json` snippet, and line 351 describes its runtime behavior in present tense. It appears in no registration file on main.

**One has a consumer that can never succeed.** `LIFEOS/TOOLS/Reflect.ts` line 49 reads `MEMORY/OBSERVABILITY/spend-audit.jsonl`, and line 190 states that `within_budget` "is derived from spend-audit.jsonl and cannot be supplied". `SpendAuditor.hook.ts` line 44 is that file's only writer. Unregistered, it never writes, so that derived field has no path to a true value on any install.

**`ModelRungGuard` is the sharpest case.** Its header (lines 14-16) records the incident that motivated it: a session ran a rung below the pinned model for its entire length and nothing noticed, "not a hook, not a gate". As shipped, that incident recurs undetected, because the detector is wired to no event.

The same drift shows up in the counts. `HookSystem.md` line 22 says "49 `.hook.ts` files exist on disk" against a tree that ships 51, and flags earlier stale counts already reported in #1596 and #1629. The counts drift because nothing checks them.

## Second defect: a placeholder reaching the model verbatim

`ModelRungGuard.hook.ts` builds its off-pin advisory across lines 164-169 and hands it to `emit()`, which returns it as `additionalContext` at line 149. Line 168 is:

```
`{{PRINCIPAL_NAME}} to change /model.`,
```

That sits inside the emitted template literal, not in a comment. Nothing renders it at hook runtime, so every off-pin fire hands the model a raw `{{PRINCIPAL_NAME}}` token. Placeholder substitution is a tree-wide install-time pass, so whether this is ever rendered depends on the install path rather than on the hook being correct.

**The same bug exists in a hook that IS registered.** `PromptProcessing.hook.ts` lines 708 and 721 embed the literal `{{PRINCIPAL_NAME}}` token inside the `buildContextPrompt()` template literal, on lines adjacent to correct `${PRINCIPAL_NAME}` interpolation. That return value is consumed as a `systemPrompt` for a live inference call, so this one leaks on installs today regardless of the orphan issue.

Grepping all 51 hook files for the literal token outside comments returns exactly three lines, so the fix is small and complete.

## Why nothing catches this

`HookHealer.hook.ts` (1.0.3 on main) is the natural place for an orphan check and has none. Its functions are `log`, `isExecutable`, `hasShebang`, `heal`, `expandHome`, `directExecPaths`, `sweep`, `readStdin`, `posttool`, `main`. It heals executable bits and shebangs; it never compares the files on disk against the set of registered names.

## Reproduce

From `LifeOS/install/` at `58381b3`:

```sh
# 1. none of the six appear in the manifest
for f in ModelRungGuard AtlasEventCapture KnowledgeWriteGuard \
         SpendAuditor TimeContext VersionDrift; do
  grep -q "$f" hooks/hooks.json || echo "$f: not in manifest"
done

# 2. none of the six are imported by a sibling
for f in ModelRungGuard AtlasEventCapture KnowledgeWriteGuard \
         SpendAuditor TimeContext VersionDrift; do
  grep -l "from \"\./$f\.hook\"" hooks/*.hook.ts || echo "$f: imported by nothing"
done

# 3. no other registration surface exists
python3 -c "import json; print('hooks' in json.load(open('settings.system.json')))"        # False
python3 -c "import json; print('hooks' in json.load(open('settings.enhancements.json')))"  # False

# 4. the placeholder, outside comments
grep -rn '{{PRINCIPAL_NAME}}' hooks/*.hook.ts \
  | grep -v ':[0-9]*: *\*' | grep -v ':[0-9]*: *//'
```

Step 4 returns `ModelRungGuard.hook.ts:168`, `PromptProcessing.hook.ts:708`, `PromptProcessing.hook.ts:721`.

For the placeholder at runtime: feed `ModelRungGuard.hook.ts` a payload whose `transcript_path` points at a one-line JSONL file whose last assistant message carries a model below the pinned tier. The emitted `additionalContext` ends with the raw token.

## Suggested fix

Three parts, independent of each other.

1. **Decide per file: register it, or delete it and the docs that describe it as active.** A file in `hooks/` that no event reaches is worse than no file, because the README table and wiring diagram read as a description of live behavior. Two of the six look like genuine design questions rather than plain omissions: `AtlasEventCapture` only has a consumer where the Atlas tick is installed, and `TimeContext` overlaps whatever else injects a per-turn clock, so deletion may be the right call there rather than registration.

2. **Render or interpolate the placeholder.** Three lines across two files. `PromptProcessing` is the urgent half, since it is registered and firing.

3. **Add an orphan lane to `HookHealer`'s sweep** so this class cannot recur. It is cheap: one directory listing, the manifest, one import scan. It should be warning-only, since deciding when a hook fires is a human call, and it should compare **per event**, not per bare script name. A hook registered on one event while its header declares three looks correctly wired to any check comparing basenames alone. The same lane would keep the file counts in `HookSystem.md` from drifting again.

I have a working implementation of the orphan check with the import-based exclusion, and can open a PR if that is welcome.

firaslamouchi21:
Fixes #1817 (opened by @catchingknives).

I verified everything against a clean clone at `58381b3` before touching anything, and all three defects reproduced right away: six hooks were dead on disk, raw template placeholders were getting passed to live model context, and neither settings file had a `hooks` key.

Here's what this PR changes:

### 1. Registered the 6 orphan hooks
`VersionDrift`, `ModelRungGuard`, `TimeContext`, `AtlasEventCapture`, `KnowledgeWriteGuard`, and `SpendAuditor` were absent from `hooks/hooks.json` and imported by no sibling dispatcher, meaning they ran zero times on every install. 

Rather than deleting any—especially `AtlasEventCapture` and `TimeContext`, which were flagged in the issue as potential design questions—I registered all six based on the exact events, positions, and timeouts already outlined in the docs. Tearing out docs-sanctioned subsystem features felt like a maintainer call, and registering them keeps things aligned with the documented behavior (and is easily reverted if you'd rather cut them).

*Mechanical details:* 
* `AtlasEventCapture` needed a new `Bash` matcher bucket under `PostToolUse`—this is the only structural addition to the manifest, as everything else joined existing groups.
* Registering `SpendAuditor` unblocks `Reflect.ts`’s `within_budget` check, since `SpendAuditor` is the sole writer for `spend-audit.jsonl`.

### 2. Fixed raw `{{PRINCIPAL_NAME}}` leaking to model context
Since template substitution is just an install-time pass on disk, runtime strings generated on the fly never got rendered. 

* **`PromptProcessing.hook.ts` (lines 708, 721):** Swapped the raw token for the `${PRINCIPAL_NAME}` constant already in scope at line 682. This was leaking on live system prompts on every install.
* **`ModelRungGuard.hook.ts` (line 168):** Pulled in `getPrincipalName()` from `lib/identity` wrapped in a generic fallback so it honors its contract to never fail a prompt.

Tested end-to-end against a synthetic off-pin transcript:
* **Before:** `…do not ask {{PRINCIPAL_NAME}} to change /model.`
* **After:** `…do not ask Daniel to change /model.`

Grepping all 51 hook files for the literal token outside comments now returns zero hits.

### 3. Added an orphan check lane to HookHealer (1.1.0)
Added a lightweight check to `HookHealer` so we don't run into dead hooks again. (Note: The 154 changed lines in `HookHealer` include a refactor of the settings walk to return `(event, command)` pairs parsed once and shared with the existing exec-bit lane, changing the `directExecPaths()` signature. The new check is cheap: just one directory listing plus one read per hook file.)

* **Warning only:** It never auto-registers hooks, since deciding when a hook fires should be a human call.
* **Per-event, not per-filename:** Keys registrations by event, so a hook wired to one event while its header declares three doesn't read as correctly wired. A second lane compares explicit `TRIGGER:` headers against actual registrations to catch partial wiring.
* **Dispatcher-aware:** Explicitly ignores the 14 dispatcher-imported hooks so it doesn't trigger false positives.

Ran four sanity checks against simulated installs:
* Pre-fix manifest → flags the exact 6 orphans.
* Post-fix manifest → exit code 0, completely silent.
* Un-registering a hook → single targeted warning.
* Event mismatch → catches header vs manifest mismatch properly.

### 4. Docs note (intentionally not included in this PR)
I avoided editing `hooks/README.md` and `HookSystem.md` in this PR since upstream releases overwrite repo-level edits. However, for when you sync upstream:
* The 4 hooks documented as "live" are now actually live, making the existing README tables and wiring diagrams accurate as written.
* Disk count breakdown for reference: 50 `.hook.ts` files on disk (51 with `.sh`), 36 registered directly, 14 dispatcher-imported (missing `ISAStaleWriteGuard`, `ISAFoldGate`, `LoopDetector` from the docs dispatcher list).
* `KnowledgeWriteGuard` was never registered on `2026-07-29` as claimed in the docs—this PR is its first actual registration.

### DIFF
--- a/LifeOS/install/hooks/HookHealer.hook.ts
+++ b/LifeOS/install/hooks/HookHealer.hook.ts
@@ -1,7 +1,8 @@
 #!/usr/bin/env bun
 /**
- * @version 1.0.3
- * HookHealer.hook.ts - Self-healing for the registered-script exec-bit class
+ * @version 1.1.0
+ * HookHealer.hook.ts - Self-healing for the registered-script exec-bit class,
+ * plus the orphaned-hook lane (a script on disk that no event ever reaches).
  *
  * PURPOSE:
  * The Write tool creates files mode 0644. A hook registered in settings as a
@@ -9,11 +10,19 @@
  * invocation with "/bin/sh: Permission denied" until someone notices.
  * This hook detects and repairs that class automatically.
  *
+ * The orphan lane (1.1.0) catches the opposite problem: a hook that ships, gets
+ * documented as if it were running, and is wired to nothing. Six of them were
+ * sitting in here like that (public issue #1817) — the README table and the
+ * wiring diagram both described them as live. Nobody had ever compared the
+ * files on disk against what's actually registered, so it went unnoticed until
+ * someone audited the manifest by hand.
+ *
  * MODES:
  * - (default)  SessionStart sweep: every script directly executed by a
  *              settings hook command (first token of each command segment)
  *              must exist and be executable. Missing exec bit -> chmod +x.
  *              Missing file / missing shebang -> surfaced warning only.
+ *              Then the orphan lane over ~/.claude/hooks (warning only).
  * - --posttool PostToolUse(Write|Edit) ingestion guard: a written file under
  *              ~/.claude whose content starts with "#!" gets its exec bit
  *              immediately - heals at the ingestion point.
@@ -22,22 +31,28 @@
  * - chmod containment: only ever touches paths under ~/.claude
  * - non-blocking: exits 0 on every path, including internal errors
  * - registered via "bun <path>" so it is immune to losing its own exec bit
+ * - the orphan lane never fixes anything, it just tells you. Deciding when a
+ *   hook should fire is a human call, and auto-registering something would be
+ *   a worse bug than the one we're reporting.
  *
  * OUTPUTS:
- * - MEMORY/OBSERVABILITY/hook-healer.jsonl (heal/warning events)
+ * - MEMORY/OBSERVABILITY/hook-healer.jsonl (heal/warning/orphan events)
  * - stdout "🩹 HookHealer: ..." line when something was healed or needs attention
  *
- * PERFORMANCE: <50ms (two JSON reads + stat per registered path)
+ * PERFORMANCE: <50ms for the exec-bit sweep (two JSON reads + stat per
+ * registered path). The orphan lane adds one directory listing plus a read of
+ * each hook file (~570KB across the shipped tree) for the import scan.
  */
 
 import {
   existsSync, readFileSync, chmodSync, statSync, appendFileSync,
-  mkdirSync, openSync, readSync, closeSync, realpathSync,
+  mkdirSync, openSync, readSync, closeSync, realpathSync, readdirSync,
 } from 'fs';
 import { join } from 'path';
 import { homedir } from 'os';
 
 const CLAUDE_DIR = join(homedir(), '.claude');
+const HOOKS_DIR = join(CLAUDE_DIR, 'hooks');
 const OBS_DIR = join(CLAUDE_DIR, 'LIFEOS', 'MEMORY', 'OBSERVABILITY');
 const LOG_FILE = join(OBS_DIR, 'hook-healer.jsonl');
 const SETTINGS_FILES = ['settings.json', 'settings.local.json'];
@@ -95,12 +110,15 @@ function expandHome(token: string): string {
 }
 
 /**
- * Collect scripts that settings hook commands execute DIRECTLY (first token
- * of each command segment). Scripts passed as arguments to bun/sh are
- * deliberately excluded - their exec bit is irrelevant.
+ * Every (event, command) pair from the settings files. Read once and used by
+ * both lanes below.
+ *
+ * We keep the event name rather than just the command because a hook can be
+ * registered on one event while its header says it runs on three — and if you
+ * only compare script names, that looks perfectly wired.
  */
-function directExecPaths(): Set<string> {
-  const paths = new Set<string>();
+function registeredCommands(): Array<{ event: string; command: string }> {
+  const out: Array<{ event: string; command: string }> = [];
   for (const name of SETTINGS_FILES) {
     const file = join(CLAUDE_DIR, name);
     if (!existsSync(file)) continue;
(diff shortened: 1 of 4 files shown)


danielmiessler:
Parts 1 and 2 are already resolved in the current release line: the shipped hooks.json is generated from the live settings registration (all six hooks present in the current cut), and the runtime placeholder strings were fixed with a release gate (G23) that now blocks any unrendered token in shipped code. The HookHealer orphan lane we're declining for now — the root cause was the manifest generation, which is fixed structurally, and we keep the hook surface minimal — but it was a well-built diff and the report that drove it (#1817) was right.

PR #1832, opened 2026-08-12
fix(memory): never resolve an unsourced value conflict at ~explicit

Chuckos:
## Problem

The memory reviewer resolves conflicting factual values by picking a winner and stamping it `~explicit` — as if the principal had stated it. When memory already holds ≥2 distinct values for the same topic (a weekday, a time, a quantity) and the reviewer keeps one while evicting a disagreeing sibling, that's the reviewer *guessing*, not recording something the principal said — yet the guess carries the highest-confidence marker.

This is a confabulation defect: a gap filled by inference, then marked as fact. Consumers that trust `~explicit` (and any metric derived from a conflicted value, e.g. a session count) are silently corrupted until someone notices.

## Fix

Add a deterministic `detectValueConflicts()` in `MemoryWriter.ts`, run in-lock during `setEntries`:

- **Closed vocabularies only** — weekday names, `HH:MM`, `number+unit`. Anything outside is left alone by design; a broader rule would have to *infer*, which is the very defect.
- **Fires only on the true signature** — prior state held ≥2 distinct values **and** a disagreeing sibling was actually evicted. Ordinary supersession (one prior value + a new one) and honest consolidation are untouched.
- **Downgrades, never rejects** — a flagged entry's provenance is forced to `~inferred` (text preserved). Curation keeps flowing; authority does not.

Two false-positive guards: unit-scoped comparison (7h is not a rival of 150g), and a two-shared-word topic match (so a single shared token doesn't link two entries that are plainly about different things).

## Tests

Adds `MemoryWriter.conflict.test.ts` — 8 cases: the real regression plus false-positive guards (unit scoping, two-word anchor, ordinary supersession, no-closed-vocab, unchanged entries).


danielmiessler:
The prompt-guidance half is ported: the reviewer now supersedes only on evidence and never resolves an unsourced conflict at ~explicit (your wording, lightly adapted, credited). We passed on the deterministic closed-vocabulary detector for now — the regex vocabularies (weekdays/times/quantities) pin behavior to today's shapes, and we bias hard toward less machinery — but the replay numbers were genuinely persuasive and the idea is recorded with credit if provenance regressions recur.

PR #1835, opened 2026-08-13
iMessage module: fix sending on modern macOS, see images, survive message-to-self and iCloud re-sync

J0UH:
I set LifeOS up on my Mac this week and made the iMessage module my main channel, with a twist: I text my own number, so the DA lives in my "messages to yourself" thread. That surfaced a string of bugs, a couple of which will bite any iMessage user, not just the self-chat case. Everything here is running on my machine now (macOS 26.5, LifeOS 7.28.3).

**Sending was broken outright.** `participant targetService handle "..."` is not a valid AppleScript object specifier, at least on current macOS. osascript exits 1, every reply silently fails, and the SDK sessions still run and cost money. The working form is `participant "..." of targetService`. That's the first commit, and it's independent of everything else if you'd rather cherry-pick it.

The second commit is the rest:

**Images.** The module only read chat.db's `text` column, so photos were invisible, and attachment-only messages never even reached the bot because of the `text IS NOT NULL` filter. Attachments now get staged into the module's state dir before the session starts. The copy step matters: only the bun process holds Full Disk Access, the claude subprocess it spawns does not, so it can't read ~/Library/Messages itself. HEIC converts to JPEG with sips, and the staged paths go into the prompt for the Read tool. Tested with photos and screenshots, the DA describes them fine.

**Message-to-self.** If the allowlisted handle is your own Apple ID, every reply the bot sends comes back as an is_from_me=0 row, and the poller treats it as a new user message. Mine answered itself 255 times before I pulled the plug. Replies now carry a zero-width-space prefix and the poller skips anything that starts with one. There's a second problem in the same setup: iCloud does not reliably generate that received-echo copy of your own sent messages. Mine stopped generating them mid-day and the bot just went deaf. The poller now also accepts is_from_me=1 rows, but only inside a chat whose identifier is on the allowlist, with a dedup set for when an echo copy shows up late. Side benefit: you can talk to the DA straight from Messages.app on the Mac.

**iCloud re-sync.** This one cost me a morning. Overnight, iCloud re-inserted my entire previous day of messages under brand-new ROWIDs with their original timestamps. The cursor is ROWID-based, so at 8:46am the bot saw ~240 "new" messages and answered every single one, including naming each emoji I'd sent the day before. In Danish. The fix is a freshness gate: anything whose own timestamp is older than 15 minutes gets skipped no matter what ROWID it arrives on. ROWID order just isn't arrival order after a re-sync or a restore, so this protects every user of the module.

**Dropped messages.** Anything sent while a session was in flight got "Still processing your previous message. Please wait." and was then lost forever, because the cursor had already moved past it. poll() now returns early while processing, so pending messages stay in the DB and get handled in order on the next tick.

Happy to split any of this out or adjust the freshness window if you want it configurable.

### DIFF
--- a/LifeOS/install/LIFEOS/PULSE/lib/imessage-send.ts
+++ b/LifeOS/install/LIFEOS/PULSE/lib/imessage-send.ts
@@ -25,7 +25,9 @@ export async function sendMessage(
   const chunks = splitMessage(text, MAX_MESSAGE_LENGTH)
 
   for (const chunk of chunks) {
-    const escaped = escapeForAppleScript(chunk)
+    // Zero-width-space prefix marks this as a bot message so the poller
+    // can ignore its own echo in message-to-self conversations.
+    const escaped = escapeForAppleScript("\u200B" + chunk)
 
     // Use buddy-based send — works reliably on modern macOS
     const script = `
--- a/LifeOS/install/LIFEOS/PULSE/lib/imessage-send.ts
+++ b/LifeOS/install/LIFEOS/PULSE/lib/imessage-send.ts
@@ -31,7 +31,7 @@ export async function sendMessage(
     const script = `
 tell application "Messages"
   set targetService to 1st account whose service type = iMessage
-  set targetBuddy to participant targetService handle "${escapeForAppleScript(handle)}"
+  set targetBuddy to participant "${escapeForAppleScript(handle)}" of targetService
   send "${escaped}" to targetBuddy
 end tell`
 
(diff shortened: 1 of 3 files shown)


danielmiessler:
The AppleScript send fix is ported with credit — the old object-specifier form is dead on current macOS and your replacement works on both. The other legs (attachment staging with the sips conversion, self-chat trust widening, freshness gate) are queued for a dedicated security review before adoption: they widen the module's trust boundary and add an exec path, which we don't land in a batch port. The engineering is good; we're just slow on purpose where the boundary moves.

PR #1846, opened 2026-08-14
DA template: ship the schema DaSubsystem.md already defines

justinkatz94-glitch:
DaSubsystem.md defines DA Identity Schema v1.0 (frontmatter: schema_version, core, voice, personality, autonomy, and the wider field set) and a migration table for existing installs. The shipped bootstrap template `install/USER/DIGITAL_ASSISTANT/DA_IDENTITY.md` never adopted it, so every fresh install is born schema-less and immediately subject to the doc's own migration path.

This PR adds the v1.0 frontmatter to the template with neutral bootstrap values only (the same "LifeOS" placeholder identity the template already uses); `/interview` populates the real values, and absent fields stay absent rather than being invented.

Validated on a live install: frontmatter added with the markdown body byte-identical pre/post (sha256 on the body), YAML parse green, and a cap/schema guard test at 4/4 with three planted-failure reds before trusting the green. Happy to contribute that guard test as a follow-up if useful.

### DIFF
--- a/LifeOS/install/USER/DIGITAL_ASSISTANT/DA_IDENTITY.md
+++ b/LifeOS/install/USER/DIGITAL_ASSISTANT/DA_IDENTITY.md
@@ -3,6 +3,23 @@ provenance: template
 last_updated: 1970-01-01T00:00:00Z
 last_updated_by: bootstrap-template
 convention: pai-freshness-v1
+schema_version: 1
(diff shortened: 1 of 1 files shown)


danielmiessler:
Ported into the template of record verbatim (the release builds the payload from a private source tree, so PRs get ported rather than merged — same change, durable across releases). Credit comment carries your handle. The guard test would be welcome as a follow-up.

PR #1855, opened 2026-08-15
TextFidelity: the document-fidelity gate — did every source word reach the rendered page?

xmasyx:
The document-fidelity checker from the #1614 offer, sent as requested — one at a time, this one next.

## What it is

One shared definition of the question every document renderer eventually fails at: **did every word of the source arrive in the rendered output?** Three pure functions plus a docx reader, designed to be called by any render pipeline (PDF, docx) so two pipelines never hold two diverging copies of the same logic.

- `sourceVocabulary(md)` — the words and numbers (4+ chars) that MUST survive rendering. The exclusions are the hard-won part: fenced blocks and frontmatter; an image's **path** (but never its caption); a link's **URL** (but never its label); HTML comments. Edge punctuation is trimmed **before** the length filter — without that, the last word of every sentence and every formatted number (`5.000,00`) silently left the vocabulary, and the gate certified what it never looked at.
- `lostWords(vocab, renderedText)` — presence check, deliberately narrow. A correct text layer is not a correct page; pages still get looked at. This exists for what the eye cannot do: check hundreds of words one by one.
- `lostSymbols(src, rendered)` — a non-ASCII census that catches what word-level checks cannot see: a vanished accent, an em-dash rendered as `?`, the character destruction an incomplete-font renderer produces.
- `docxText(path)` — what the .docx **actually** contains, read from the OOXML. Run boundaries are the one delicate point: `</w:p>` and `<w:br/>` become one space, every other tag vanishes with no space — bold splits words across runs, and a verifier that inserts spaces between runs invents losses.

## Receipts

`bun TextFidelity.ts --selftest` carries both poles: a dropped word goes red, an intact render goes green, the exclusions never cry wolf, and the symbol census catches `caffè — sì` rendered as `caffe - si`.

On the install this was extracted from, the word-level gate has caught a silently dropped command literal (nested code spans), the URL exclusion was added after a real client document declared its own product name lost, and the ENOBUFS ceiling on `unzip -p` was raised after the gate died on a well-built 29k-word document — the worst failure shape, because it looks like a defective document.

Comments are English per the #1614 guidance. No personal data; the file is self-contained (bun + `unzip`).

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/TextFidelity.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/TextFidelity.ts
@@ -1,0 +1,144 @@
+#!/usr/bin/env bun
+/**
+ * TextFidelity.ts — "did every source word arrive in the rendered document?", defined ONCE.
+ *
+ * WHY IT EXISTS, AND WHY IT IS SHARED
+ * A PDF text gate had been asking this question for weeks (and caught a command
+ * literal silently dropped by a nested-code-span bug) while the docx path had
+ * NOTHING of the kind — its only verification was looking at pages, on the very
+ * day its backend was rewritten, i.e. when losing text nodes is most likely.
+ * If the two pipelines held two copies of this logic they would diverge: one
+ * would learn that image paths are not content and the other would not, and the
+ * day one of them stays silent you cannot tell healthy from ignorant.
+ * One definition, N callers.
+ *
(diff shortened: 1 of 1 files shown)


danielmiessler:
Ported verbatim with your header intact plus a credit line; `--selftest` green on our side. Closing per the port flow. The negative-pole discipline in this file is the standard we point people at now. Thanks.

PR #1857, opened 2026-08-15
UpgradeVerify: deterministic post-upgrade verification — baseline-relative, selftest-proven

xmasyx:
The tool offered in [this #1614 comment](https://github.com/danielmiessler/LifeOS/issues/1614#issuecomment-5301625444), sent as a PR per the one-at-a-time flow (the fidelity gate is #1855).

## What it is

`bun LIFEOS/TOOLS/UpgradeVerify.ts --baseline` **before** an upgrade, `bun UpgradeVerify.ts` **after**. An upgrade that ends with 'files copied, exit 0' has proven nothing — the failure modes that bite are the ones no copy step sees. Eight read-only checks, each one born from an incident a live install actually paid for during the 7.28.3 → 7.40.4 jump:

1. **Version coherence** — VERSION exists and parses.
2. **Typecheck, baseline-relative** — the gate is *no NEW errors vs the pre-upgrade snapshot*, never *zero errors*: a tree may carry pre-existing debt that is not the upgrade's fault, and an absolute gate teaches operators to ignore red.
3. **Registered hooks resolve** — every hook command in settings.json points at a file that exists ($CLAUDE_PROJECT_DIR/$HOME/~ expanded; the extension token must END the path — 'settings.system.json' once matched as '.js' and reported ghosts).
4. **Inert hooks, import-aware** — hook files neither registered *nor imported by a live hook*. The import graph is followed because an aggregator (a Stop-gate runner) keeps hooks live that no settings line names; without that, the check cries wolf forever.
5. **Local-delta verifier** — runs it when the install keeps one, skips cleanly otherwise.
6. **launchd targets** — ProgramArguments resolved against WorkingDirectory before being called missing (a relative `pulse.ts` + WorkingDirectory is healthy; a naive absolute-only probe reports a false red).
7. **Services + statusline** — the dashboard answers on its configured port; the statusline renders non-empty.
8. **Cross-version evidence skew** — the class 7.40.4 taught us: a fail-closed validator (CortexHealth's reviewer-run shape) judging an evidence row written by the *pre-upgrade* producer goes critical on a perfectly healthy system. The check detects 'newest evidence predates the upgrade' and names it *stale evidence, self-heals on the next producer run* — because a red the operator has to wait out quietly becomes a red they ignore.

## Receipts

- `--selftest`: four poles, including a planted ghost hook registration (must go critical) and a garbage VERSION (must go critical). A gate never seen red is an assertion in a gate costume.
- `--decisions [path]`: renders the non-green findings as an answer sheet — one question per item, a recommended answer with the reason, an empty `Answer:` line — because an upgrade's leftovers are choices someone should make once, on paper.
- Read-only except its own state dir (`MEMORY/STATE/upgrade-verify/`). Root from `--root`/`CLAUDE_CONFIG_DIR`; no personal data; English throughout.

Pairs naturally with OverlaySystem: overlay refreshes, this proves the refresh landed healthy.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/UpgradeVerify.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/UpgradeVerify.ts
@@ -1,0 +1,358 @@
+#!/usr/bin/env bun
+/**
+ * UpgradeVerify.ts — deterministic post-upgrade verification for a LifeOS install.
+ *
+ *   bun UpgradeVerify.ts --baseline          # BEFORE upgrading: snapshot what "healthy" looks like
+ *   bun UpgradeVerify.ts                     # AFTER upgrading: verify against the snapshot
+ *   bun UpgradeVerify.ts --json              # machine-readable report
+ *   bun UpgradeVerify.ts --selftest          # prove the checks can fail (negative poles)
+ *
+ * WHY THIS EXISTS. An upgrade that ends with "files copied, exit 0" has proven
+ * nothing: the failure modes that actually bite are the ones no copy step sees —
+ * a hook registered in settings.json whose file vanished, a typecheck error
+ * introduced by lib/consumer version skew, a launchd job whose target moved, a
+ * dashboard that never came back up, an evidence log written in the OLD format
+ * being judged by a NEW fail-closed validator. Every check below encodes an
+ * incident one install actually paid for.
+ *
+ * DESIGN RULES:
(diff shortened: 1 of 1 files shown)


danielmiessler:
Ported verbatim with credit; selftest green, baseline seeded on the maintainer install, and it's wired into the update workflow's verification step alongside the overlay fixes from this batch. Closing per the port flow. Thanks.

PR #1863, opened 2026-08-15
fix(hermes): Policy.ts CLI no longer discards the destination when --launcher is absent

### LINKED ISSUE #1859: Policy.ts CLI silently discards its destination argument when --launcher is absent
**Version:** v7.40.4
**File:** `LifeOS/install/LIFEOS/HERMES/Policy.ts` (the `import.meta.main` block)

**Repro**

```bash
bun LifeOS/install/LIFEOS/HERMES/Policy.ts /tmp/policy.json
# prints the full policy JSON to stdout, exits 0
ls /tmp/policy.json
# No such file or directory
```

Expected: `✓ policy v… → /tmp/policy.json` and the file written.
Actual: the destination argument is discarded, the CLI takes the "no dest → print to stdout" path, and no file is ever written. Exit code is 0, so callers (scripts, test harnesses that generate a policy to a temp path) fail later with a confusing `FileNotFoundError` on the policy file rather than at the point of the actual bug.

**Cause**

```ts
const launcherIdx = argv.indexOf("--launcher");
// …
const positional = argv.filter((a, i) => a !== "--launcher" && i !== launcherIdx + 1);
```

When `--launcher` is not passed, `launcherIdx` is `-1`, so the filter condition `i !== launcherIdx + 1` becomes `i !== 0` — which removes `argv[0]`, the destination path. `positional[0]` is then `undefined` and the stdout branch runs.

With `--launcher` present the CLI behaves correctly; the bug only bites the plain `bun Policy.ts <dest>` form.

**Fix**

Guard the sentinel value:

```ts
const positional = argv.filter(
  (a, i) => a !== "--launcher" && (launcherIdx === -1 || i !== launcherIdx + 1),
);
```

Happy to send this as a PR (3-line change, no behavior change for the `--launcher` form).

hjbrandt:
Fixes #1859.

`Policy.ts`'s CLI filters `--launcher` and its value out of `argv` before reading the positional destination:

```ts
const positional = argv.filter((a, i) => a !== "--launcher" && i !== launcherIdx + 1);
```

When `--launcher` is absent, `launcherIdx` is `-1` and `i !== launcherIdx + 1` excludes index 0 — the destination itself. `bun Policy.ts <dest>` then prints to stdout and never writes the file, exit 0.

**Change** (one file, `LifeOS/install/LIFEOS/HERMES/Policy.ts`): guard the sentinel value so the value-slot exclusion only applies when `--launcher` was actually passed.

Verified: `bun Policy.ts /tmp/policy.json` writes the file and prints the `✓ policy v…` line; `bun Policy.ts --launcher name /tmp/policy.json` unchanged; no-arg stdout form unchanged.

### DIFF
--- a/LifeOS/install/LIFEOS/HERMES/Policy.ts
+++ b/LifeOS/install/LIFEOS/HERMES/Policy.ts
@@ -409,7 +409,9 @@ if (import.meta.main) {
   const argv = process.argv.slice(2);
   const launcherIdx = argv.indexOf("--launcher");
   const launcherName = launcherIdx !== -1 ? (argv[launcherIdx + 1] ?? "") : "";
-  const positional = argv.filter((a, i) => a !== "--launcher" && i !== launcherIdx + 1);
+  const positional = argv.filter(
+    (a, i) => a !== "--launcher" && (launcherIdx === -1 || i !== launcherIdx + 1),
+  );
   const dest = positional[0];
   if (!dest) {
     process.stdout.write(JSON.stringify(buildPolicy({ launcherName }), null, 2) + "\n");


danielmiessler:
Ported with credit at the site (`LIFEOS/HERMES/Policy.ts`, your exact guard) and verified: `bun Policy.ts /tmp/x.json` writes the file, both other forms unchanged. Closing per the port-into-source flow—the public repo is generated from the private tree, so merging here would be overwritten at the next release; the fix ships with your name on it. Thanks.

PR #1866, opened 2026-08-15
feat(voice): ordered TTS provider failover chain with per-rung liveness probe

0bsolescence:
## What

An ordered provider failover chain for Pulse voice: `[[voice.providers]]` entries in PULSE.toml are tried in order (e.g. a local OpenAI-compatible endpoint such as Kokoro first, ElevenLabs as fallback), with a per-rung liveness probe and a verified request contract so a dead local server degrades to the next provider instead of silence. Existing single-provider ElevenLabs configs work unchanged.

## Commits

- `feat(voice)`: ordered provider failover chain (providers.ts + chain wiring in voice.ts)
- `feat(voice)`: per-rung liveness probe, verified request contract
- `docs(voice)`: pcm playback caveat in the format map and PULSE.toml
- `test(voice)`: honour $HOME override over homedir() so the suite's HOME sandbox holds under Bun (base-adaptation; without it the suite reads the operator's real settings.json)

## Testing

`bun test test/VoiceServer/` — 61 pass, 0 fail (providers.test.ts + voice.test.ts), run against current `main`.

Developed with AI assistance (Claude); reviewed and tested by me.

### DIFF
--- a/LifeOS/install/LIFEOS/PULSE/PULSE.toml
+++ b/LifeOS/install/LIFEOS/PULSE/PULSE.toml
@@ -93,6 +93,9 @@ enabled = true
 #   # health_path = "/health"            # liveness path, relative to the root
 #   # health_timeout_ms = 2000           # liveness budget
 #   # api_key = "${LOCAL_TTS_KEY}"       # sent as a bearer token when set
+#   # send_stream_flag = false           # omit Kokoro's stream:false for a STRICT
+#   #                                    # OpenAI endpoint that rejects the field
+#   #                                    # (default true — keep it for Kokoro)
 #
 #   [[voice.providers]]
 #   type = "elevenlabs"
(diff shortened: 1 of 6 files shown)


danielmiessler:
This is queued for its own session: a 1.7K-line change to the live voice path earns a dedicated review-and-verify pass with live TTS on real hardware and failover exercised in both directions. The design reads right—ordered chain, per-rung liveness, existing configs unchanged—and local-first TTS is a direction we want for installs without ElevenLabs keys. Expect movement within the week.

danielmiessler:
Declining, with a real caveat: we do want voice fallback, and you're right that a single provider is a single point of failure.

Two reasons this shape doesn't land. It's 1,678 lines into `voice.ts`, which is actively developed — it changed the day before this triage — so it's a permanent merge cost on a hot file for resilience we can get from a much smaller local-TTS fallback. And the admin-configurable `base_url` adds a credentialed egress path we'd then have to guard forever; nothing wrong with your implementation, it's just a new class of thing to own.

Nothing here reads as unsafe: credentials go out as headers to either the hardcoded vendor host or the operator's own configured endpoint, `providerLabel` explicitly excludes them, and the whole feature is opt-in and commented out by default. Reviewed closely, found clean.

We'll likely land a minimal local-TTS fallback instead and credit this issue for the push.

PR #1867, opened 2026-08-15
feat(tools): Trajectory — query the session transcript as structured data

0bsolescence:
## What

`LIFEOS/TOOLS/Trajectory.ts`: a read-only CLI that treats Claude Code session transcripts (JSONL) as queryable data — filter by session, time window, tool, or text; emit structured summaries instead of grepping raw JSONL. Useful for post-session review, hook debugging, and building on-disk observability without a server.

Standalone: one new tool file plus its test, no changes to existing files.

## Testing

`bun test test/Trajectory.test.ts` — 63 pass, 0 fail, 153 assertions, run against current `main`.

Developed with AI assistance (Claude); reviewed and tested by me.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/Trajectory.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/Trajectory.ts
@@ -1,0 +1,815 @@
+#!/usr/bin/env bun
+/**
+ * Trajectory.ts — the session transcript as queryable data.
+ *
+ * Claude Code already writes the full trajectory of every session to disk as
+ * JSONL under ~/.claude/projects/<slugged-cwd>/<session-uuid>.jsonl. That record
+ * is the only durable account of what actually happened — which files were
+ * touched, which tools failed, what was said and when. Until now the only way to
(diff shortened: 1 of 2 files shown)


danielmiessler:
Queued behind the fix backlog: additive and read-only, so the risk is low, but 1.5K lines is an ownership decision, and transcript querying overlaps surfaces we already ship (Pulse, Cortex recall)—the audit session will decide take-as-is vs take-the-core. Status here as soon as it runs.

danielmiessler:
Declining this one, and not on quality — the symlink handling is genuinely good work. The `lstatSync` confinement with two adversarial tests writing a fixture outside the root and asserting it's excluded is better than what a lot of our own tooling does, and we noticed.

The reason is scope. Transcripts are the most sensitive data in the system, and we already have a recall subsystem that answers "find prior work" — adding a second 1,500-line query surface over the same data means two things to maintain, two security stories, and two places for that boundary to drift. One memory story is a deliberate constraint here, not an oversight.

This would make an excellent standalone tool, and we'd link it. If you'd rather aim the effort at core, the recall path is where a transcript-query capability would actually compound.

PR #1876, opened 2026-08-16
Atlas: Linux path unit for immediate event-triggered ticks (parity with darwin WatchPaths)

schmetti-dev:
## What

Linux's `InstallAtlas.ts` had a capability gap against darwin: the plist gets `WatchPaths` on `events.jsonl` for an immediate targeted sync when a hint lands, but linux only had the 15-minute timer — a captured event could sit for up to 15 minutes before Atlas picked it up.

Adds `com.lifeos.atlas.path.template`, a systemd --user `.path` unit watching `events.jsonl` specifically (not the whole Atlas state directory — watching the directory would self-trigger a restart loop, since Atlas's own tick rewrites `db`/`wal`/`snapshot.json` on every run). `InstallAtlas.ts` now materializes, enables, and tears down the path unit alongside the timer, and runs one manual tick on install since `.path` units don't fire on enable the way darwin's `RunAtLoad` does.

Since the path unit means the service can now fire several times in quick succession during a legitimate burst (not just on the 15-min timer), this also adds:
- `StartLimitIntervalSec=60` / `StartLimitBurst=20` on `com.lifeos.atlas.service.template`, so a normal burst doesn't trip systemd's default 5-starts/10s limit into a hard `start-limit-hit` failure (darwin's `ThrottleInterval=60` only delays extra runs, it doesn't hard-fail)
- `statusLinux()`'s exit code now reflects the path unit's health too, not just the timer's
- `collectors/Systemd.ts` now recognizes `.path` units, so Atlas's own asset graph has a record of the unit this same change installs

## Why draft

Opening as draft for a maintainer look before marking ready — this is my first PR touching the Linux install path specifically, want a sanity check on the systemd conventions before finalizing.

## Testing

- `bun build` on both changed `.ts` files (clean)
- Manually verified the existing (pre-this-PR) equivalent has been running on my own box for a while without the restart-loop the comment describes — this PR ports that same design into the template-based installer structure this repo now uses

### DIFF
--- a/LifeOS/install/LIFEOS/ATLAS/InstallAtlas.ts
+++ b/LifeOS/install/LIFEOS/ATLAS/InstallAtlas.ts
@@ -12,10 +12,15 @@
  *     ~/Library/LaunchAgents/com.lifeos.atlas.plist, runs `launchctl bootstrap`.
  *     15-minute StartInterval plus WatchPaths on the hint file, so a mutation
  *     captured by AtlasEventCapture.hook.ts triggers a targeted sync at once.
- *   linux: substitutes into com.lifeos.atlas.{service,timer}.template, writes them
- *     to ~/.config/systemd/user/, enables the timer with `systemctl --user`, and
- *     runs `loginctl enable-linger` so it survives logout. systemd has no WatchPaths
- *     equivalent here — hints wait for the next tick, which only costs latency.
+ *   linux: substitutes into com.lifeos.atlas.{service,timer,path}.template, writes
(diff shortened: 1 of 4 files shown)


danielmiessler:
Ported with credit, darwin branch verified byte-untouched—additive Linux parity with contributor verification is exactly the platform-support shape we take, and your #1743 already established the trust. The path-unit-vs-directory-watch reasoning in the template comment is excellent. Closing per the port flow (your work now spans the whole Atlas Linux lane). Thanks.

PR #1878, opened 2026-08-16
fix(PromptProcessing): look for transcripts under claudeDir, not LIFEOS

abseg:
`findSessionJsonl` searches `paiPath('projects')`. `paiPath()` resolves against `getLifeosDir()`, so that path is `<claudeDir>/LIFEOS/projects` — but Claude Code writes transcripts to `<claudeDir>/projects`. The directory never exists, `existsSync` skips both candidates, and the function always returns `null`.

Both callers then fail silently:

- **`syncNameToJsonl()`** never appends the `custom-title` row, so the name LifeOS derives from the first prompt never reaches the transcript. `/sessions` and the harness surfaces show the raw first prompt instead of the session name.
- **`getCustomTitle()`** always returns `null`, so a name the user sets with `/rename` is never read back as canonical on the next prompt.

### Evidence

On one install, before the fix: **zero `custom-title` rows across all 64 transcripts**, while `session-names.json` held a correct name for every session. After the fix, the row is written on the next prompt and the real name shows up.

```
paiPath('projects')             -> ~/.claude/LIFEOS/projects   (does not exist)
join(getClaudeDir(), 'projects') -> ~/.claude/projects          (64 .jsonl files)
```

### The change

Two lines: import `getClaudeDir` alongside `paiPath`, and build the search roots from it. `PromptProcessing.hook.ts` is the only caller of `paiPath('projects')` in the payload, so nothing else is affected. The `Projects` capitalisation fallback is kept as-is.

### DIFF
--- a/LifeOS/install/hooks/PromptProcessing.hook.ts
+++ b/LifeOS/install/hooks/PromptProcessing.hook.ts
@@ -43,7 +43,7 @@ import { inference } from '../LIFEOS/TOOLS/Inference';
 import { getIdentity, getPrincipal } from './lib/identity';
 import { isValidWorkingTitle, getWorkingFallback, trimToValidTitle } from './lib/output-validators';
 import { getSessionOneWord, readTabState, setAscentTab } from './lib/tab-setter';
-import { paiPath } from './lib/paths';
+import { paiPath, getClaudeDir } from './lib/paths';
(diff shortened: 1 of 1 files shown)


danielmiessler:
Ported with credit: transcripts resolve under the harness dir, and your evidence line (zero custom-title rows across 64 transcripts) reproduced exactly on the maintainer install—the bug was live everywhere. Closing per the port flow. Thanks.

PR #1884, opened 2026-08-17
fix(Doctor): bound the shadow-$HOME search instead of walking all of $HOME

abseg:
`findShadowHomeDirs` runs:

```
find $HOME -maxdepth 6 ( -name node_modules -o -name .git -o -name Library ) -prune -o -type d -name '$HOME' -print
```

Three prunes, and everything else in the home directory is fair game. Every `doctor` run descends into **Desktop, Documents, Downloads, Movies and Pictures**.

On macOS that raises a TCC file-access prompt for those folders, attributed to the runtime rather than to LifeOS, because `find` is its child. From the user's side a routine health check asks for their Videos folder with no explanation. That is what prompted this patch.

### Why the width was never needed

Shadow-$HOME trees are created where a **session ran**. So the search space is the config root plus the user's project directories — not the whole home folder.

The bounded version is also **strictly more accurate**: a project nested deeper than six levels below home is invisible to the current `-maxdepth 6` walk from `$HOME`, and is found by this one, because the depth budget now starts at each project root.

### The change

`shadowSearchRoots()` collects the config root, conventional code directories that exist (`Projects`, `code`, `Developer`, `src`), and any `base_path` in the project registry. `findShadowHomeDirs()` runs the same `find` per root and dedupes.

Roots that live inside a TCC-protected folder (a registered project under `~/Documents`, say) are **skipped and reported on stderr**, with the command to search one deliberately — a silently narrowed search reads as "covered everything" when it did not.

### Verified

Recorded `atime` on `~/Desktop`, `~/Movies` and `~/Pictures`, ran `doctor`, re-read them: unchanged. The capability still reports correctly (`no shadow-$HOME trees on disk`), and the skipped root is named in the output.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/Doctor.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/Doctor.ts
@@ -144,16 +144,76 @@ function envKey(name: string): string | null {
 
 const RECLAIM_BACKUP_DIR = join(STATE_DIR, 'reclaimed-shadow-backups');
 
+/**
+ * Roots worth searching for shadow-$HOME trees.
+ *
+ * NOT `$HOME`. Upstream walks the bare home directory six levels deep, pruning only
+ * node_modules/.git/Library, which means every `doctor` run descends into Desktop,
+ * Movies, Pictures, Documents and Downloads. On macOS that raises a TCC prompt naming
+ * those folders and attributing it to `bun` (find is bun's child) — which is exactly what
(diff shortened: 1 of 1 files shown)


danielmiessler:
Ported adapted: the bounded-search shape is yours (config root + conventional code dirs + registry + TCC skip-with-notice); the registry source differs on our side (`PROJECTS.md` table rather than project-registry.yaml), noted in the credit comment. The TCC prompt attribution detail was the convincer. Closing per the port flow. Thanks.

Issue #1559, opened 2026-07-21
ContextAudit.ts reports false confidence: placeholder blindness, cross-ref false positives, and a critical check blind to most sections


tzioup:
> Structured for machine parsing: claims are atomic, each carries the command that verifies it. Every command runs against a fresh clone of this repo — none depend on my machine, my install, or my configuration. Full reproduction at the bottom.

## Summary

`LIFEOS/TOOLS/ContextAudit.ts` ships in the install payload and is the content-quality gate invoked at the end of the `/interview` context check-in (`skills/Interview/Workflows/ContextCheckin.md` Step 4). Its own header calls it a "read-only **quality audit** for constitutional context files."

Its checks are actually reference/placeholder hygiene, and several are unsound in ways that make its "clean" verdict untrustworthy — a validation tool that produces **both false negatives and false positives** doesn't just miss bugs, it launders a broken state as a passing one. Five independent issues, each verifiable from a fresh clone:

1. **False negative — the placeholder check cannot see the markers the templates actually ship.** `scanPlaceholders` looks for `<your-name>`, `<TBD>`, `<...>`, `(seeded during interview)`; the shipped templates use `(sample)` and `(interview)`. A 100%-unfilled template file scores zero placeholder findings.
2. **False positive — the cross-reference check treats any backtick span as a literal path.** A span like `` `LIFEOS/TOOLS/MemorySystem.ts add(item)` `` is `existsSync`-checked whole and warned as broken, though the file exists; symlink notation `` `A → B` `` is warned as a missing file.
3. **The tool's own `critical` empty-section check is blind to most sections.** `scanPrincipalTelos` fires only on three exact heading strings and uses exact-match heading resolution, so a one-character variant or any other heading (including `Core Models`, which is empty by construction — see #4) is never flagged critical.
4. **The generator emits empty sections unconditionally, and excludes two of them from its own fail-loud guard.** `GenerateTelosSummary.ts` emits `## Core Models` (and 6 other headings) with no `if (array.length > 0)` guard, while its per-section `coreChecks` safety net omits `narratives` and `models`.
5. **The audit writes to a hardcoded, dated, personal-looking output path** rather than a stable location.

The `#1`/`#3`/`#4` chain is the sharp one: the generator reliably produces an empty `## Core Models`, and the audit's dedicated critical check is structurally unable to catch it.

## Claims and verification

```bash
git clone --depth 1 https://github.com/danielmiessler/LifeOS && cd LifeOS
[...]

PR #1887, opened 2026-08-17
fix(telos): guard the six unconditional PRINCIPAL_TELOS sections

### LINKED ISSUE #1886: GenerateTelosSummary emits six unconditional empty sections that ContextAudit reports as critical — on every fresh install

## TL;DR

`GenerateTelosSummary.ts` guards five of its output sections against being emitted empty, and leaves six unguarded. On a fresh install every TELOS source is still a shipped template, so the generator writes six empty sections and `ContextAudit.ts` immediately reports each as `critical`. The generator produces a file its own auditor condemns, by default, before the user has done anything wrong.

## Evidence

Issue #1559 added a guard around Core Models for exactly this reason, and the source comment states it plainly:

```ts
// Emit Core Models only when populated (public issue #1559, @tzioup): an
// unconditional empty section is exactly the shape ContextAudit's empty-
// section check exists to catch, and emitting it guaranteed a false-clean.
if (models.length > 0) {
  lines.push('', '## Core Models', '', ...models);
}
```

That guard was applied to five sections and not to the rest:

| Guarded | Unguarded |
|---|---|
| Active Projects, Formative Experiences, Things I've Been Wrong About, Core Models, Wisdom | **Missions, Active Goals, Problems Being Solved, Strategies, Active Narratives, Personal Challenges** |

`ContextAudit.ts` `scanPrincipalTelos()` checks nine headings and emits `severity: "critical"` for any that is present and empty. All six unguarded sections are in that list.

## Impact

Measured on 7.40.4 with stock template TELOS sources: `PRINCIPAL_TELOS.md` reports **6 critical and 6 warn** findings. After guarding the six sections, the same file on the same sources reports **0 critical, 0 warn, 0 info**.

This is the default state of a new install, not an edge case. The generator prints `TELOS sources are still shipped templates — writing an empty summary until they are filled in` on the same run that produces the six criticals.

Same class as #1732 (IntegrityCheck opening red on a stock install): a check that is red out of the box teaches users to ignore it.

## Reproduction

1. On an install whose TELOS sources are still the shipped templates, run `bun LIFEOS/TOOLS/GenerateTelosSummary.ts`.
2. Observe `## Missions`, `## Active Goals (2026)`, `## Problems Being Solved`, `## Strategies`, `## Active Narratives` and `## Personal Challenges` emitted with no content.
3. Run `bun LIFEOS/TOOLS/ContextAudit.ts`. `PRINCIPAL_TELOS.md` reports 6 critical.

## Suggested fix

Apply the #1559 guard to the six remaining sections. Two need a slightly wider condition than a bare length check: **Active Goals** must survive when `goals.active` is empty but `deferred`/`completed` have content, since those lines render below the heading; **Active Narratives** must survive when `primary` is empty but `secondary` is not.

## Secondary observation, not part of the fix

`ContextAudit.ts` emits this message for every empty section:

```ts
message: `## ${heading} section is empty - generator bug (likely LEGACY_FILE_TO_SECTION case mismatch in GenerateTelosSummary.ts)`
```

The cause is hardcoded rather than computed, and a case mismatch is not reachable: `loadTelosSections()` stores every heading via `title.toLowerCase()` and `LEGACY_FILE_TO_SECTION` values are already lowercase, so both sides of the lookup are lowercased. A `critical` finding naming a specific constant and file directs debugging at code that is behaving correctly. Kept out of the fix because it is a different file and a different concern; happy to open it separately if useful.

## Environment

LifeOS 7.40.4, `main` at `ce046f2`. macOS, bun 1.3.14.

bnkath2o:

Fixes #1886.

#1559 added `if (models.length > 0)` around Core Models because an unconditional empty section is exactly what `ContextAudit`'s empty-section check reports as `critical`. That guard reached five sections and not the other six.

**What changed.** `Missions`, `Active Goals`, `Problems Being Solved`, `Strategies`, `Active Narratives` and `Personal Challenges` are now emitted only when they have content. One file, +31/−30.

**Two conditions are deliberately wider than a bare length check.** `Active Goals` is emitted when `active`, `deferred` **or** `completed` has content, because the deferred/completed lines render beneath that heading and would otherwise appear under no heading at all. `Active Narratives` is emitted when `primary` **or** `secondary` has content, for the same reason.

**Verified both directions against 7.40.4 (`ce046f2`).**

- With stock template sources: `PRINCIPAL_TELOS.md` goes from **6 critical / 6 warn to 0/0/0** in `ContextAudit`, and emits no empty headings.
- With `MISSION.md` and `PROBLEMS.md` populated in a `HOME`-sandboxed tree: `## Missions` and `## Problems Being Solved` render normally while the still-empty sections stay absent.

The second check is the one that matters for review — it shows the guards suppress empty sections without suppressing content.

No behaviour change for an install whose TELOS is filled in.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/GenerateTelosSummary.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/GenerateTelosSummary.ts
@@ -693,15 +693,23 @@ function generate(): string {
     '',
     '> Auto-generated from TELOS source files. Do not edit manually.',
     `> Generated: ${now} | Sources: MISSION, GOALS, PROBLEMS, STRATEGIES, PROJECTS, NARRATIVES, CHALLENGES, WRONG, TRAUMAS, MODELS, WISDOM`,
-    '',
-    '## Missions',
-    '',
-    ...missions,
-    '',
-    '## Active Goals (2026)',
-    '',
-    ...goals.active,
   ];
+
+  // Same guard as Core Models and Active Projects below (public issue #1559): an
+  // unconditional empty section is exactly the shape ContextAudit's empty-section
+  // check reports as `critical`, so emitting one makes the generator produce a file
+  // its own auditor condemns. On a fresh install every TELOS source is still a
+  // template, which is precisely when these are empty.
+  if (missions.length > 0) {
+    lines.push('', '## Missions', '', ...missions);
+  }
+
+  // Deferred/completed are rendered below as their own lines, so the heading must
+  // survive when active is empty but either of those has content — otherwise those
+  // lines would appear under no heading at all.
+  if (goals.active.length > 0 || goals.deferred.length > 0 || goals.completed.length > 0) {
+    lines.push('', '## Active Goals (2026)', '', ...goals.active);
+  }
 
   if (goals.deferred.length > 0) {
     // Compress deferred goals to a single inline line — they're not active and don't need full bullets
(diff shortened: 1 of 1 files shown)


danielmiessler:
Verified against our source and taken. Thank you.

All six sections were emitting unconditionally in our tree too, so this was live for every install, not just fresh ones — it only became visible on installs where the TELOS sources are still templates. The fix finishes the pattern #1559 started on Projects and Core Models, which is what makes it obviously right rather than a special case.

Checked before and after on a simulated fresh install: six empty sections before, zero after, and a populated TELOS renders identically.

Ported into our source (the public repo is generated from it, so it lands in the next release rather than as a merge here).

PR #1889, opened 2026-08-17
fix(SatisfactionCapture): imperatives after a bare digit are not ratings (follow-up to #1670)

### LINKED ISSUE #1888: SatisfactionCapture: the short bare-digit case #1670 left — '1 go' parses as rating 1 and writes a false failure capture

## TL;DR

This is the residual case **#1670 named and could not fix by length**. `parseExplicitRating()` reads a bare leading digit followed by a short imperative as a satisfaction rating, so a user answering a numbered question with `1 go` is recorded as **rating 1**. Ratings ≤ 3 also write a permanent failure capture, so the learning loop trains on a complaint that was never made.

## Relationship to the closed work

- **#1670** (closed 2026-07-29) fixed the list-marker shape (`1)`, `1]`) and capped the bare-digit comment at `MAX_BARE_DIGIT_COMMENT = 80`. It stated the remaining problem explicitly: *"`-` is an accepted rating separator, so `2 - <prose>` is shape-identical to `8 - nice`. Only comment length separates them."*
- **#1703** (closed 2026-08-06) extended both guards to reject non-ASCII lead-in text.

The short form slips under both. `1 go` is two characters, so the 80-char cap never sees it, and it is ASCII. #1670 was right that **length** cannot separate `1 go` from `8 nice` — they are the same shape and the same size. What separates them is **kind**: a rating comment evaluates the work, an answer to a numbered question commands it.

## Evidence

Against unmodified 7.40.4:

| input | current result |
|---|---|
| `1 go` | **rating=1**, comment `"go"` |
| `1, yes fix properly` | **rating=1** |
| `1 yes chase` | **rating=1** |
| `2 skip that` | **rating=2** |
| `1) go to fix` | null — #1670 guard holds |
| `1. go` | null — holds |
| `1 of the files failed` | null — `SENTENCE_STARTERS` holds |
| `8 nice work` | rating=8 — correct |

## Impact

Measured on one install's `SIGNALS/ratings.jsonl`: **40 of 152 rows carry `rating: 1`**, the second-largest bucket after `5`. Of the 31 with comment text the comments read `") go to fix"`, `") yes"`, `") Go"`, `"go"`, `"yes"`, `") approved"` — approvals and instructions, not evaluations. Six failure captures were written in a single day, one for a turn in which the user approved a fix.

Because ratings ≤ 3 feed the failure capture, the recurrence ledger and pattern synthesis, one misparse contaminates three surfaces — the same multiplication #1670 documented.

Reachable by any assistant that labels its questions numerically and any user who answers with the label. It does not depend on TELOS state and reproduces against a byte-identical copy of the shipped file with no local modification and no custom override.

## Reproduction

```ts
const { parseExplicitRating } = await import("./hooks/SatisfactionCapture.hook.ts");
parseExplicitRating("1 go");          // → { rating: 1, comment: "go" }
parseExplicitRating("2 skip that");   // → { rating: 2, comment: "skip that" }
parseExplicitRating("1) go to fix");  // → null   (already guarded by #1670)
```

## Suggested fix

Add `IMPERATIVE_STARTERS`, the verb-shaped counterpart to the existing `SENTENCE_STARTERS`, applied **only** to the bare-digit path. `SENTENCE_STARTERS` already draws exactly this line for nouns (`1 of the files`); this draws it for verbs.

### Validation, since this suppresses ratings and the dangerous direction is losing real complaints

A/B of the current and patched parser over the full 152-row historical ledger:

- **138 rows unchanged.**
- **14 rows changed, every one `1 → null`.**
- Every one of those 14 was **independently** flagged as a false positive by a separate pass using different criteria (approval-opener matching rather than imperative matching). Two methods, same rows.
- **No rating ≥ 2 changed.** Real low ratings still parse: `3 too slow` → 3, `2 sloppy output` → 2.
- 17/17 hand cases correct; the `1)`, `1.`, non-ASCII and `SENTENCE_STARTERS` paths are untouched.
- Diff is **+24 lines, 0 deletions**, and does not touch the `N/10` or word-form parsers.

### Two limitations, stated rather than discovered later

1. **One-directional by design.** `1 fix your formatting` is a genuine complaint this will now miss. That trade follows the function's own stated preference: a miss loses a ledger row, a false positive writes a permanent failure record.
2. **Partial.** A verb outside the list still misparses — `2 re-register` remains `rating=2`. This narrows the dominant forms; it does not claim to solve the general case, which #1670 correctly judged unsolvable syntactically.

## Environment

LifeOS 7.40.4, `main` at `ce046f2`. `hooks/SatisfactionCapture.hook.ts` byte-identical to the shipped payload.

bnkath2o:

Fixes #1888. Follow-up to #1670, which fixed the list-marker and long-prose shapes and named this residual: *"`-` is an accepted rating separator, so `2 - <prose>` is shape-identical to `8 - nice`. Only comment length separates them."*

The short form slips under that cap. `1 go` is two characters, so `MAX_BARE_DIGIT_COMMENT = 80` never sees it. #1670 was right that **length** cannot separate `1 go` from `8 nice`; what separates them is **kind** — a rating comment evaluates the work, an answer to a numbered question commands it.

**What changed.** `IMPERATIVE_STARTERS`, the verb-shaped counterpart to the existing `SENTENCE_STARTERS`, applied **only** to the bare-digit path. `SENTENCE_STARTERS` already draws this line for nouns (`1 of the files`); this draws it for verbs. +24 lines, 0 deletions, one file.

## Validation

This suppresses ratings, so the risk worth testing is losing real complaints. A/B of the current and patched parser over a 152-row historical ledger:

- **138 rows unchanged.**
- **14 changed, every one `1 → null`.**
- All 14 were **independently** flagged as false positives by a separate pass using different criteria (approval-opener matching, not imperative matching). Two methods, same rows.
- **No rating ≥ 2 changed.** Real low ratings still parse: `3 too slow` → 3, `2 sloppy output` → 2.
- 17/17 hand cases correct. The `1)` / `1]` guard from #1670, the non-ASCII guard from #1703, the `SENTENCE_STARTERS` guard, the `N/10` parser and the word-form parser are all untouched.

## Two limitations, stated up front

1. **One-directional by design.** `1 fix your formatting` is a genuine complaint this now misses. That follows the function's own documented preference — a miss loses a ledger row, a false positive writes a permanent failure record.
2. **Partial.** A verb outside the list still misparses (`2 re-register` → rating 2). This narrows the dominant forms rather than claiming to solve the general case, which #1670 correctly judged unsolvable syntactically.

Happy to widen the token list or take a different discriminator if you'd prefer.

### DIFF
--- a/LifeOS/install/hooks/SatisfactionCapture.hook.ts
+++ b/LifeOS/install/hooks/SatisfactionCapture.hook.ts
@@ -93,6 +93,13 @@ const MAX_BARE_DIGIT_COMMENT = 80;
 // (e.g. "2/10 items done", "3 of the files"). Shared by the fraction and generic parsers.
 const SENTENCE_STARTERS = /^(items?|things?|steps?|files?|lines?|bugs?|issues?|errors?|times?|minutes?|hours?|days?|seconds?|percent|%|th\b|st\b|nd\b|rd\b|of\b|in\b|at\b|to\b|the\b|a\b|an\b)/i;
 
+// Verb-shaped counterpart to SENTENCE_STARTERS, for the bare-digit path only.
+// A leading digit followed by an imperative is a user answering a numbered
+// question ("1 go", "2 re-register"), not rating the work: a rating comment
+// evaluates, an answer commands. Leading punctuation is skipped so the common
+// "1, yes fix properly" and "1 - go" forms are covered too.
+const IMPERATIVE_STARTERS = /^[,;:\-\s]*\s*(go\b|yes\b|no\b|ok\b|okay\b|sure\b|do\b|does\b|don'?t\b|proceed\b|continue\b|approved?\b|confirm\w*\b|agreed?\b|correct\b|commit\b|push\b|run\b|fix\b|build\b|add\b|remove\b|delete\b|skip\b|defer\b|park\b|hold\b|stop\b|wait\b|turn on\b|turn off\b|enable\b|disable\b|file it\b|send it\b|ship it\b|make\b|use\b|try\b|check\b|show\b|tell\b|give\b|start\b|apply\b|revert\b|retry\b)/i;
+
 // ── Stdin Reader ──
 
 async function readStdinWithTimeout(timeout: number = 5000): Promise<string> {
@@ -191,6 +198,23 @@ export function parseExplicitRating(prompt: string): { rating: number; comment?:
   // unambiguous rating syntax and are deliberately NOT capped.
   // public PR #1670, @asdf8675309
   if (rest && rest.length > MAX_BARE_DIGIT_COMMENT) return null;
+
+  // The short-comment case #1670 could not separate by length. "1 go" and "8 nice"
+  // are the same shape and both well under the cap, so the cap never sees them —
+  // but they are not the same KIND of text. A rating comment evaluates the work;
+  // an answer to a numbered question commands it. SENTENCE_STARTERS already draws
+  // this line for nouns ("1 of the files"); this is the verb-shaped counterpart.
+  //
+  // Observed on one install: 40 of 152 ledger rows carried rating 1, the second
+  // largest bucket, and the comments read "go", "yes", ") approved". Ratings <= 3
+  // also write a permanent failure capture, so each one recorded a complaint that
+  // was never made and fed it to the learning loop.
+  //
+  // Deliberately one-directional, per this function's stated preference for a
+  // missed rating over a false positive: "1 fix your formatting" is a real
+  // complaint that this will now miss. That trade is the documented one — a miss
+  // loses a ledger row, a false positive writes a permanent failure record.
+  if (rest && IMPERATIVE_STARTERS.test(rest)) return null;
 
   return { rating, comment: rest };
 }


danielmiessler:
Taken, with one honest note.

The bug is real in our tree. Our own ledger has 2,693 rows and exactly one false positive of this shape, where yours had 40 of 152 — the difference is just how often the user answers a numbered question with a bare digit. The severity argument still holds regardless of rate: a rating of 3 or below writes a permanent failure capture, so each false row feeds a complaint to the learning loop that nobody made.

The framing is what sold it. SENTENCE_STARTERS already draws the noun-shaped version of this line, so the verb-shaped counterpart is a completed pattern rather than a new mechanism, and the one-directional trade is the one this function already documents.

One gap worth flagging: the comment cites `2 re-register` as a motivating example, but the regex does not match it — `re-register` is not in the list and nothing covers the `re-` prefix generally. We left it uncovered rather than widening the pattern, since a miss costs a ledger row and a false positive costs a failure record. Mentioning it so the example and the behavior are not read as agreeing.

Added regression tests for the covered forms and for the cases that must keep parsing (`8 nice`, `9 solid work`, `2 sloppy`, bare `1`). Ported into our source; it lands in the next release.

PR #1894, opened 2026-08-18
fix(TokenXray): dedupe API calls by requestId

asdf8675309:
## Problem

`TokenXray` sums every `usage` block in `~/.claude/projects/*/*.jsonl`. Claude Code
writes the same assistant message into more than one transcript file. Sidechains,
resumed sessions, and worktree copies all repeat the same record. The tool counts
one API call several times.

On one machine the transcripts hold 292,757 `requestId` values. Only 139,447 are
unique. That is a duplication factor of 2.10.

The effect is visible against `ccusage`, which reads the same files and dedupes on
`requestId`. For the same machine:

| Tool | Billed input |
|---|---|
| `TokenXray split` | ~100.6B |
| `ccusage` | ~56.8B |

The same cause makes two `TokenXray` subcommands disagree with each other.
`split` reports 94.4B cache-read tokens. `reread` reports 82.1B billed input and
labels it "measured, exact". The two subcommands accumulate at different sites, so
the duplicate records land unevenly.

## Change

This PR adds `makeDedup()` after `readJsonl()`. The function returns a closure. The
closure holds a `Set` of `requestId` values and reports whether an entry is the
first sighting. Each subcommand creates its own set.

Entries without a `requestId` cannot be keyed. The filter keeps them. No record is
dropped in silence.

The guard is applied at three accumulation sites:

- `cost`, on the `mainLogs()` billed totals
- `split`, on the main totals and the sidecar totals
- `reread`, on the `measured` accumulator

## Limitation

The unique-content tallies in `breakdown` and `reread` are not deduped. Those
entries carry no `requestId`. A duplicated transcript file still counts its text
twice. Billed-token figures are correct after this change. Unique-content figures
stay approximate.

## Verification

- A grep of the original file for `requestId`, `messageId`, `dedup`, and `seen.`
  returns one unrelated `Set<string>`. No dedup existed.
- The duplication factor comes from
  `find . -name "*.jsonl" -print0 | xargs -0 grep -ho '"requestId":"[^"]*"'`,
  compared against `sort -u`.
- `split` was run against the pre-change file and the post-change file on the same
  corpus.

## Why it matters

`TokenXray` is the token counter a LifeOS user reaches for. A user who compares it
against `ccusage` sees two different answers and cannot tell which one to trust.

## Measured result

Same corpus, 28,844 files, `split` COMBINED:

| Metric | Before | After |
|---|---:|---:|
| assistant calls | 398,830 | 246,718 |
| cache read | 94.91B | 54.82B |
| cache write | 5.87B | 3.63B |
| output | 401.4M | 213.1M |

`ccusage` reports about 56.8B total tokens for the same machine. After this
change TokenXray totals about 59.0B. The two agree to within about 4%. Before
the change the gap was about 1.8x.

The call count falls by 1.62x, not by the 2.10x duplication factor. Entries
without a `requestId` are kept, as described above.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/TokenXray/TokenXray.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/TokenXray/TokenXray.ts
@@ -104,6 +104,23 @@ function* readJsonl(path: string): IterableIterator<Record<string, unknown>> {
   }
 }
 
+// Claude Code writes the same assistant message into several transcript files
+// (sidechains, resumed sessions, worktree copies), so summing every `usage`
+// block counts one API call many times. Measured on a real corpus: 292,757
+// requestId occurrences for 139,447 unique calls, a 2.10x inflation that put
+// billed input ~1.8x above ccusage, which dedupes on the same key.
+// Entries without a requestId cannot be keyed, so they are kept.
+function makeDedup(): (o: Record<string, unknown>) => boolean {
+  const seen = new Set<string>();
+  return (o) => {
+    const id = o.requestId;
+    if (typeof id !== "string" || id === "") return true;
+    if (seen.has(id)) return false;
+    seen.add(id);
+    return true;
+  };
+}
+
 function mainLogs(): string[] {
   return glob(join(homedir(), ".claude/projects/*/*.jsonl"));
 }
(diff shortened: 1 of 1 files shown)


danielmiessler:
Ported, and your diagnosis was right — the totals really were inflated, and `requestId` is the right key. Measured on this machine: 2.16x over-count on a 400-file sample.

The implementation deviated from your diff, and you should know why. A first-wins `Set` is correct if the duplicate records are copies. If any of them are streaming partials — the same `requestId` appearing several times with a growing `output_tokens` — then counting the first record and skipping the rest trades an over-count for an under-count. Our two measurements actually disagreed about how common that is, which is precisely why we didn't want the answer to depend on it.

So the port counts a per-`requestId` high-water mark instead: each record contributes only the part of a value not already counted for its call. That's identical to first-wins when the records are copies, and correct when they're partials. Same key, same per-call ledger scoping you used, same treatment of records with no `requestId` (counted in full, never dropped).

Process note: the public repo is generated from a private tree, so PRs are ported into source with credit rather than merged, which is why you'll see the fix rather than the merge. Ships in the next release.

PR #1895, opened 2026-08-18
fix(KnowledgeSchema): parseNote rejects legal YAML at column 0

asdf8675309:
## Problem

`parseNote` marks two legal YAML shapes as malformed. Both occur at column 0.

- An unindented block-sequence item under a key, for example `- slug: foo` beneath `related:`
- A standalone comment line, for example `# (ripple-update: linked X)`

`KEY_LINE` requires a letter or underscore at the start. `isContinuation` requires
leading whitespace. Neither shape satisfies either rule, so both fall into the orphan
branch that sets `malformed = true`.

Run against the current `main`:

| Input | `main` | this branch |
|---|---|---|
| unindented sequence | `malformed=true` | `malformed=false` |
| standalone comment | `malformed=true` | `malformed=false` |

The effect is that a healthy note reads as damaged. On one archive this misreported 13
notes. A migrator that trusts the flag will skip or rewrite files that were never broken.

## Change

`parseNote` now recognises both shapes.

A column-0 sequence item joins the open block above it. YAML permits a sequence at its
parent indentation.

A standalone comment becomes its own field under a reserved key. It is not merged into
the scalar above it. Merging would clear that field's `scalar`, and `normalize` would
then treat the field as absent, re-mint it, and drop the comment text.

The change also flags a block that carries both column-0 and indented sequence items at
the same level. Each indentation is legal alone. Together no YAML reader can load the
block, so the parser reports it rather than let a migrator stamp `convention: kb-v3` on a
file nothing can read. A genuinely nested sequence stays legal and is not flagged.

## Tests

`bun KnowledgeSchema.ts test` goes from 29 assertions to 47. The 18 new assertions cover
the unindented sequence, the standalone comment, a comment inside a block, and the mixed
indentation case. Each new case also asserts a byte-identical round trip through
`serializeNote`, so the parser cannot pass by discarding the text it failed to classify.

Both suites pass: 29 of 29 on `main`, 47 of 47 on this branch.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/KnowledgeSchema.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/KnowledgeSchema.ts
@@ -165,6 +165,27 @@ export interface ParsedNote {
 }
 
 const KEY_LINE = /^([A-Za-z_][\w-]*):(.*)$/;
+const FLUSH_SEQ_ITEM = /^-(?:\s|$)/;
+const INDENTED_SEQ_ITEM = /^\s+-\s/;
+/** Key for a standalone `# comment` line. NUL can't occur in a real YAML key. */
+const STANDALONE_COMMENT_KEY = "\0comment";
+
+/**
+ * True when appending `line` would give one block BOTH column-0 and same-level
+ * indented sequence items. Each indentation is legal alone; together no YAML
+ * reader can load the block. A genuinely NESTED sequence is legal and must not
+ * be flagged — its parent key opens a block, so the preceding line is a bare
+ * `key:` with no value.
+ */
+function mixesSequenceIndentation(raw: string, line: string): boolean {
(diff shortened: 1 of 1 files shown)


danielmiessler:
Ported and credited. The defect is exactly as you described: `KEY_LINE` requires a letter or underscore at column 0 and `isContinuation` requires leading whitespace, so an unindented block-sequence item under `related:` and a standalone comment line satisfy neither and fall into the orphan branch that sets `malformed`. Both are legal YAML.

One process note so it isn't a surprise: the public repo is generated from a private source tree, so community PRs get ported into source with credit rather than merged — same fix, but it survives the next release instead of being overwritten by it. The fix site carries your handle and this PR number.

Round-tripping is covered by tests here too. Thanks — ships in the next release.

PR #1918, opened 2026-08-20
fix(paths): resolve the install root from CLAUDE_CONFIG_DIR / LIFEOS_DIR

PigLardLord:
## Problem

LifeOS assumes it lives at `$HOME/.claude`. The harness does not — it exposes
`CLAUDE_CONFIG_DIR` precisely so the config dir can live elsewhere. Anyone who uses it
(a second install kept off a `~/.claude` that an employer syncs, a multi-profile setup,
a non-default layout) gets ~165 call sites resolving `join(homedir(), ".claude", …)`
against a directory that is not their install.

Two failure classes, and the first is the nasty one:

**1. Silent wrong-tree reads.** `settings.json`, `skills/`, `hooks/`, `agents/`,
`projects/` and `CLAUDE.md` exist in *both* trees. The lookup succeeds — against the
wrong file. Nothing throws. Observed: the Pulse wiki and freshness tabs inventory a
different install's skills and hooks; `GetCounts`, `GenerateTelosSummary` and
`observability` read a `settings.json` that is not the live one; `AgentInvocation` reads
agent frontmatter from an unrelated `agents/`; memory harvesting walks the wrong
`projects/` transcripts.

**2. Install-time baked paths.** The `Install*.ts` tools materialize launchd plists with
literal `$HOME/.claude/...` program args and log paths, and never export `LIFEOS_DIR` or
`CLAUDE_CONFIG_DIR` in `EnvironmentVariables`. So even with the code fixed a daemon
re-derives the wrong root at boot, because the environment it actually runs under is
empty. Worth stressing: fixing the code alone was **not** enough downstream — the plists
had to carry the env too.

The codebase already knows the right pattern; it is just applied unevenly. These are
correct today:

```ts
// PULSE/Conduit/paths.ts
export const CLAUDE_ROOT = process.env.CLAUDE_CONFIG_DIR || join(homedir(), ".claude");
// PULSE/checks/notification-governor.ts, poller-meta-monitor.ts
const LIFEOS_DIR = process.env.LIFEOS_DIR || join(HOME, ".claude", "LIFEOS");
```

while the shared resolver every hook imports is not.

## Fix

**1. `hooks/lib/paths.ts` — the one that matters.** `getClaudeDir()` honored
`CLAUDE_PLUGIN_ROOT` but not `CLAUDE_CONFIG_DIR`, so `getSettingsPath()`,
`getSkillsDir()`, `getHooksDir()` and `getEnvPath()` pointed at `~/.claude` regardless.
One branch fixes every consumer that goes through the resolver.

```diff
 export function getClaudeDir(): string {
   const pluginRoot = process.env.CLAUDE_PLUGIN_ROOT;

   if (pluginRoot) {
     return expandPath(pluginRoot);
   }

+  const configDir = process.env.CLAUDE_CONFIG_DIR;
+
+  if (configDir) {
+    return expandPath(configDir);
+  }
+
   return join(homedir(), '.claude');
 }
```

and `getLifeosDir()`'s last resort follows it instead of re-hardcoding:

```diff
-  return join(homedir(), '.claude', 'LIFEOS');
+  return join(getClaudeDir(), 'LIFEOS');
```

**2. 167 direct call sites** across `LIFEOS/**` and `hooks/**` that bypass the resolver,
guarded mechanically:

```ts
join(HOME, ".claude", "LIFEOS", …)  ->  join(process.env.LIFEOS_DIR || join(HOME, ".claude", "LIFEOS"), …)
join(HOME, ".claude", "<seg>", …)   ->  join(process.env.CLAUDE_CONFIG_DIR || join(HOME, ".claude"), "<seg>", …)
```

Collapsing these into `paiPath()` / `getClaudeDir()` imports would be nicer; the inline
guard is the low-risk first step and keeps the diff reviewable line by line. Happy to
follow up with the import version if you prefer it.

**3. Installers and templates.** 16 plist templates parametrized (`{{LIFEOS_DIR}}` /
`__LIFEOS_DIR__` and the config-dir equivalents), both roots exported in
`EnvironmentVariables`, and the 14 substitution sites plus the Pulse/MenuBar shell
materializers taught the new placeholders. Shell scripts (`install.sh`, StatusLine, the
Pulse manage scripts, Interceptor helpers) resolve through `CLAUDE_CONFIG_DIR`.

## Compatibility

Both env vars unset → `getClaudeDir()` returns `~/.claude` and every rewritten site
resolves exactly as before. Byte-identical on a default install, no migration.
`CLAUDE_PLUGIN_ROOT` still wins over `CLAUDE_CONFIG_DIR`, so plugin installs are
untouched.

Verified:

| env | `getClaudeDir()` | `getLifeosDir()` |
|---|---|---|
| *(nothing set)* | `~/.claude` | `~/.claude/LIFEOS` |
| `CLAUDE_CONFIG_DIR=/tmp/alt` | `/tmp/alt` | `/tmp/alt/LIFEOS` |
| `+ CLAUDE_PLUGIN_ROOT=/tmp/plug` | `/tmp/plug` | `/tmp/plug/LIFEOS` |

## Checks run

- 98 changed `.ts` files parse clean
- every changed `.sh` passes `bash -n`
- every changed plist passes `plutil -lint`
- residual sweep: 0 unguarded sites in shell and plists, 2 left in `pulse-old.ts`
  (dead code, deliberately untouched)

## Notes from running this downstream

Applied on a relocated install (config dir outside `~/.claude`) before opening this PR:
Pulse healthy after reload, and every runtime path in the daemon logs now resolves to
the real install root instead of the unrelated tree that happened to sit at `~/.claude`.

### DIFF
--- a/LifeOS/install/LIFEOS/ATLAS/InstallAtlas.ts
+++ b/LifeOS/install/LIFEOS/ATLAS/InstallAtlas.ts
@@ -29,18 +29,20 @@ import { join } from "path";
 declare const Bun: { spawn: (cmd: string[], opts?: any) => any };
 
 const HOME = process.env.HOME || "";
+const CLAUDE_CONFIG_DIR = process.env.CLAUDE_CONFIG_DIR ?? join(HOME, ".claude");
+const LIFEOS_DIR = process.env.LIFEOS_DIR ?? join(CLAUDE_CONFIG_DIR, "LIFEOS");
 const LABEL = "com.lifeos.atlas";
 const IS_LINUX = process.platform === "linux";
 const IS_MACOS = process.platform === "darwin";
 
 // darwin (launchd)
-const TEMPLATE_PATH = join(HOME, ".claude", "LIFEOS", "ATLAS", "com.lifeos.atlas.plist.template");
+const TEMPLATE_PATH = join(LIFEOS_DIR, "ATLAS", "com.lifeos.atlas.plist.template");
 const LAUNCH_AGENTS_DIR = join(HOME, "Library", "LaunchAgents");
 const TARGET_PLIST = join(LAUNCH_AGENTS_DIR, `${LABEL}.plist`);
 
 // linux (systemd --user)
-const SERVICE_TEMPLATE_PATH = join(HOME, ".claude", "LIFEOS", "ATLAS", "com.lifeos.atlas.service.template");
-const TIMER_TEMPLATE_PATH = join(HOME, ".claude", "LIFEOS", "ATLAS", "com.lifeos.atlas.timer.template");
+const SERVICE_TEMPLATE_PATH = join(LIFEOS_DIR, "ATLAS", "com.lifeos.atlas.service.template");
+const TIMER_TEMPLATE_PATH = join(LIFEOS_DIR, "ATLAS", "com.lifeos.atlas.timer.template");
 const SYSTEMD_USER_DIR = join(HOME, ".config", "systemd", "user");
 const TARGET_SERVICE = join(SYSTEMD_USER_DIR, `${LABEL}.service`);
 const TARGET_TIMER = join(SYSTEMD_USER_DIR, `${LABEL}.timer`);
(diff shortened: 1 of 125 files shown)


Drizzt321:
Not directly like this this, but in the general vicinity, I wanted to alter the CWD during startup to be a different home or home/local dir, I created https://github.com/danielmiessler/LifeOS/pull/1973

danielmiessler:
The resolver kernel (getClaudeDir honoring CLAUDE_CONFIG_DIR) landed on 08-19, and the rest of this is a 125-file sweep against 7.40.4 that would overwrite months of local change if applied. I'm doing the call-site sweep locally with credit to you at the resolver. Closing, thanks.

Issue #1910, opened 2026-08-20
MemorySystem.md documents MemoryRetriever as the producer of memory-retrievals.jsonl, but it never writes it — CortexHealth's retrieval-missing warn can never clear


jacobo-ortiz:
## Summary

`LIFEOS/DOCUMENTATION/Memory/MemorySystem.md` documents a producer that does not exist. Its artifact table says:

| Per-turn retrievals | `MEMORY/OBSERVABILITY/memory-retrievals.jsonl` | MemoryRetriever |

`LIFEOS/TOOLS/MemoryRetriever.ts` contains no write of any kind — grepping it for `appendFile|writeFile|OBSERVABILITY|.jsonl` returns zero lines. Nothing else in the tree writes that path either: five files mention `memory-retrievals`, and all five read it or document it (`CortexHealth.ts`, `MemoryStatus.ts`, and three docs).

`CortexHealth.ts` reads the file and, finding nothing, emits `retrieval-missing`. On a stock install that warn can never clear, so it sits on the status line permanently. Once a more severe finding resolves, it becomes the loudest thing on screen — a warn that no amount of work can close, on a subsystem that is in fact used on every single turn.

## Why the warn is misleading rather than merely noisy

Retrieval genuinely happens. `hooks/MemoryTurnStart.hook.ts` imports `getRelevantContext` from `MemoryRetriever` and injects the resulting `## RELEVANT MEMORY` block into the prompt. So the health check is right that there is no evidence and wrong-by-omission about what that means: it reads as "memory is not being used" when the truth is "memory is used constantly and never instrumented".

That is the worst shape for a health signal — it trains the operator to ignore the line that would carry a real failure.

## The consumer's contract is already fully specified

`validRetrievalRow` in `CortexHealth.ts` accepts exactly:

- required: `ts` (parseable date), `query_hash` (non-empty string), `returned_count` (non-negative integer), `duration_ms` (non-negative finite)
- optional: `top_score` (non-negative finite)
- no additional keys (`hasExactKeys`)

`getRelevantContext` already has every one of those values at hand.

## Suggested fix

Wrap `getRelevantContext` rather than instrumenting its five return points. One write site covers every path — including the cache hit — and cannot be forgotten when a sixth return is added later:

```ts
export function getRelevantContext(query, options = {}) {
  const startedAt = Date.now();
  const result = computeRelevantContext(query, options);
  logRetrieval(query, result, Date.now() - startedAt);
  return result;
}
```

[...]

PR #1920, opened 2026-08-20
fix(cortex): flip observability age check from oldest to newest write

### LINKED ISSUE #1915: CortexHealth warns on observability logs older than 30 days, but nothing in the tree prunes or rotates them — the warn is unreachable-to-satisfy on any install older than a month
## Summary

`CortexHealth.ts` warns when observability logs exceed 256 MiB or 30 days:

```ts
observabilityMaxAgeMs: 30 * 24 * 60 * 60 * 1000,
...
if (ageMs > thresholds.observabilityMaxAgeMs) add("observability-age-exceeded", "warn", ...)
```

`ObservabilitySystem.md` documents the same limits ("Recursive `.jsonl`/`.log` bytes and oldest mtime under `MEMORY/OBSERVABILITY/` | 256 MiB or 30 days | WARN").

Nothing in the tree prunes, rotates, or expires those logs. Grepping all of v7.40.4 for `prune|rotate|retention|unlinkSync|rmSync` in anything touching `OBSERVABILITY` returns only three things: the threshold definition, the check that reads it, and the `CORTEX_OBSERVABILITY_MAX_AGE_MS` env override. There is also no documented location to rotate *to* — `MEMORY/` has no `ARCHIVE/` convention.

So the warn is unreachable-to-satisfy by design: **any install older than 30 days that has not manually deleted files will show it forever**, and there is no shipped action that clears it.

## Why it lands the day you upgrade

`CortexHealth.ts` does not exist in v7.28.3 — it is new in v7.40.4. So on adoption the check appears fully formed and immediately measures a log directory that predates it by however long the install has existed. On this install the oldest file is a launchd log from the day the machine was set up; the check went live 5 days ago and reported a 31-day-old directory on its first run.

That is a poor first impression for a health signal: the operator's first encounter with it is a finding they cannot act on with anything the project ships.

## Why it matters beyond tidiness

This is the same structural shape as an adjacent issue I filed today (#1910, where `MemoryRetriever` never writes the file its own docs assign to it), with the halves swapped: there the *producer* of the evidence was missing, here the *remedy* for the finding is missing. Both end the same way — a permanent amber that trains the operator to stop reading the health line, which is the line that will one day carry something real.

## Suggested fix, in preference order

1. **Ship the rotation.** A retention step that moves or compresses `.jsonl`/`.log` files older than the threshold out of the measured tree, run from whatever already runs periodically. Then the warn means "rotation is broken", which is actionable.
2. **Or document the archive destination** and make the warn say it, so the operator knows the intended remedy rather than inventing one per install.
3. **Or scope the measurement** to the logs the system actively appends to, so historical run directories do not age the whole tree.

Whichever way it goes, the check and its remedy should ship together. A threshold with no mechanism behind it is a warning about the project's own housekeeping addressed to the user.

mcardinali-data:
Fixes #1915.

## Problem

`CortexHealth.ts` (new in v7.40.4) warns when the **oldest** file under `MEMORY/OBSERVABILITY/` exceeds 30 days. Nothing in the tree prunes or rotates these logs, so on any install older than a month the warning is permanently on and cannot be cleared — a health signal with no remedy, which trains operators to stop reading the health line (same class as #1910).

It also measures the wrong property. Observability holds append-only telemetry plus files that are legitimately stable; an old file is not a defect. The two real failure modes are:

- **unbounded growth** — already covered directly by the existing 256 MiB bytes check, untouched here
- **a dead telemetry pipeline** — nothing writes anymore; this is what an age signal can actually detect

## Fix

Flip the polarity: track the **newest** mtime instead of the oldest, and warn as `observability-stale` when nothing has written within the threshold. That warning is actionable (check hook wiring) and self-clearing on every healthy install.

- `observabilityEvidence()`: `oldestMs`/`Math.min` → `newestMs`/`Math.max`
- finding id `observability-age-exceeded` → `observability-stale`, message names the likely cause
- `CORTEX_OBSERVABILITY_MAX_AGE_MS` env knob keeps its name and shape
- bytes check unchanged

This is remedy option 3 from the issue (scope the measurement to what active logging proves), implemented as the minimal 6-line diff rather than shipping a rotation mechanism and a retention policy every install would have to adopt.

## Verification

Synthetic fixtures against `assessCortexEvidence`:

- dir with a 80-day-old file **plus** a fresh file → no observability finding (previously: permanent warn)
- dir where the newest file is >30 days old → `observability-stale`
- 300 MiB evidence → `observability-bytes-exceeded` still fires

Live install (logs written minutes ago, some files months old): overall `ok`, zero observability findings.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/CortexHealth.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/CortexHealth.ts
@@ -27,7 +27,7 @@ export interface CortexEvidence {
   reviewer?: ReviewerEvidence;
   retrieval?: RetrievalEvidence;
   proposals?: { pending: number; evidence: string; available?: boolean; malformedLines?: number[] };
-  observability?: { bytes: number; oldestMs: number | null; evidence: string; files?: number; available?: boolean };
+  observability?: { bytes: number; newestMs: number | null; evidence: string; files?: number; available?: boolean };
   index?: { status: "absent" | "no-index-v1" | "ok" | "mismatch" | "invalid"; manifest?: string; policy?: string; measuredAt?: number; canonicalHash?: string; manifestCanonicalHash?: string; indexHash?: string; manifestIndexHash?: string; indexPath?: string; indexedAt?: string; ageMs?: number };
   thresholds?: CortexThresholds;
 }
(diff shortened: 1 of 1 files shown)


danielmiessler:
Ported with credit: newest write instead of oldest, finding renamed to observability-stale, doc updated. Closing, next release. Thanks.

PR #1924, opened 2026-08-20
fix(skills): resolve customization lookup via CLAUDE_CONFIG_DIR, not hardcoded ~/.claude

neilinger:
## What

Every affected skill's `## Customization` block hard-codes
`~/.claude/LIFEOS/USER/CUSTOMIZATIONS/SKILLS/<Name>/` as a literal path in
the prose the model reads before running the skill. On any install where
`CLAUDE_CONFIG_DIR` points somewhere other than `~/.claude`, that path
doesn't exist — and the skill's own documented fallback ("if the
directory does not exist, proceed with skill defaults") turns the wrong
path into a **silently ignored customization**. No error, the user's
config is just quietly dropped.

## Why this form

Claude Code does not substitute shell variables into a skill body — only
`` !`cmd` `` dynamic-context injections are expanded before the model
sees the text. This PR replaces the static path line with a `find`-based
injection that resolves at invocation time, falling back to the
documented default (`$HOME/.claude`) when `CLAUDE_CONFIG_DIR` is unset —
so a default-config-dir install behaves exactly as it did before this
change.

Two details that matter and were found the hard way:
- **`find`, not a glob.** The harness evaluates these injections in zsh,
  whose default `nomatch` turns an unmatched glob into a hard expansion-time
  error — and an empty customization directory is the common case, so a
  glob here breaks the skill body on nearly every invocation.
- **Always exits 0.** A non-zero exit from the injected command aborts the
  entire skill invocation, so the command ends in `; true`.

## Scope

35 skills whose `## Customization` block used the exact canonical wording
(`**Before executing, check for user customizations at:**` + a standalone
path line). A 36th skill matching that wording, `ContextSearch`, has since
been removed from this tree, so it's not part of this diff.

Not included: `WorldThreatModel`, `CMUX`, and `Webdesign` use differently
worded customization blocks (different phrasing / embedded prose rather
than the standalone canonical line), so a safe mechanical fix isn't
possible without hand-editing skill-specific text — left for a follow-up.

## One more thing worth your attention

`Telos`'s `Workflows/Update.md` already uses this same `` !`cmd` ``
injection syntax (e.g. lines 81, 105, 108) — so the mechanism is already
an established idiom in this codebase. But that file is reached through
`Telos/SKILL.md`'s "Workflow Routing" table and read on demand rather than
loaded at skill-invocation time the way `SKILL.md` itself is, so the
injection there doesn't appear to actually be live. Same bug family
(hardcoded-path-adjacent), different manifestation — flagging it here
since I didn't want to scope-creep this PR into fixing it too.

## Behavior check

Verified in both bash and zsh, with `CLAUDE_CONFIG_DIR` unset (falls back
to `$HOME/.claude`, matches current behavior) and set to a directory that
does/doesn't contain a matching customization (correctly lists the file
or emits nothing) — exits 0 in every case.


danielmiessler:
Took the intent, not the mechanism. The dynamic-context injection runs a shell on every skill load across 35 public skills, which is an exec surface I won't add. The prose form (configRoot = $CLAUDE_CONFIG_DIR else ~/.claude) is going into the CreateSkill template and the 52 SKILL.md files in the config-root sweep, with credit. Closing.

PR #1929, opened 2026-08-21
Make the TELOS dashboard work out of the box

hjbrandt:
# Make the TELOS dashboard work out of the box

A fresh install ships a TELOS template whose examples don't match what the dashboard can actually read, so /life and /telos look empty or broken until you dig into the parser source to learn the expected format. This PR fixes both sides: the parsers understand more, and the template's examples now show exactly the format that works.

## What a fresh install gets

- **Goal progress bars.** The goals card has always said "Progress tracking appears once each goal has a progress field," but no code ever read such a field. Now you write `- Progress: 40%` under a goal and get a bar on /life, /telos, and the goal's detail page. Just a hand-typed percent, nothing more to maintain.
- **The Spark card shows actual sparks.** It used to show the section headings (like "Sensory / aesthetic") instead of the entries under them. Sparks are now read from bold bullets: `- **Gardening** — why it refuels you.` The template now also ships a `SPARKS.md` with sample categories (H3 headings) and sparks (bold bullets) — the shape the morning voice brief (`checks/life-morning-brief.ts`, which picks a random H3 from `SPARKS.md`) and the Spark card both read, so both work on a fresh install. A `## Sparks` section in the unified TELOS.md still takes precedence for single-file setups.
- **Sections written in the unified TELOS.md render correctly.** Entries like `### M1: ...` are recognized, the Summary and References labels no longer leak into card text, and a trailing notes block no longer shows up as fake extra entries.
- **The preferences card fills itself** from the Wisdom section and a hobbies file, skipping "(sample)" placeholders.
- **Two small bug fixes.** The owner column in work views never filled in because of a capitalization mismatch, and the /life projects card picked up rows from unrelated tables in PROJECTS.md.
- **The work card shows which project each thread belongs to** (second commit). Items carrying a `Property:` label get one row per project, with the strategy tag and life-area tags read from the item's labels. Boards that don't use those labels look exactly as before.
- **A template that teaches the right format.** The samples in `USER/TELOS/TELOS.md` are rewritten to the exact shapes the parsers read. Replace the words with your own, keep the layout, and the dashboard works on day one. The footer notes the few formats that matter.

## Three behaviors change on purpose

- Sparks are read from bold bullets instead of headings. If you used headings as individual sparks, move those names into bullets.
- The /life projects card reads only the projects section of PROJECTS.md when one exists, instead of scanning every table in the file.
- The work card on /telos groups threads by their `Property:` label when present, instead of always showing one flat "Work in flight" row.

Everything else keeps working as before: legacy per-topic files and bullet-style sections parse the same as they always did.

## Tested

All parser changes run live on the install they came from, and the rewritten template was fed through this branch's own parser functions: every sample renders — goals with bars, sparks, state fields, next actions, and dimensions.

### DIFF
--- a/LifeOS/install/LIFEOS/PULSE/Observability/observability.ts
+++ b/LifeOS/install/LIFEOS/PULSE/Observability/observability.ts
@@ -2267,15 +2267,28 @@ function parseCurrencyCell(cell: string): number {
   return plain ? parseFloat(plain[0]) : 0
 }
 
-function parseGoals(content: string): { id: string, text: string }[] {
+function parseGoals(content: string): { id: string, text: string, progress?: number }[] {
+  // Explicit-ID goals come in two shapes: bullet form (- **G0**: …) and
+  // heading form (### G0: …) — unified-TELOS installs often write headings.
   const withIds = content.split("\n")
-    .filter(l => /^[-*]\s*\*{0,2}G\d+\*{0,2}:/.test(l))
+    .filter(l => /^(?:[-*]\s*|#{2,4}\s+)\*{0,2}G\d+\*{0,2}:/.test(l))
     .map(l => {
       const m = l.match(/\*{0,2}(G\d+)\*{0,2}:\s*(.+)/)
       return m ? { id: m[1], text: m[2].trim() } : null
     })
-    .filter(Boolean) as { id: string, text: string }[]
-  if (withIds.length > 0) return withIds
+    .filter(Boolean) as { id: string, text: string, progress?: number }[]
+  if (withIds.length > 0) {
+    // Hand-set progress: a `- Progress: N%` bullet inside the goal's H3 block —
+    // the contract the /life goals card hint promises. Percent only,
+    // deliberately no milestone machinery.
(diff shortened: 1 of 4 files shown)


hjbrandt:
Related finding while adapting the Daemon skill on my install: `skills/Daemon/Tools/DaemonAggregator.ts` has the same root cause this PR addresses. It still reads the per-file TELOS layout (`TELOS/MISSION.md`, `GOALS.md`, `WISDOM.md`), so on installs using the unified single-file `TELOS.md` those sources are skipped cleanly (`--sources` reports them MISSING) and the generated daemon silently loses mission, goals, and wisdom. Once the unified-TELOS parsing from this PR lands, the aggregator could reuse the same section-based approach — happy to send that as a follow-up PR.

danielmiessler:
Ported selectively with credit: the scope fix, the Summary/References strip, progress bars, the sparks parser and SPARKS.md template. Skipped the TELOS.md template rewrite (it's on the old freshness header) and the Property grouping (it would have renamed my own board's row). Closing, next release. Thanks for the big one.

PR #1933, opened 2026-08-21
fix(cortex-health): accept pre-7.40.4 reviewer dispatch summaries

### LINKED ISSUE #1930: 7.40.4 CortexHealth marks a successful pre-upgrade reviewer run CRITICAL (`skipped_guard`/`skips` required)
### Summary

v7.40.4 taught `MemoryReviewer` to record guard refusals as `skipped_guard` / `skips` so an
`ESUSPECT_EROSION` block would not flip the run to `ok: false`. `CortexHealth.validDispatchSummary`
then required those two keys on **every** success row.

The health check only reads the latest `LIFEOS/MEMORY/OBSERVABILITY/reviewer-runs.jsonl` line.
After an upgrade that line is almost always a pre-7.40.4 success — `ok: true`, `parse_ok: true`,
three items written, zero failures — and it omits both new fields. The checker therefore reports
CRITICAL `Latest reviewer run is invalid` on a run that actually succeeded.

This is an upgrade footgun, not a dead memory loop. The next post-upgrade review (new writer)
would clear it. Until then every upgraded install with a prior successful review looks broken.

### Evidence

Stock `LIFEOS/TOOLS/CortexHealth.ts` (`validDispatchSummary`):

```ts
hasExactKeys(value, ["total", "by_type", "succeeded", "failed", "failures", "skipped_guard", "skips", "proposals_auto_applied", "proposals_auto_apply_failed"])
```

Stock `LIFEOS/TOOLS/MemoryReviewer.ts` (`DispatchSummary`) added `skipped_guard` and `skips` in
the same release. Pre-7.40.4 summaries are the 7-key set: `total`, `by_type`, `succeeded`,
`failed`, `failures`, `proposals_auto_applied`, `proposals_auto_apply_failed`.

`ObservabilitySystem.md` documents `dispatch_summary { total, by_type, succeeded, failed, failures }`
— a 5-key shape the checker no longer accepts (it also predates the two `proposals_*` keys the
pre-7.40.4 writer already emitted). Either way, the documented contract is not what the validator
now requires.

### Proposed fix

Treat a missing skip pair as zero guard refusals — the same fact the 7.40.4 writer records as
`skipped_guard: 0`, `skips: []`. Keep fail-closed on:

- hard failures (`failed !== 0` or non-empty `failures`)
- `proposals_auto_apply_failed !== 0`
- one skip field present without the other
- unknown keys
- total ≠ succeeded + skipped_guard

No log rewrite. No change to the writer.

### Why not "just run the reviewer again"

That clears one machine. Every other 7.1.x → 7.40.4 upgrade will still boot CRITICAL until its
first post-upgrade review happens to fire and succeed. The checker should accept its own
previous success schema.

## PR title

fix(CortexHealth): accept pre-7.40.4 reviewer dispatch summaries

## PR body

`validDispatchSummary` started requiring `skipped_guard` and `skips` in 7.40.4. Those fields
are new writer output. A successful pre-7.40.4 row — the usual latest row right after an
upgrade — was therefore classified `invalid` and `MemoryHealthCheck` went CRITICAL.

Missing skip fields now mean zero guard refusals. A half-new shape (one field without the
other) still fails. Smoke coverage is on `bun LIFEOS/TOOLS/CortexHealth.ts`.

## Patch

Apply on `LifeOS/install/LIFEOS/TOOLS/CortexHealth.ts` (and the live `LIFEOS/TOOLS/CortexHealth.ts`
copy it installs). Also update the reviewer-runs row in
`LifeOS/install/LIFEOS/DOCUMENTATION/Observability/ObservabilitySystem.md` so the documented
schema matches both writers.

The live-tested change is: make `skipped_guard` / `skips` optional-as-a-pair, defaulting to
`0` / `[]`, and add the six-case smoke test at the bottom of `CortexHealth.ts`
(`bun LIFEOS/TOOLS/CortexHealth.ts` → 6/0).

jbmml:
Fixes #1930.

`validDispatchSummary` required `skipped_guard`/`skips` on every success row, but those keys were added in 7.40.4. The health check reads only the latest `reviewer-runs.jsonl` line — right after an upgrade that's a pre-7.40.4 success lacking both keys — so a run that succeeded is reported CRITICAL until the next review fires.

Treats a missing skip pair as zero guard refusals (both-or-neither; one field without the other still rejects). All existing fail-closed cases unchanged. Adds a self-contained smoke test (`bun …/CortexHealth.ts` → 6/0) and updates the ObservabilitySystem doc line. No writer change, no log rewrite.


danielmiessler:
Ported with credit; the in-file smoke test moved into a proper test file. Closing, next release. Thanks.

PR #1934, opened 2026-08-21
fix(prompt-processing): search the real transcript root in findSessionJsonl

### LINKED ISSUE #1931: `findSessionJsonl` searches `LIFEOS/projects` — a directory that never exists — so `/rename` custom-title sync can never fire
### Summary

`hooks/PromptProcessing.hook.ts` `findSessionJsonl` looks for the session transcript under `paiPath('projects')` (and the `'Projects'` casing). `paiPath` resolves inside the LifeOS tree (`~/.claude/LIFEOS/...`), and nothing anywhere creates `LIFEOS/projects` — Claude Code writes transcripts to `~/.claude/projects/<slug>/` (or `$CLAUDE_CONFIG_DIR/projects`), one level shallower. Two other stock files (`ModelMix.ts`, `AgentTokenRollup.ts`) resolve the real transcript root correctly, confirming the correct location. So `findSessionJsonl` returns null on every call, `getCustomTitle` returns null, and the custom-title append that syncs a `/rename` into the session JSONL never fires.

### Evidence

- `hooks/PromptProcessing.hook.ts` (line 674): iterates `[paiPath('projects'), paiPath('Projects')]`, both nonexistent by construction.
- `hooks/lib/paths.ts`: `paiPath` roots at the LifeOS dir — the same file's other callers (`MEMORY/STATE/...`) confirm the root.
- Transcripts actually live at `<claude config dir>/projects/...` — one level up from where this searches.

### Suggested fix (one line, plus two hardening notes)

Search `join(process.env.CLAUDE_CONFIG_DIR || join(homedir(), '.claude'), 'projects')` instead of `paiPath('projects')`. Verified on a live install carrying this correction: `custom-title` records begin appearing only after the path is fixed, and none exist before it.

jbmml:
Fixes #1931.

`findSessionJsonl` searched `paiPath('projects')`, which resolves to `~/.claude/LIFEOS/projects` — a directory that never exists. Claude Code writes transcripts to `<config dir>/projects`. So the function always returned null, `getCustomTitle` always returned null, and the `/rename` custom-title JSONL sync never fired.

One-line fix: search `process.env.CLAUDE_CONFIG_DIR || ~/.claude` + `/projects` (`join`/`homedir` are already imported). Two other stock files resolve the transcript root this way already.

### DIFF
--- a/LifeOS/install/hooks/PromptProcessing.hook.ts
+++ b/LifeOS/install/hooks/PromptProcessing.hook.ts
@@ -671,7 +671,7 @@ function storeName(sessionId: string, label: string, source: string): void {
 /** Find Claude Code's session JSONL path for a given session ID. */
 function findSessionJsonl(sessionId: string): string | null {
   try {
-    for (const dir of [paiPath('projects'), paiPath('Projects')]) {
+    for (const dir of [join(process.env.CLAUDE_CONFIG_DIR || join(homedir(), '.claude'), 'projects')]) {
       if (!existsSync(dir)) continue;
       const r = Bun.spawnSync(['find', dir, '-maxdepth', '2', '-name', `${sessionId}.jsonl`],
         { stdout: 'pipe', stderr: 'pipe', timeout: 2000 });


danielmiessler:
Already ported from #1878 on 08-17; the local file keeps the Projects casing fallback too. Closing with thanks.

Issue #1932, opened 2026-08-21
`MigrateApprove.ts` resolves feedback-memory paths through a literal, never-interpolated `"${HARNESS_USER_DIR}"`


jbmml:
### Summary

`LIFEOS/TOOLS/MigrateApprove.ts` (~line 89), in the `memory/feedback` branch:

```ts
return join(HOME, ".claude", "projects", "${HARNESS_USER_DIR}", "memory");
```

The `${...}` sits inside an ordinary double-quoted TypeScript string — it is never interpolated, so the branch resolves to a directory literally named `${HARNESS_USER_DIR}`. No real slug can ever match; every path this branch produces is wrong, silently. This is a wrong-path/silent-miswrite bug rather than dead code: the branch runs and returns a path that cannot correspond to any real session directory.

### Suggested fix

`SessionHarvester.ts` in the same directory already implements the pattern this branch meant to use: it derives the project slug with `CLAUDE_DIR.replace(/[\/\.]/g, "-")` and builds `join(CLAUDE_DIR, "projects", CWD_SLUG)`. (`HARNESS_USER_DIR` exists nowhere as a real symbol or env var — it's only a doc-comment placeholder for that slug.) The fix is to compute the slug the same way and interpolate it, rather than embedding the literal string — likely `join(HOME, ".claude", "projects", <cwd-slug>, "memory")`.

> **Note:** the unguarded `join(LIFEOS_DIR, target)` in the same function (accepts `..`/absolute segments) is a potential path-traversal if `proposed_target` is model-influenced. That one is being routed to the maintainer through GitHub private vulnerability reporting rather than this public issue, per upstream's SECURITY.md.

waveman2020-sudo:
This is Shiva, Eugene's AI Assistant, reporting on Eugene's behalf.

Confirmed on this install: `LIFEOS/TOOLS/MigrateApprove.ts:89` contains the literal, non-interpolated string `"${HARNESS_USER_DIR}"` inside a double-quoted TS string — it can never match a real session directory. The suggested fix (deriving the slug the same way `SessionHarvester.ts` already does, via `CLAUDE_DIR.replace(/[\/\.]/g, "-")`, and interpolating it) checks out against this file.

danielmiessler:
[...]

PR #1935, opened 2026-08-21
fix(migrate-approve): interpolate the project slug, not the literal placeholder

jbmml:
Refs #1932. Opened as **draft** — the fix direction is right but the intended slug source needs a maintainer call (see below).

The `memory/feedback` branch returned `join(HOME, '.claude', 'projects', '${HARNESS_USER_DIR}', 'memory')` — the `${...}` sits in a plain double-quoted string and is never interpolated, so the path names a directory literally called `${HARNESS_USER_DIR}` and can never match a real session dir.

This patch derives the slug the way `SessionHarvester.ts` does (`CLAUDE_DIR` path slugified). **Open question:** `SessionHarvester` slugifies `~/.claude` itself; if the feedback path should key off a different cwd, the slug source should change accordingly. Happy to adjust once you confirm the intended value — the issue (#1932) has the full context.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/MigrateApprove.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/MigrateApprove.ts
@@ -85,8 +85,11 @@ function resolveTargetPath(target: string): string {
     return join(LIFEOS_DIR, target.startsWith("USER/") ? target : target);
   }
   if (target === "memory/feedback") {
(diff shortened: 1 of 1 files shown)


danielmiessler:
Ported with credit, adapted: the slug is derived from the config root (getClaudeDir) so it follows CLAUDE_CONFIG_DIR, and the principal-specific literal is gone from the shipped file. Closing, next release. Thanks.

PR #1941, opened 2026-08-22
Honor CLAUDE_CONFIG_DIR in hook paths and runtime

MarnuLombard:
Installing under a relocated config root (`--config-root` / `CLAUDE_CONFIG_DIR`, e.g. a project-scoped `.claude`) copies the hook scripts into that root but still wires them to `$HOME/.claude`.

The result: every hook resolves to the global `~/.claude`, which holds none of the copied scripts, and the `SessionStart` MergeSettings hook rewrites the global `~/.claude/settings.json` from inside the relocated install.

This roots the hook commands at `${CLAUDE_CONFIG_DIR:-$HOME/.claude}` and makes `getClaudeDir()` honor `CLAUDE_CONFIG_DIR`, the same relocation every install tool and runtime module (Doctor.ts, PULSE Conduit) already respects.

Default installs are byte-identical: with `CLAUDE_CONFIG_DIR` unset both paths fall back to `~/.claude`.

---

- 05a1e86 **fix: honor CLAUDE_CONFIG_DIR in hook paths and runtime**
  - hooks.json commands rooted at `${CLAUDE_CONFIG_DIR:-$HOME/.claude}` so a relocated install runs its own hooks.
  - `getClaudeDir()` / `getLifeosDir()` resolve settings, skills, and memory under the same config root at runtime.

### DIFF
--- a/LifeOS/install/hooks/hooks.json
+++ b/LifeOS/install/hooks/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "$HOME/.claude/hooks/ContextReduction.hook.sh"
+            "command": "${CLAUDE_CONFIG_DIR:-$HOME/.claude}/hooks/ContextReduction.hook.sh"
           }
         ]
       },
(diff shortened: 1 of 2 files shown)


MarnuLombard:
Closing to prepare a more complete change: route every hook and tool through CLAUDE_CONFIG_DIR/paths.ts, fix the settings-merge dedup on upgrade, deploy settings.system.json into the relocated root, and have the launcher export CLAUDE_CONFIG_DIR - so the install can fully live inside a given config root, not just the 9 hooks that already import paths.ts.

Issue #1911, opened 2026-08-20
CLAUDE_CODE_FORK_SUBAGENT=1 by default breaks isSubagentContext() on every main-session hook (Claude Code 2.1.232+)


dedafrost:
## Summary

`hooks/lib/subagent.ts`'s `isSubagentContext()` treats `process.env.CLAUDE_CODE_FORK_SUBAGENT === '1'` as proof the current process is a forked subagent (added for #1831). As of Claude Code 2.1.232+, this variable is set to `"1"` by default in every interactive session — it now means "fork mode is enabled for this session," not "this process is a fork."

Effect: `isSubagentContext()` returns `true` on every real turn in a normal main session, so every hook that early-exits on it (MemoryTurnStart and 7 others per that file's own docstring) silently no-ops before doing anything — no memory injection, no delta surface, no retrieval-evidence logging, ever, in ordinary use. It's fail-open by design, so nothing looks broken from the outside; I only found it chasing a stale `retrieval-stale` CortexHealth warning that turned out to be the shallow end of it.

## Reproduction

From inside a real (non-subagent) interactive session, dump the markers `isSubagentContext()` checks from within an actual UserPromptSubmit hook subprocess:

```json
{"CLAUDE_PROJECT_DIR":"/Users/<home>/.claude","CLAUDE_AGENT_TYPE":null,"CLAUDE_CODE_SUBAGENT_NAME":null,"CLAUDE_CODE_SUBAGENT_TYPE":null,"CLAUDE_CODE_FORK_SUBAGENT":"1","CLAUDE_AGENT_SDK":null}
```

Every marker is unset except the fork one — in a session that is not a fork.

## Fix applied locally

Dropped `CLAUDE_CODE_FORK_SUBAGENT` from the union in `isSubagentContext()`. This reopens the narrower #1831 case (a genuine fork that sets *only* that marker would now read as main-session) until there's a reliable replacement — but the alternative, every main session losing this hook's output on every turn, is strictly worse.

The real fix is probably a `SessionStart` hook with `matcher: "fork"` (Claude Code does still expose that distinction at session-start, per its hooks reference) writing a per-session state file that `isSubagentContext()` reads, since `UserPromptSubmit` hooks don't get a fork-scoped matcher and it's unclear whether their input JSON carries a usable `source" field.

Happy to open a PR with the local patch (removes the stale check, documents why) if useful — the SessionStart-based replacement is a bigger piece I haven't built yet.

[...]

Issue #1864, opened 2026-08-15
settings.system.json ships CLAUDE_CODE_FORK_SUBAGENT=1 — every main session classifies itself as a subagent and nine hooks skip their work


jlacour-git:
**Affected:** v7.40.4, fresh install (verified against the tagged payload and a
clean sandbox install)
**Component:** `LifeOS/install/settings.system.json` + `LifeOS/install/hooks/lib/subagent.ts`
**Impact:** high — context loading, memory surfacing and ISA sync are silently
disabled for every session on a fresh install. Trivial fix.

### Summary

`settings.system.json` ships this in its `env` block (line 11):

```json
"CLAUDE_CODE_FORK_SUBAGENT": "1",
```

`isSubagentContext()` in `hooks/lib/subagent.ts` treats exactly that value as
"this process is a subagent" (line 33):

```ts
process.env.CLAUDE_CODE_FORK_SUBAGENT === '1' ||
```

Since a settings `env` block applies to **every** session the install starts,
the main session tests positive too, and every consumer of
`isSubagentContext()` skips its work: `LoadContext`, `LoadMemory`,
`MemoryTurnStart`, `MemoryDeltaSurface`, `MemoryReviewFire`, `ISASync`,
`SystemChangeSurface`, `ConfigEvalFire`, `KittyEnvPersist` (nine hooks sharing
the lib).

The marker itself is the fix for #1831 (forked subagents set only
`CLAUDE_CODE_FORK_SUBAGENT`, so the union test needed it). The env line that
ships alongside it makes the test always-true: the harness should be the only
thing that ever sets this variable, per fork process — not the install's
settings, for every process.

### Repro

Fresh v7.40.4 install into a clean HOME, then any session:

```
claude -p "Say: ready."
```

**Observed:** SessionStart logs `🤖 Subagent session - skipping context loading`
(`hooks/LoadContext.hook.ts:467`). In our sandbox, 8 of 8 headless sessions
logged the skip; no LifeOS context, no memory surfacing, no ISA sync.

**Control:** remove the key from the installed `settings.json` `env` block, run
the same prompt again.

**Observed:** `✅ LifeOS session initialization complete`, advisory digest and
dynamic context present. (Overriding via the process environment did not help
in our runs — the settings `env` value won; removing the key is the fix that
worked.)

**Expected:** a main session never classifies itself as a subagent;
`CLAUDE_CODE_FORK_SUBAGENT` is only ever present in the environment of an
[...]

Issue #1831, opened 2026-08-12
isSubagentContext() misses forked subagents — they set only CLAUDE_CODE_FORK_SUBAGENT, which is not in the union, so 8 hooks treat a fork as a main session


DRAZY:
**Affected:** v7.28.3 (current release)
**Component:** `LifeOS/install/hooks/lib/subagent.ts` → `isSubagentContext()`
**Harness:** Claude Code 2.1.228
**Impact:** Functional/efficiency, not security. Delegates receive main-session context they already inherited.

## Summary

`isSubagentContext()` is documented as "the UNION of every known marker," and its header states the failure mode it exists to prevent:

> a missed subagent leaks main-session context (and a duplicate surface line) into a delegate

A **forked** subagent (`Agent` tool with `subagent_type: "fork"`) is a marker the union has not learned. It sets `CLAUDE_CODE_FORK_SUBAGENT=1` and `CLAUDE_CODE_CHILD_SESSION=1`, and sets **none** of the five things the guard tests. So `isSubagentContext()` returns `false` inside a fork, and every consumer treats it as the main session.

## The union vs. what a fork actually sets

```ts
// LifeOS/install/hooks/lib/subagent.ts
export function isSubagentContext(): boolean {
  const projectDir = process.env.CLAUDE_PROJECT_DIR || '';
  return Boolean(
    projectDir.includes('/.claude/Agents/') ||   // unset in a fork
      process.env.CLAUDE_AGENT_TYPE ||           // unset in a fork
      process.env.CLAUDE_CODE_SUBAGENT_NAME ||   // unset in a fork
      process.env.CLAUDE_CODE_SUBAGENT_TYPE ||   // unset in a fork
      process.env.CLAUDE_AGENT_SDK === '1',      // unset in a fork
  );
}
```

Every branch misses. This is not a partial match — it is a clean miss on all five.

## Reproduction

Spawn a fork subagent and have it dump its environment:

> Run exactly this one Bash command and return its raw stdout verbatim as your entire final message:
> `env | grep -iE 'CLAUDE|SUBAGENT|FORK|AGENT' | sort`

Result on Claude Code 2.1.228 (trimmed to the relevant lines; `PATH` and socket paths removed):

```
AI_AGENT=claude-code_2-1-227_agent
CLAUDE_CODE_CHILD_SESSION=1
CLAUDE_CODE_ENTRYPOINT=cli
CLAUDE_CODE_FORK_SUBAGENT=1
[...]

PR #1942, opened 2026-08-22
fix(hooks): CLAUDE_CODE_FORK_SUBAGENT is a feature flag, not a subagent marker — disables 9 hooks on every install

Autodidact-med:
## Problem

`settings.system.json` sets `CLAUDE_CODE_FORK_SUBAGENT: "1"` in the global `env` block, because that variable is the **feature-enable flag** for forked subagents (per the Claude Code changelog: *"Forked subagents can now be enabled on external builds by setting `CLAUDE_CODE_FORK_SUBAGENT=1`"*).

`hooks/lib/subagent.ts` also treats that same variable as a **runtime marker** identifying a subagent.

Because the flag is set globally, `isSubagentContext()` returns `true` in every session, including the main one. All nine consumer hooks that gate on it silently no-op.

## Impact

On a default install, these stop firing entirely:

`MemoryReviewFire`, `LoadMemory`, `MemoryTurnStart`, `MemoryDeltaSurface`, `ISASync`, `ConfigEvalFire`, `SystemChangeSurface`, `LoadContext`, `KittyEnvPersist`

Nothing errors. The autonomic memory loop simply stops: `review-state.json`, `MEMORY/STATE/memory-inject`, and the delta-surface heartbeat stop being written, while every hook *not* behind this guard keeps working normally. That asymmetry is the only visible symptom.

Observed on a 7.40.4 install: the memory review loop was dead for five days before anyone noticed, and the hot layer stopped accumulating entries.

## Why the flag cannot work as a marker

A forked subagent inherits its parent's environment. Even without the global `env` entry, the flag would be present in both parent and fork, so it cannot distinguish them in principle.

## Fix

Remove `CLAUDE_CODE_FORK_SUBAGENT` from the union in `isSubagentContext()`. The explicit markers (`CLAUDE_AGENT_TYPE`, `CLAUDE_CODE_SUBAGENT_NAME`, `CLAUDE_CODE_SUBAGENT_TYPE`, `CLAUDE_AGENT_SDK`, and the `/.claude/Agents/` path test) are unaffected.

## Trade-off

This removes the mitigation added for the fork double-context-injection issue referenced in the original comment. A fork may again re-inject main-session context. That is a duplicated context line; the current behaviour is a dead memory subsystem on every install. If fork detection is still wanted, it needs a marker the harness sets *only* in forks, which this flag is not.

## Verification

With the flag removed, `isSubagentContext()` returns `false` in a main session, the Stop hook resumes writing per-session state and `review-state.json`, and the reviewer fires on its normal cadence again.

### DIFF
--- a/LifeOS/install/hooks/lib/subagent.ts
+++ b/LifeOS/install/hooks/lib/subagent.ts
@@ -27,10 +27,12 @@ export function isSubagentContext(): boolean {
       process.env.CLAUDE_AGENT_TYPE ||
       process.env.CLAUDE_CODE_SUBAGENT_NAME ||
       process.env.CLAUDE_CODE_SUBAGENT_TYPE ||
-      // Forked subagents set ONLY the fork marker — none of the above.
-      // Without it, 8 hook consumers re-inject main-session context into
-      // forks that inherited it via cache. (public issue #1831, @DRAZY)
-      process.env.CLAUDE_CODE_FORK_SUBAGENT === '1' ||
+      // CLAUDE_CODE_FORK_SUBAGENT is NOT a runtime fork marker — it is the
+      // FEATURE-ENABLE flag for forked subagents, and settings.system.json sets
+      // it globally in `env`. Every session therefore carried it, this function
+      // returned true unconditionally, and all nine consumer hooks silently
+      // no-oped. A fork inherits the parent env anyway, so the flag cannot
+      // discriminate a fork from its parent even in principle.
       process.env.CLAUDE_AGENT_SDK === '1',
   );
 }


theNetworkChuck (commented):
Independent confirmation on the **upgrade** path, plus before/after evidence for the exact change in this PR.

**Environment:** macOS, Claude Code 2.1.241, LifeOS 7.40.4 upgraded from 7.28.3 via the documented Update workflow (`install.sh` + `OverlaySystem.ts --apply`).

**Symptom as observed:** `MemoryHealthCheck` went `review-stale` ("Last reviewer fire was 8 days ago"). `reviewer-fires.jsonl` last row was the day before the 7.40.4 upgrade; `review-state.json` mtime never moved afterward, i.e. `MemoryReviewFire` was not even reaching its state write on Stop.

**Isolation:**

- Piping a synthetic Stop payload into `MemoryReviewFire.hook.ts` under the live session env: exit 0, no state write, no stderr. Same payload under `env -i HOME=… PATH=…`: state written. Adding back only `CLAUDE_CODE_FORK_SUBAGENT=1` reproduces the no-op.
- `isSubagentContext()` evaluated in the live main session: `true` before this PR's change, `false` after, with the variable still set to `1` in the environment.

**Verification of this PR's diff applied locally:**

- `SystemChangeSurface` (one of the nine consumers) emitted its delta on the very next PostToolUse.
- Driving eight Stop events for the live session spawned `MemoryReviewer.ts`; the run completed `ok: true`, `parse_ok: true`, inference 2.7 s.
- `MemoryHealthCheck`: `review-stale` and `reviewer-stale` cleared, `review-fresh` ok. Remaining warnings on that install are unrelated (`retrieval-missing`, `observability-age-exceeded`).
- `MemoryDeltaSurface` fed its `<lifeos-memory-delta>` block on the next UserPromptSubmit; the 🧠 line had been absent since the upgrade.

Two notes for the maintainer:

1. #1911 reports Claude Code ≥ 2.1.232 sets this variable to `1` by default in interactive sessions. If so, removing the `env` entry from `settings.system.json` would not have been sufficient on current harness versions; the guard change in this PR is the one that works.
2. With the row gone, #1831's original concern (forks re-injecting main-session context) is open again. That needs a marker that is actually fork-specific; tracked separately, out of scope here.

LGTM on the diff as-is.

danielmiessler:
The flag removal landed 08-17 under #1864, same hunk. Closing with thanks; you're credited as independent confirmation at the fix site.

PR #1943, opened 2026-08-22
Honor CLAUDE_CONFIG_DIR across hooks, tools, and launcher

MarnuLombard:
LifeOS hooks and install tools hardcoded `~/.claude`, so an install under a relocated config root (`--config-root` / `CLAUDE_CONFIG_DIR`) copied files into its own root but resolved them against the global `~/.claude`.

Worse, a SessionStart `MergeSettings` hook would rewrite the GLOBAL `settings.json` from inside a relocated install.

This routes all 36 hooks and the install tooling through the resolved config root, has the `lifeos` launcher export `CLAUDE_CONFIG_DIR` so a relocated install actually runs in its own path, and closes two latent gaps: `settings.system.json` was never deployed, and the upgrade merge did not dedup the new templated command form. Default installs are byte-identical.

---

- 543f20e **fix: honor CLAUDE_CONFIG_DIR across hooks, tools, and launcher**
  - Route every hook, install tool, and the launcher through the resolved config root.
  - Deploy the missing `settings.system.json` (repointed on relocation) and dedup the upgrade path.

### DIFF
--- a/LifeOS/Tools/DeployCore.ts
+++ b/LifeOS/Tools/DeployCore.ts
@@ -22,10 +22,11 @@
  *   (dry-run by default — reports the plan per target without writing)
  */
 
-import { existsSync, mkdirSync, readdirSync } from "node:fs";
+import { existsSync, mkdirSync, readdirSync, readFileSync } from "node:fs";
 import { homedir } from "node:os";
 import { basename, dirname, join } from "node:path";
 import { copyMissing, detectDevTree } from "./InstallEngine";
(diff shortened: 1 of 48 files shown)


danielmiessler:
Kernel landed 08-19; the 48-file sweep is stale against local hooks. Two pieces of yours are going in on their own: lifeos.ts exporting CLAUDE_CONFIG_DIR when unset, and the ROOT_FORMS normalized-command dedup for upgrade merges, both credited. Closing, thanks.

PR #1958, opened 2026-08-24
fix(hooks): detect subagents from hook stdin (agent_id / agent_type, SessionStart source=fork), drop the fork feature-flag test

theNetworkChuck:
## Summary

`isSubagentContext()` keys on environment variables. That cannot work for forked subagents, and the workaround it shipped with (`CLAUDE_CODE_FORK_SUBAGENT === '1'`) turned every main session into a "subagent" and silently disabled nine hooks (#1864, #1911, PR #1942).

This PR removes the fork-flag row (same hunk as #1942) **and** replaces it with the signal that actually distinguishes a fork: the hook's stdin JSON. Consumers that already parse stdin pass it to the guard; the two SessionStart consumers that did not read stdin now do, best-effort.

## Why an env var can never mark a fork

Measured on Claude Code 2.1.241 by spawning a fork and dumping its environment from inside: **a fork's environment is byte-identical to its parent's** — same `CLAUDE_CODE_SESSION_ID`, same `CLAUDE_PID`, no `CLAUDE_AGENT_TYPE` / `CLAUDE_CODE_SUBAGENT_NAME` / `CLAUDE_CODE_SUBAGENT_TYPE`, and `CLAUDE_CODE_FORK_SUBAGENT=1` in both. Nothing in `process.env` separates them, so with the flag row removed (#1942) the original #1831 leak (forks re-injecting main-session context) is open again.

## What does mark a fork

The [hooks reference](https://code.claude.com/docs/en/hooks) documents `agent_id` and `agent_type` as common input fields "for subagent hooks", and lists `fork` as a `SessionStart` matcher (i.e. `source: "fork"` on that event's input). Observed on the same 2.1.241 fork: its PostToolUse input carries `agent_type: "fork"` plus an `agent_id`; an ordinary delegate carries `agent_type: "general-purpose"`; main-session input carries neither.

## Change

`lib/subagent.ts`

- `isSubagentContext(input?: unknown)` — optional parsed-stdin argument; result is the existing env union **OR** the stdin stamp. Every zero-arg call site keeps compiling and behaves as before, minus the false positive.
- New `isSubagentHookInput(input)` — true when `agent_id` or `agent_type` is a non-empty string, or `hook_event_name === "SessionStart"` with `source === "fork"`. Pure, zero-dep, never throws.
- Fork-flag row removed (identical to #1942).

Consumers now pass their parsed stdin: `ISASync`, `SystemChangeSurface` (`HookInput` gains optional `agent_id`/`agent_type`), `ConfigEvalFire`, `MemoryReviewFire`, `MemoryTurnStart`, `KittyEnvPersist` (guard moved after its existing stdin read; `source` comment now includes `fork`), `LoadContext` (adds a best-effort stdin read). `LoadMemory` and `MemoryDeltaSurface` are unchanged: their guarded paths are standalone probes, and their real caller `MemoryTurnStart` is now guarded on stdin.

`lib/hook-io.ts` — `HookInput` documents the optional `agent_id`, `agent_type`, `source` fields.

No new dependencies; hooks stay importable before `bun install`.

## Verification

All runs with `CLAUDE_CODE_FORK_SUBAGENT=1` in the environment (what `settings.system.json` and Claude Code ≥ 2.1.232 both set), against a throwaway HOME, using stdin shapes copied from real captured hook input:

- Guard unit cases (18): main-session input with the flag set → `false` for PostToolUse / Stop / SessionStart-startup; fork PostToolUse (`agent_type: "fork"`) and SessionStart `source: "fork"` → `true`; ordinary delegate stamp → `true`; each surviving env marker alone → `true`; null / string / array / empty-string / non-string / `source: "fork"` on a non-SessionStart event → `false`, none throw.
- Hook-level: `MemoryReviewFire` writes its per-session state on a main Stop and writes nothing on a fork-stamped Stop; `LoadContext` runs on `source: "startup"` and prints its skip line on `source: "fork"`; `KittyEnvPersist` same split; `SystemChangeSurface.run()` returns its surface line for a main write and `null` for a fork-stamped one.
- Each edited hook transpiles (`bun build --target=bun --no-bundle`).

Relationship to open work: stacks on #1942 (the removal hunk is the same; merging either first leaves a trivial conflict on that one hunk). Closes the re-opened #1831 concern and the "what is the reliable fork marker" follow-up noted on #1864 / #1911.

### DIFF
--- a/LifeOS/install/hooks/ConfigEvalFire.hook.ts
+++ b/LifeOS/install/hooks/ConfigEvalFire.hook.ts
@@ -41,15 +41,17 @@ function isSentinel(path: string): boolean {
   return /(^|\/)CLAUDE\.md$/.test(path);
 }
 
-function readInput(): { filePath: string | null } {
+// `input` is the whole parsed payload — the subagent guard reads its agent stamp,
+// which is the only marker a forked subagent sets (see lib/subagent.ts).
+function readInput(): { filePath: string | null; input: unknown } {
   try {
     const raw = readFileSync(0, 'utf8');
-    if (!raw.trim()) return { filePath: null };
+    if (!raw.trim()) return { filePath: null, input: null };
     const j = JSON.parse(raw);
     const fp = j?.tool_input?.file_path;
-    return { filePath: typeof fp === 'string' ? fp : null };
+    return { filePath: typeof fp === 'string' ? fp : null, input: j };
   } catch {
-    return { filePath: null };
+    return { filePath: null, input: null };
   }
 }
 
@@ -76,9 +78,9 @@ function saveLastFire(iso: string): void {
 
 function main(): void {
   try {
-    if (isSubagent()) process.exit(0);
+    const { filePath, input } = readInput();
+    if (isSubagent(input)) process.exit(0);
 
-    const { filePath } = readInput();
     if (!filePath || !isSentinel(filePath)) process.exit(0);
     if (minutesSince(loadLastFire()) < DEBOUNCE_MINUTES) process.exit(0);
     if (!existsSync(RUNNER)) process.exit(0);
--- a/LifeOS/install/hooks/ISASync.hook.ts
+++ b/LifeOS/install/hooks/ISASync.hook.ts
@@ -153,7 +153,7 @@ async function main(): Promise<string | null> {
   // without a phase edit. Per-session dedupe file; subagents never emit (their
   // ISA edits would strip-spam their own contexts, which helps nobody).
   let stripDelta: string | null = null;
-  if (input.session_id && fm.slug && !isSubagentContext()) {
+  if (input.session_id && fm.slug && !isSubagentContext(input)) {
     try {
       const stripDir = join(homedir(), '.claude/LIFEOS/MEMORY/STATE/ascent-strip');
       const stripFile = join(stripDir, `${String(input.session_id).replace(/[^\w-]/g, '')}.json`);
--- a/LifeOS/install/hooks/KittyEnvPersist.hook.ts
+++ b/LifeOS/install/hooks/KittyEnvPersist.hook.ts
@@ -20,21 +20,24 @@ import { isSubagentContext } from './lib/subagent';
 
 const paiDir = getLifeosDir();
 
-// Skip for subagents
-if (isSubagentContext()) process.exit(0);
-
 // Read session_id + source from stdin (SessionStart hook input)
-// source ∈ {"startup", "resume", "compact", "clear"}; absent on older CC versions.
+// source ∈ {"startup", "resume", "compact", "clear", "fork"}; absent on older CC versions.
 let sessionId = '';
 let source = '';
+let parsed: unknown = null;
 try {
   const raw = readFileSync(0, 'utf-8');
   if (raw) {
-    const parsed = JSON.parse(raw);
-    sessionId = String(parsed.session_id || '');
-    source = String(parsed.source || '');
+    const json = JSON.parse(raw);
+    parsed = json;
+    sessionId = String(json.session_id || '');
+    source = String(json.source || '');
   }
 } catch { /* best-effort */ }
+
+// Skip for subagents — a fork announces itself as `source: "fork"` on stdin and
+// nowhere in the env, so this reads the payload above (see lib/subagent.ts).
+if (isSubagentContext(parsed)) process.exit(0);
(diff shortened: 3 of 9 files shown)


christauff:
Independently reproduced and root-caused this on a live 7.40.4 install before finding these issues — landing on the same line and the same conclusion. Adding field data, since the verification above is (correctly) synthetic and this is what it looks like in production.

**Outage window, measured:** dark from `2026-08-16T03:14Z` to `2026-08-23T08:52Z` — seven days. Pinned by three independent state files all frozen within the same minute (`reviewer-fires.jsonl`, `review-state.json`, and a delta-surface heartbeat), which is what made the shared cause obvious.

**Blast radius matches your list:** nine consumers silent — `MemoryReviewFire`, `MemoryTurnStart`, `LoadContext`, `LoadMemory`, `MemoryDeltaSurface`, `SystemChangeSurface`, `ConfigEvalFire`, `KittyEnvPersist`, `ISASync`.

**Why it survived a week undetected, which I think is worth calling out in the fix's rationale:** every consumer fails open. No error, no exit code, no log line. A hook that decides not to run is indistinguishable from a hook that had nothing to do, so the install reported full coverage the entire time. The user-visible symptom was purely negative — memory/system status lines that simply never appeared across a long session, which reads as "nothing to report."

**Independent confirmation that the stdin signal is the right one.** Separately from this bug, I built a feature on `agent_id` from `PostToolUse` input (classifying whether a skill invocation was self-routed, user-invoked, or delegated). Across 356 real transcripts it discriminated cleanly: main-session `PostToolUse` carries neither `agent_id` nor `agent_type`, delegated calls carry both. That's a different use case arriving at your observation from the other direction, so the field you're keying on holds up outside this fix.

**One note on the stacking.** My local bridge only removed the flag row, which restores the nine hooks and reopens #1831 exactly as your summary predicts. The stdin replacement is the part that makes it a fix rather than a trade, so from here this looks like the right shape to converge on rather than #1942 alone.

Happy to run the patch against a production install and report back if that's useful.

christauff:
Tested this against a live 7.40.4 install (Claude Code 2.1.24x, Linux, bun). **All green — and it is strictly better than the removal-only fix, which I had been running.**

Method: reverted my local removal-of-the-flag-row bridge to restore upstream state, confirmed the bug reproduced (`isSubagentContext()` → `true` in a main session with `CLAUDE_CODE_FORK_SUBAGENT=1`), then applied this PR unmodified.

**Applies clean.** `git apply --check` passes on all 9 files against a live tree — no conflict, despite my having independently touched the same hunk.

**Transpile:** 9/9 via `bun build --target=bun --no-bundle`.

**Guard cases: 14/14**, run with the flag set in the environment throughout. The decisive one — main-session `PostToolUse` / `Stop` / `SessionStart source=startup` with the flag set → `false`. Fork `agent_type:"fork"` → `true`; `SessionStart source:"fork"` → `true`; ordinary delegate `agent_type:"general-purpose"` → `true`; `agent_id` alone → `true`. Rejections behave: `null`, `undefined`, string, array, empty-string `agent_type`, non-string `agent_id`, and `source:"fork"` on a non-SessionStart event all → `false`, none throw.

**Hook-level discrimination, real stdin shapes:**

| hook | main session | fork-stamped | delegate |
|---|---|---|---|
| `MemoryReviewFire` | writes per-session state | no write | no write |
| `LoadContext` | loads context | prints skip line | — |
| `SystemChangeSurface` | emits the surface line | suppressed (empty) | — |

**One portability note, not a defect in this PR.** `MemoryReviewFire` derives its paths from `pathResolve(homedir(), ".claude")` rather than the `LIFEOS_DIR` env var, so an attempt to sandbox it by redirecting `LIFEOS_DIR` silently writes to the real tree instead. It cost me one confused test round. Worth knowing for anyone writing tests against these hooks, and possibly worth a separate issue.

Net for this install: the nine hooks are live again and forks are correctly excluded. I have replaced my own bridge with this PR and recorded it as running-an-unmerged-PR, to be retired on merge. Happy to re-run anything specific if it would help.

danielmiessler:
Ported by hand with credit to you, christauff and lgertel: `isSubagentHookInput` reads agent_id / agent_type / SessionStart source=fork from stdin and is OR'd into the existing detection at all seven consumers. Couldn't apply the diff directly because those hooks moved locally since 08-14. Tests added. Closing, next release. Thanks.

Issue #1689, opened 2026-07-28
SystemFileGuard's deny-list default path can never exist on an install — the #1504 warning will fire forever, and creating the directory disables seven install tools


christauff:
Follow-on to #1504, not a re-report of it. That issue said the guard "looks active but may be inert," and your fix — say so loudly, once — is the right call. This is about what a user can do *after* they see the warning, which as far as I can tell is nothing.

Verified against `main` @ `d1d6240`.

## The trap

`hooks/lib/system-file-guard-core.ts:21-22`:

```ts
const DEFAULT_DENY_LIST_PATH = join(CLAUDE_ROOT, "skills/_LIFEOS/DENY_LIST.txt");
const DEFAULT_HASHES_PATH    = join(CLAUDE_ROOT, "skills/_LIFEOS/DENY_HASHES.json");
```

`loadPatterns()` returns `[]` at `:70` when that file is absent, and `loadHashContext()` returns `null` at `:115`. So the guard evaluates every SYSTEM write against an empty filter.

That path cannot be populated on an install, for two independent reasons:

**1. `skills/_LIFEOS` is stripped from the release by design.** `ls skills/_*` in the payload returns nothing — no underscore-prefixed skill ships. Your own comment in `InstallEngine.ts:153-155` states it: the private maintenance skill "exists ONLY in the author's source repo — never in a public install (release tooling strips all `_ALLCAPS` skills)."

**2. Creating it locally is worse than leaving it empty.** `detectDevTree()` at `InstallEngine.ts:159-161` is exactly `existsSync(join(configRoot, "skills", "_LIFEOS"))`, and seven tools refuse to run when it returns true:

| Tool | Refusal |
|---|---|
| `LinkUser.ts` | "refusing to relink" |
| `InstallHooks.ts` | "refusing to mutate" |
| `DeployCore.ts` | "refusing to deploy core" |
| `DeployComponents.ts` | "refusing to deploy components" |
| `SeedPulse.ts` | "refusing to seed Pulse" |
| `ScaffoldUser.ts` | "refusing to scaffold" |
| `ActivateImports.ts` | "refusing to edit" |

So a user who reads the warning, understands it, and does the obvious thing — `mkdir -p skills/_LIFEOS && $EDITOR DENY_LIST.txt` — silently converts their install into something the installer and deploy tooling will no longer touch. The marker is a good design; it just collides with the one directory the guard needs.

[...]

Issue #1902, opened 2026-08-19
DeriveDenyHashes.ts can never write on a public install: #1488's guard outlived #1689's path move


pybe:
## Summary

`LIFEOS/TOOLS/DeriveDenyHashes.ts` returns before writing on every public install, so `LIFEOS/USER/SECURITY/DENY_HASHES.json` can never be generated. Token derivation runs, reports its count, and the result is discarded at exit 0.

This is a residual of two of your own fixes interacting, not a new defect.

## Cause

**#1488** (closed 16 Jul) added the early return:

```ts
if (!existsSync(join(CLAUDE, "skills", "_LIFEOS"))) {
  console.log("[DeriveDenyHashes] skills/_LIFEOS absent (public install) — skipping hash write");
  return;
}
```

Correct at the time. `OUT_PATH` then pointed inside that directory, whose existence is the dev-tree marker, so creating it on a public install disables seven installer tools. Skipping was the right call and your closing note says exactly that.

**#1689** (closed 29 Jul) then moved both writer and reader:

```ts
const OUT_PATH = join(CLAUDE, "LIFEOS", "USER", "SECURITY", "DENY_HASHES.json");
```

Your closing note: *"Both the writer and the reader now use LIFEOS/USER/SECURITY/DENY_HASHES.json, with the output dir created first."*

The move removed the guard's premise. The guard stayed. Thirteen days apart, and neither issue had reason to look at the other: #1488 was about a failing scheduled job, #1689 about an inert guard path.

The file's own header now documents the new location and the reason for it, three lines above a check that still tests the old one.

## Repro

Any public install with no maintainer skill directory:

```
$ bun ~/.claude/LIFEOS/TOOLS/DeriveDenyHashes.ts
[DeriveDenyHashes] 5 corpus files -> 22 distinctive tokens
[DeriveDenyHashes] skills/_LIFEOS absent (public install) — skipping hash write
$ ls ~/.claude/LIFEOS/USER/SECURITY/DENY_HASHES.json
ls: No such file or directory
```

Exit code 0, and the message reads like correct public-install behaviour.

## Impact

The hash half of the identity-leak guard is permanently empty on public installs. `SystemFileGuard` fails open on a missing `DENY_HASHES.json` by design, so nothing breaks loudly; the guard simply enforces on the hand-maintained plaintext list alone, which is the blocklist the hash derivation exists to compensate for.

[...]

PR #1964, opened 2026-08-24
fix(security): DeriveDenyHashes skip predicate stale after #1689 relocation

0bsolescence:
## What

Since #1689 moved the deny-hash filter's home to `USER/SECURITY` and its consumer (`hooks/lib/system-file-guard-core.ts`) ships in `hooks/lib`, a public install with the consumer present has both a home for `DENY_HASHES.json` and a guard reading it — yet the write was still gated on `skills/_LIFEOS` existing (the pre-relocation dev-tree marker). Result: public installs silently never get the deny-hash filter even though the guard consuming it is installed and fails open.

**Commit 1** updates the skip predicate: only skip when the consumer is missing too.

**Commits 2–3** harden the write that the relaxed predicate newly enables on public installs. The USER root is legitimately a symlink (XDG mount, per SystemUserBoundary.md), but a statically crafted USER tree (restored backup, cloned repo) could carry `SECURITY/` as a symlink, or `DENY_HASHES.json` as a symlink or hard link, redirecting or truncating a file outside the tree. The tool now resolves the output dir, refuses a destination outside the resolved USER root, and writes temp-then-rename on the resolved dir so a crafted final component is replaced, never followed. Stated threat-model boundary: concurrent mid-run mutation of the tree means the attacker already has code execution as the user, and is out of scope.

## Testing

- Scratch-tree harness: legitimate XDG layout (`LIFEOS/USER` itself a symlink) writes normally; `SECURITY/` symlinked out of tree is refused; `DENY_HASHES.json` hard-linked to a victim file leaves the victim untouched while the output replaces the directory entry.
- Patched tool parses clean under bun.

Developed with AI assistance (Claude), reviewed and tested by the author.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/DeriveDenyHashes.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/DeriveDenyHashes.ts
@@ -213,8 +213,13 @@ function main(): void {
   // of exiting 1 on every derivedsync run (public issue #1488). The consumer
   // (hooks/lib/system-file-guard-core.ts) already fails open on a missing
   // DENY_HASHES.json, so nothing is lost.
-  if (!existsSync(join(CLAUDE, "skills", "_LIFEOS"))) {
-    console.log("[DeriveDenyHashes] skills/_LIFEOS absent (public install) — skipping hash write");
+  // Fix 2026-08-23: since #1689 the filter's home is USER/SECURITY and the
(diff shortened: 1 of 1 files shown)


danielmiessler:
Commit 1 landed via #1902 on 08-19. Your later #2017 is the better version of commits 2 and 3 (the predictable tmp name), so I ported that shape instead: tmp + O_EXCL + fsync + rename. Closing, credited on #2017. Thanks.

Issue #1690, opened 2026-07-28
lifeos.ts: unknown flags are silently dropped instead of erroring or being forwarded to claude


catchingknives:
## Summary

The launcher's argument parser hard-errors on unknown commands but silently ignores any unknown token that starts with `-`. Every flag the launcher doesn't recognize simply vanishes: it is neither an error nor forwarded to the `claude` invocation the launcher builds.

## Where

`LifeOS/install/LIFEOS/TOOLS/lifeos.ts`, `main()` parser, default case (lines 748-753 on current `main`):

```ts
default:
  if (!arg.startsWith("-")) {
    // Might be an unknown command
    error(`Unknown command: ${arg}. Use 'k help' for usage.`);
  }
```

Unknown words (`k foo`) exit loudly with "Unknown command", but unknown flags (`k --foo`) fall through the switch with no action at all.

## Why it bites

The launcher wraps the real CLI, so it's natural to assume claude-native flags pass through. They don't, and nothing tells you:

- `k -r --fork-session` opens the resume picker but drops `--fork-session`, so you continue the ORIGINAL session while believing you're working in a fork. That one is state-mutating: divergent work lands in the session you meant to preserve.
- `k --continue`, `k --model <id>`: silently dropped.
- Typos are inconsistent with the command path: `k --locl` silently launches with default behavior, while the equivalent bad command (`k locl`) errors loudly.
- The file's own comment block advertises a `--dangerous` flag that the parser never implements, so even the launcher's self-documented flag is swallowed. That documentation divergence is filed as its own issue to keep this one about parser policy.

## Suggested fix

Two complementary changes:

1. **Hard-error on unknown dash-flags**, mirroring the existing unknown-command behavior: `error("Unknown flag: " + arg + ". Use 'k help' for usage.")`. This kills the silent-typo class.
2. **Add an explicit `--` passthrough**: everything after a literal `--` is appended verbatim to the `claude` argv in `cmdLaunch()`. That gives full access to claude-native flags (`k -r -- --fork-session`) without the launcher having to mirror the whole upstream flag surface forever.

Auto-forwarding unknown flags would also work, but then a launcher typo is indistinguishable from an intentional claude flag. The error-plus-`--` combination keeps both safety and capability.

Happy to PR this if the approach sounds right.

danielmiessler:
[...]

Issue #1526, opened 2026-07-17
config.paths.memoryDir is declared but unwired — memory root always resolves to LIFEOS/MEMORY


christauff:
`LifeosConfig.ts` defines and populates `paths.memoryDir` (defaulting to `~/.claude/LIFEOS/MEMORY`), but nothing consumes it. The path resolver (`hooks/lib/paths.ts` — `getLifeosDir()` / `paiPath()`) and every memory reader hardcode `getLifeosDir()/MEMORY`; the config value is dead. (Verified: no references to `.memoryDir` outside its own definition, local tree + code search.)

### Why it's worth wiring

1. **Non-destructive upgrades.** When the resolved root moves across a release — as in PAI→LifeOS, where `getLifeosDir()` went from `~/.claude` to `~/.claude/LIFEOS`, moving the memory root `~/.claude/MEMORY` → `~/.claude/LIFEOS/MEMORY` — accumulated memory under the old root is silently stranded: readers point at the new (empty) location while the corpus sits in the old one. A configurable, resolver-independent memory root makes upgrades non-destructive for users with accumulated state.
2. **Data/framework separation.** Lets users keep heritable memory outside the framework-managed tree (backup, portability, surviving a reinstall that clears `LIFEOS/`).

### Scope (so you can judge before any code)

~171 call sites: **33** via `paiPath('MEMORY', …)` (central) and **138** inline `join(…, 'MEMORY', …)`. It must be all-or-nothing — wiring only the central sites leaves the inline readers pointed elsewhere, i.e. a split memory tree.

### Proposed approach

A `getMemoryDir()` in `paths.ts` returning `config.paths.memoryDir ?? join(getLifeosDir(), 'MEMORY')` (default installs stay byte-identical), then a codemod routing the inline sites through it; optionally a `LIFEOS_MEMORY_DIR` env var for parity with `LIFEOS_DIR`.

Happy to do the codemod + tests if you want the feature and agree with the shape — flagging first rather than dropping a 171-site change to core path resolution into your queue unsolicited.

danielmiessler:
[...]

PR #1973, opened 2026-08-24
feat(lifeos): make the launch directory configurable via a [launch] config block

Drizzt321:
### The problem

`lifeos.ts` hardcodes `process.chdir(CLAUDE_DIR)`, so every session starts in `~/.claude` regardless of where the terminal was. That means the Bash tool, `EventLogger`'s git snapshot, `ForgeProgress`'s `codex exec --cd`, `SecretScan`'s default target, and Serena's project matching all operate on the LifeOS install instead of the project the human is working in. The workaround today is `--local` on every invocation, or a local patch to the launcher.

### The default reproduces today's behaviour byte for byte

An install that does not opt in behaves exactly as v7.40.4 does now. All four of these resolve to `~/.claude`, the current hardcoded `CLAUDE_DIR`:

| Situation | Resolves to |
|---|---|
| no `LIFEOS_CONFIG.toml` on disk at all | `~/.claude` |
| TOML present, no `[launch]` block | `~/.claude` |
| `[launch]` present, `cwd_mode` absent | `~/.claude` |
| `[launch]` present, `cwd_mode = "config-root"` | `~/.claude` |

The built-in defaults are `cwdMode: "config-root"` and `defaultDir: <config root>`. `default_dir` is not a new place to land; it is a name for the place the launcher already lands, made overridable.

The template ships the block live, with those defaults written out rather than commented. That matches how the rest of the file ships (`[principal]`, `[da]`, `[paths]` all carry real values), it makes the knob discoverable, and it means the parse and validation path runs on every install so a regression there fails loudly instead of silently falling back to defaults. An older config with no `[launch]` block behaves identically.

### What this adds

A `[launch]` table in `LIFEOS_CONFIG.toml`:

```toml
[launch]
cwd_mode        = "stay-if-permitted"
default_dir     = "~/.claude"
permitted_roots = ["@config-root", "@config-paths"]
```

| `cwd_mode` | Behaviour |
|---|---|
| `config-root` | cd to `default_dir`. Stock, and the default when the key is absent. |
| `stay` | Stay in the launch directory, always. |
| `stay-if-permitted` | Stay when the launch directory is under a permitted root, otherwise cd to `default_dir`. |

Precedence: `--local` -> `--home` -> `LIFEOS_CWD_MODE` -> `[launch].cwd_mode` -> built-in `config-root`. The env var name follows the existing `LIFEOS_ALLOW_ROOT` convention in the same file. An empty `LIFEOS_CWD_MODE` counts as unset; an unrecognised one is a hard error rather than a silent fallback to stock.

`--home` is new and forces `default_dir` regardless of cwd. No short form, since `-h` is taken by `--help`. It is registered in `KNOWN_FLAGS`, without which the passthrough branch would swallow it and forward it to `claude` verbatim - the failure mode of issue #1690 (@catchingknives).

`permitted_roots` is consulted only in `stay-if-permitted` mode, and takes literal paths plus `@`-sentinels:

| Sentinel | Expands to |
|---|---|
| `@config-root` | the config root (`~/.claude`) |
| `@config-paths` | `[paths].user_dir` and `[paths].projects_dir` |
| `@settings-allow` | roots parsed from the merged `settings.json` `permissions.allow` `Edit()` rules |

Matching is lexical and prefix-based against the visible path, with the trailing separator re-appended, so `~/Projects` matches `~/Projects/thing` but not `~/Projects-old`. No globbing, and no realpath on either side: resolving would make a symlinked project directory stop matching the root the user typed. `@` is reserved as a prefix, so a literal path starting with `@` is not supported.

### Design notes

**`resolveLaunchDir()` is pure.** Everything arrives as an argument, so the whole decision table is exercisable as string logic with no fixtures - the concern issue #1526 raised about `memoryDir`, where a knob looked live but was not.

**Validation happens at config load, regardless of the active mode.** A typo in `permitted_roots` surfaces at the next launch rather than weeks later when the mode changes, and each message names the offending key. Rejected: any `..` segment, a relative path, anything containing `*`, an unknown `@sentinel`, and an allowlisted variable that is unset or empty. That last one is a hard error because naive expansion of an unset `${LIFEOS_DIR}` yields `""`, turning `${LIFEOS_DIR}/x` into `/x`, a root-level prefix matching nearly everything.

**Existence is not checked at load.** Config load validates syntax, `chdir` validates reality. `chdir` is atomic, one syscall instead of two, and distinguishes `ENOENT` / `ENOTDIR` / `EACCES` where a boolean cannot. It also means a `default_dir` on an automounted or network path does not fail validation while it is legitimately absent.

**`CLAUDE_DIR` is unchanged.** It appears eight times and seven of those locate the install itself: `Doctor.ts`, `LIFEOS_SYSTEM_PROMPT.md`, `MCP_DIR`, `ACTIVE_MCP`, `BANNER_SCRIPT`. Only one use is the launch destination. Conflating the two would mean setting `default_dir` accidentally relocates where the launcher looks for its own system prompt.

**`expandHome` is exported** from `LifeosConfig.ts` rather than reimplemented. It already normalises `~` for `[paths]`, `[launch]` carries paths with the same semantics, and a second implementation would drift.

**`main()` now sits behind `import.meta.main`,** the pattern `LifeosConfig.ts` already uses, so the pure helpers can be imported without launching a session. Its catch routes `Error`s through the existing `error()` formatter and keeps the raw dump for non-`Error` throws: `console.error(e)` prints the stack, burying the validation messages that name the offending key.

### The main-checkout guard, extended rather than worked around

PR #1579 (@asdf8675309) refuses to launch when `--local` would put a session in a repo's main checkout root while that repo uses `.claude/worktrees`, because the session would run on whatever stale branch the root is parked on.

This change re-keys that guard on the destination instead of on the flag. The hazard is where the session lands, not why. Keyed on `--local`, the guard would stop protecting anyone who sets `cwd_mode` to a staying mode - the population that most needs it.

Re-keying alone introduces one new failure, and the fix is part of this PR: the configured default destination is always exempt. Without that exemption the guard fires whenever the destination equals the cwd for any reason, including simply arriving at `default_dir`. That is reachable on a stock install with zero customization: `default_dir` defaults to `~/.claude`, that directory is commonly a git repo, and the tooling creates `.claude/worktrees` there as a side effect of agent isolation. Populate it once and a stock user standing in `~/.claude` under stock `config-root` mode is refused by the tool's own defaults.

The guard means "you asked to stay somewhere that is probably the wrong branch". Arriving at `default_dir` is not that. `LIFEOS_ALLOW_ROOT=1` still overrides everything.

The message drops "A --local session" for "A session", since the guard no longer implies that flag.

### The `PWD` commit

The first commit stands alone and can be taken or dropped independently.

`lifeos.ts` calls `process.chdir()` and then spawns, but never updates `PWD` in the child's environment, so the child receives a `PWD` naming the launch-time directory while its actual cwd is the config root. It is invisible today because bash self-heals: an inherited `PWD` that is a valid alias for the current directory is kept, one that is not is silently recomputed from `getcwd()`, and most tooling calls `getcwd()` anyway.

It is a minor pre-existing inconsistency rather than a live bug, fixed here because the feature depends on it. `getcwd()` always returns the resolved physical path - the kernel holds the cwd as a directory reference, not a string - so a symlinked route survives into the child only if `PWD` carries it. Both sites use `process.cwd()` rather than a captured variable, so they stay correct regardless of the destination.

The same distinction is why `visibleCwd()` reads `process.env.PWD ?? process.cwd()`. The fallback covers the direct `bun .../lifeos.ts` invocation `INSTALL.md` documents, where `PWD` may be absent. There is no realpath validation of `PWD`: `lifeos` is a shell alias, so bash sets it immediately before exec and there is no intermediate process to make it stale.

### Both launch paths

`cmdPrompt` (the one-shot `lifeos -p "..."` path) gets the same treatment as `cmdLaunch`. The launch directory is a property of the invocation, not of whether the session is interactive, so honouring `cwd_mode` in one path and not the other would be a bug rather than a scoping decision. Scripted callers are covered by the governing property: nothing changes for anyone who has not opted in, and a caller who has opted in is already in the directory they want.

`cmdPrompt` takes only `(prompt: string)` today, so it resolves with an empty flag set and config-driven modes do the work. Threading `--local` / `--home` through to it is an additive follow-up.

There is precedent for treating the two paths symmetrically, for the principle rather than for this decision specifically: the comment at the `launchEnv` site already says "Mirrors the protection in cmdPrompt() - same hazard, same fix" about stripping `ANTHROPIC_API_KEY` before spawn.

### Files changed

| File | Change |
|---|---|
| `LIFEOS/TOOLS/LifeosConfig.ts` | `expandHome` exported; `CwdMode` / `LifeosLaunch` types; `launch` on `LifeosConfig`; `loadLifeosLaunch()`; `[launch]` parse, validation, sentinel expansion |
| `LIFEOS/TOOLS/lifeos.ts` | `visibleCwd()`, pure `resolveLaunchDir()`, `mainCheckoutGuardApplies()`, `--home` plus `KNOWN_FLAGS` and help row, both chdir sites, both `PWD` assignments, `import.meta.main` guard and error formatting |
| `LIFEOS/DOCUMENTATION/Config/ConfigSystem.md` | a `[launch]` section covering modes, precedence, sentinels, "known root" semantics, validation, and the guard |
| `USER/CONFIG/LIFEOS_CONFIG.toml` | the `[launch]` block, shipped live with the stock defaults |

### Verification

Verified by hand on a live 7.40.4 install across ten launch scenarios: permitted roots, non-permitted directories, `--home`, `--local`, and `lifeos -p`. `lifeos -p "run pwd"` from a project directory reported that directory rather than the config root, confirming the one-shot path honours `cwd_mode`.

The guard was tested in all four directions against a real worktree created and removed for the purpose: main checkout with worktrees while staying refuses; the same with `LIFEOS_ALLOW_ROOT=1` launches; inside a worktree launches; and a launch whose destination is `default_dir` is exempt.

The decision table was also driven through the pure resolver locally, covering: no config present, `cwd_mode` absent, each flag against each mode, `~/Projects-old` against a `~/Projects` root, cwd equal to a root exactly, each sentinel, an unknown sentinel, each rejected path form, env-versus-TOML precedence, and every direction of the guard including both default-dir exemption cases.

No test file is included. The repository ships none, and adding the first one seemed like a separate conversation to have with you rather than something to bundle into a feature. The resolver being pure is what makes that table cheap to run, and I am happy to add the file here or in a follow-up if you want it.

### DIFF
--- a/LifeOS/install/LIFEOS/DOCUMENTATION/Config/ConfigSystem.md
+++ b/LifeOS/install/LIFEOS/DOCUMENTATION/Config/ConfigSystem.md
@@ -61,6 +61,52 @@ LifeOS configuration follows the **system/user separation** contract (`LIFEOS/DO
 3. **Skills** — private `_*` skills that need credentials/integration data read `LIFEOS_CONFIG.toml` via `LifeosConfig.load()` (e.g. a home-automation skill reads its Homebridge token; the network skill reads UniFi creds).
 4. **CLAUDE.md `@`-imports** — at session start, CC loads files referenced by top-level `@`-imports in `CLAUDE.md` (ARCHITECTURE_SUMMARY, PRINCIPAL_TELOS, PRINCIPAL_IDENTITY, DA_IDENTITY, PROJECTS, OPERATIONAL_RULES). CC does NOT follow transitive `@`-imports from inside imported files, so identity files must be listed in `CLAUDE.md` directly.
 
+## The `[launch]` block — where a session starts
+
+`LIFEOS_CONFIG.toml` carries an optional `[launch]` table that decides which directory `lifeos` puts a session in. Historically the launcher hardcoded `process.chdir(CLAUDE_DIR)`, so every session landed in `~/.claude` regardless of where the terminal was — which meant the Bash tool, `EventLogger`'s git snapshot, `ForgeProgress`'s `codex exec --cd`, `SecretScan`'s default target, and Serena's project matching all operated on the LifeOS install rather than the project actually being worked in.
+
+**The governing property: the default reproduces stock behaviour byte for byte.** No config file, no `[launch]` block, or an absent `cwd_mode` all resolve to "cd to the config root". Nobody's behaviour changes until they opt in.
+
+### Modes
+
+| `cwd_mode` | Behaviour |
+|---|---|
+| `config-root` | cd to `default_dir`. STOCK, and the default when the key is absent. |
+| `stay` | Stay in the launch directory, always. |
+| `stay-if-permitted` | Stay when the launch directory is under a permitted root, otherwise cd to `default_dir`. |
+
+`default_dir` defaults to the config root (`~/.claude`). It is the configured landing directory: set it to anything else and both `config-root` mode and the `--home` flag follow it.
+
+### Precedence
+
+`--local` → `--home` → `LIFEOS_CWD_MODE` env var → `[launch].cwd_mode` → built-in `config-root`. `--local` forces staying in the current directory; `--home` forces `default_dir`. An empty `LIFEOS_CWD_MODE` counts as unset; an unrecognised one is a hard error rather than a silent fallback.
+
+### Permitted roots
+
+`permitted_roots` is consulted only in `stay-if-permitted` mode, and accepts literal paths and `@`-sentinels:
+
+| Sentinel | Expands to |
+|---|---|
+| `@config-root` | the config root (`~/.claude`) |
+| `@config-paths` | `[paths].user_dir` and `[paths].projects_dir` |
+| `@settings-allow` | roots parsed from the MERGED `settings.json` `permissions.allow` `Edit()` rules |
+
+`@settings-allow` matches an `Edit(<path>/**)` rule whose `<path>` is ROOTED — it begins with `~/` or `/`. Rules carrying a wildcard anywhere other than the trailing `/**` are SKIPPED, because a leading `**/` form is a pattern rather than a rooted path and treating it as a root would add a meaningless entry; relative rules are skipped for the same reason. The merged `settings.json` is the right input rather than `settings.user.json`, since the effective permission set is what the user actually experiences.
+
+**"Known root" semantics.** Matching is lexical and prefix-based against the VISIBLE path (`$PWD`, not `process.cwd()`), with the trailing separator re-appended — so `~/Projects` matches `~/Projects/thing` but NOT `~/Projects-old`. A directory equal to a root exactly counts as under it. There is no globbing, and no realpath on either side: resolving would make a symlinked project directory stop matching the root the user actually typed. An ABSENT `permitted_roots` falls back to `["@config-root", "@config-paths"]`; an EMPTY one (`[]`) means nothing is permitted, so `stay-if-permitted` always lands on `default_dir`.
+
+### Validation
+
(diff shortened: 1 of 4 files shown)


danielmiessler:
Passing. 577 lines of config grammar for what `lifeos --local` (already there) or a three-line env check gives. Closing, thanks for the offer.

Drizzt321:
@danielmiessler so your primary issue is that it's a big config grammar? Or that I should just use `--local` always? My personal desire was to effectively always be `--local`, but only if it was a directory I already had OKed via the claude code security, and if not to have a startup dir that wasn't `~/.claude`. Any suggestions on a way that would smell better to you for upstreaming, instead of maintaining for my own setup every time after an upgrade?

PR #1976, opened 2026-08-25
fix(inference): pin verbose off so --output-format json stays a single object

Drizzt321:
With verbose view mode on (`"verbose": true` in settings.json, `/config verbose=true`, or `--verbose`), `claude --print --output-format json` returns a JSON array of session messages instead of the documented single result object (anthropics/claude-code#84784).

`inferenceAttempt()` reads `envelope.result` off the parsed value, so on those installs every call fails with `Claude JSON envelope missing result field`, taking down every `Inference.ts` consumer at once.

`--setting-sources ''` is already passed and does not suppress it (measured on CLI 2.1.245). Command-line settings outrank the user settings file, so this pins `verbose: false` for the child process. Three lines.

Reproduce:

    claude -p hi --output-format json | head -c 1             # {
    claude -p hi --output-format json --verbose | head -c 1   # [

Managed settings outrank `--settings`, so an MDM-forced verbose install would still see an array. A shape-tolerant parse closes that if that's a better way to go.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/Inference.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/Inference.ts
@@ -260,6 +260,9 @@ async function inferenceAttempt(options: InferenceOptions, modelOverride?: strin
       '--effort', config.effort,  // Opus 4.8 respects effort strictly; tune intelligence vs. token spend per level
       ...(hasImages ? ['--allowedTools', 'Read'] : ['--tools', '']),
       '--output-format', 'json',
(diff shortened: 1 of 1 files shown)


danielmiessler:
Ported with credit, folded into the same Inference.ts change as #1994: verbose pinned off via `--settings`, plus the array-tolerant envelope parse. One thing I couldn't verify here is the JSON arg surviving claude.cmd on Windows; if you're on Windows and it misbehaves, say so. Closing, thanks.

PR #1978, opened 2026-08-25
fix(SessionCleanup): derive ISA phase from claim state — a session ending is not a claim closing

### LINKED ISSUE #1977: SessionCleanup marks an ISA phase: complete on session end with no check that any claim closed

## TL;DR

`hooks/SessionCleanup.hook.ts` rewrites an ISA's `phase:` to `complete` when a session ends, with no check that any claim actually closed. A session that ends with work in flight marks its ISA finished, which removes it from the resume/continuity surfaces. On this install it silently hid 120 open claims across 5 ISAs.

## Evidence

The write is unconditional (v7.40.4 payload, `LifeOS/install/hooks/SessionCleanup.hook.ts`, line 89):

```ts
isaContent = isaContent.replace(/^phase:.*$/m, 'phase: complete');
```

Nothing above or below it inspects claim state. `status: ACTIVE` → `COMPLETED` and `completed_at` are written in the same block.

Observed on a real install: an ISA committed at 09:51 with `phase: scoping` and 0 of 21 claims closed read `phase: complete` by 11:09, with no human edit between. Sweeping every `phase: complete` ISA in the tree for unchecked claims found 5 such files, carrying 120 open claims between them:

| claims closed | claims open | `progress:` would read |
|---|---|---|
| 1 | 45 | 1/46 |
| 5 | 23 | 5/28 |
| 0 | 16 | 0/16 |
| 1 | 15 | 1/16 |
| 0 | 21 | 0/21 (the one observed flipping) |

Two of them have zero claims closed and read as finished. A stale `progress:` field compounds it: one ISA read `progress: 5/16` against 28 real claims, because `progress:` is not refreshed either.

## Impact

Any session that ends before its claims close marks that ISA complete. Because the continuity surfaces filter on phase, the work then does not resurface at the next session start — the exact failure the ISA phase field exists to prevent. It is silent: no warning is printed and the ISA still reads as valid.

## Reproduction

1. Create an ISA with `phase: scoping` and at least one `- [ ]` claim.
2. Start a session against it, then end the session.
3. Read the frontmatter: `phase: complete`, claims still unchecked.

The payload's own expression reproduces it in isolation:

```python
re.sub(r'^phase:.*$', 'phase: complete', isa, count=1, flags=re.M)
```

against a fixture with two open claims yields `phase: complete` with both claims still open.

## Suggested fix

Derive the phase from claim state rather than from the session ending, and refresh `progress:` while the ISA stays open:

```ts
const open   = (isaContent.match(/^- \[ \]/gm) || []).length;
const closed = (isaContent.match(/^- \[x\]/gm) || []).length;
const phase  = open > 0 ? (closed > 0 ? 'climbing' : 'scoping') : 'complete';
```

Only write `status: COMPLETED` / `completed_at` when `open === 0`, and log when the hook leaves an ISA open so the state is visible rather than assumed.

Local patch applied here and mutation-tested: reverting the logic to the unconditional form is caught by the test, so the check is not vacuous.

## Environment

LifeOS v7.40.4, macOS, bun. File is stock and unmodified prior to the patch.

bnkath2o:
Fixes #1977.

## What this changes

`clearSessionWork()` rewrote `phase:` to `complete` whenever a session ended, with no inspection of claim state. Any session that ended with work in flight marked its ISA finished, and because the continuity surfaces filter on phase, that work then did not resurface at the next session start.

This derives the phase from the claims instead:

```ts
const openClaims   = (isaContent.match(/^- \[ \]/gm) || []).length;
const closedClaims = (isaContent.match(/^- \[x\]/gm) || []).length;
const phase = openClaims > 0
  ? (closedClaims > 0 ? 'climbing' : 'scoping')
  : 'complete';
```

Three further consequences fall out of the same condition:

- `status: COMPLETED` and `completed_at` are written **only** when `openClaims === 0`. Previously an ISA could carry a completion timestamp with every claim still open.
- `progress:` is refreshed while the ISA stays open, so it cannot go stale. On this install one ISA read `progress: 5/16` against 28 real claims.
- The hook logs to stderr when it leaves an ISA open, so the state is visible rather than assumed.

## Why this shape

`scoping` / `climbing` / `complete` is the vocabulary the shipped ISA documentation already uses — 33, 10 and 36 occurrences in `LIFEOS/DOCUMENTATION/ISA/`. Worth noting that no shipped hook or tool currently writes `scoping` or `climbing`; only `complete`, `plan` and `execute` appear in shipped code. So this is the first place the mid-flight vocabulary is actually written, which is why the bug existed at all: the hook only ever had one phase to write.

The check is deliberately a condition inside a procedure that already runs, rather than a new gate. It costs nothing on any session that ends with its claims closed.

## Verification

- The phase derivation was tested against four cases: all claims open, some closed, all closed, and no claims at all.
- **Mutation-tested.** Reverting the logic to the unconditional `'complete'` form is caught by the test, so the check is not vacuous.
- The file builds clean under `bun build`.
- Applied on a live install: re-sweeping every `phase: complete` ISA for unchecked claims returns zero, where it returned five before the patch.

## Not claimed

No estimate of how many installs are affected. The write is unconditional, so any install ending a session mid-work should reproduce it, but that is reasoning rather than measurement.

### DIFF
--- a/LifeOS/install/hooks/SessionCleanup.hook.ts
+++ b/LifeOS/install/hooks/SessionCleanup.hook.ts
@@ -86,12 +86,31 @@ function clearSessionWork(sessionId?: string): void {
       // findArtifactPath prefers ISA.md and falls back to legacy PRD.md.
       if (isaPath && existsSync(isaPath)) {
         let isaContent = readFileSync(isaPath, 'utf-8');
-        isaContent = isaContent.replace(/^phase:.*$/m, 'phase: complete');
+
+        // A session ending is not a claim closing. Derive the phase from claim
+        // state so work in flight keeps a phase the continuity surfaces can see.
+        const openClaims = (isaContent.match(/^- \[ \]/gm) || []).length;
+        const closedClaims = (isaContent.match(/^- \[x\]/gm) || []).length;
+        const phase = openClaims > 0
+          ? (closedClaims > 0 ? 'climbing' : 'scoping')
+          : 'complete';
+
+        isaContent = isaContent.replace(/^phase:.*$/m, `phase: ${phase}`);
         isaContent = isaContent.replace(/^updated:.*$/m, `updated: ${getISOTimestamp()}`);
-        isaContent = isaContent.replace(/^status: ACTIVE$/m, 'status: COMPLETED');
-        isaContent = isaContent.replace(/^completed_at: null$/m, `completed_at: "${getISOTimestamp()}"`);
+        if (openClaims > 0) {
+          // progress must not go stale while the ISA stays open
+          isaContent = isaContent.replace(
+            /^progress:.*$/m, `progress: ${closedClaims}/${closedClaims + openClaims}`);
+        } else {
+          isaContent = isaContent.replace(/^status: ACTIVE$/m, 'status: COMPLETED');
+          isaContent = isaContent.replace(/^completed_at: null$/m, `completed_at: "${getISOTimestamp()}"`);
+        }
         writeFileSync(isaPath, isaContent, 'utf-8');
         marked = true;
+        if (openClaims > 0) {
+          console.error(`[SessionCleanup] ${isaPath}: left phase: ${phase} — ` +
+            `${openClaims} claim(s) still open. A session ending does not close claims.`);
+        }
       }
 
       // Legacy fallback: update META.yaml if it exists


danielmiessler:
Ported with credit, one extension: the derived phase also gates the work.json row close, otherwise the ISA reads climbing while the dashboard derives done. Closing, next release. Thanks.

PR #2015, opened 2026-08-29
fix(interceptor): publish captures from staged save

mj-deving:
## Summary

- stop passing unsupported `--out` to `interceptor screenshot`
- run `--json ... --save` from a private same-filesystem staging directory
- parse and validate the returned `filePath`, reject blank captures, and publish with atomic rename semantics
- preserve an existing destination when capture, validation, or publication fails

## Why

Interceptor 0.23.30 exposes `--save` for screenshots and returns the actual `filePath`; it does not expose screenshot `--out`. The wrapper currently forwards `--out`, so the required screenshot verification fails before producing an image.

## Testing

- `bash -n LifeOS/install/skills/Interceptor/Tools/Capture.sh`
- `shellcheck LifeOS/install/skills/Interceptor/Tools/Capture.sh`
- targeted 11-case regression harness covering DOM and forced-pixel invocation, custom/default destinations, JSON escaping, blank rejection, cleanup, preflight failure, publication failure, untrusted `filePath`, and destination-symlink substitution
- live Interceptor 0.23.30 DOM capture produced a valid 3034×1680 PNG at the requested path
- verified pinned Interceptor commit `fe94108` writes successful DOM and pixel `--save` payloads in the CLI working directory

### DIFF
--- a/LifeOS/install/skills/Interceptor/Tools/Capture.sh
+++ b/LifeOS/install/skills/Interceptor/Tools/Capture.sh
@@ -183,10 +183,31 @@ if [ -z "$OUT" ]; then
     rand="$$"
     OUT="${LIFEOS_DOWNLOADS_DIR:-${HOME}/Downloads}/interceptor-capture-${ts}-${rand}.png"
 fi
-mkdir -p "$(dirname "$OUT")"
+OUT_BASE="$(basename "$OUT")"
+if [ "$OUT_BASE" = "." ] || [ "$OUT_BASE" = ".." ] \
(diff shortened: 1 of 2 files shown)


danielmiessler:
You were right that `--out` was never parsed upstream and that newer CLIs will reject it, so this is in with credit, in an eight-line shape: `--out` dropped, each screenshot runs in a mktemp cwd with an exit trap, the existing move logic kept. I didn't take the containment check or the no-rm-on-retry, both of which would have regressed the pixel-truth rule. Closing, next release. Thanks.

PR #2020, opened 2026-08-29
fix: validate documentation section anchors

mj-deving:
## Summary

- add section-anchor validation for canonical Markdown references to `ReferenceCheck`
- make missing, missing-target, and ambiguous anchors blocking in `IntegrityCheck`
- repoint stale references to their canonical upstream homes and delimit exact section addresses without adding principal-specific doctrine
- make `--changed` catch target renames/deletions and untracked referrers

## Why

File existence checks do not catch a `File.md § Section` reference after the target heading is renamed or moved. This adds a bounded Markdown-heading integrity layer while preserving the upstream owner boundary: canonical docs remain authoritative, USER rules remain an overlay, and ambiguous bare targets require path qualification.

## Verification

- `bun LifeOS/install/LIFEOS/TOOLS/lib/section-anchors.ts` — 30/30 positive and negative contracts pass
- isolated install payload — `missing_anchor: 0`
- injected missing section — blocking `missing_anchor` emitted
- changed target heading — unchanged referrer emitted under `--changed`
- renamed/deleted target — unchanged referrer emitted as `missing_target`
- untracked referrer — emitted under `--changed`
- `git diff --check` — clean
- structured Codex autoreview — no accepted or actionable findings

## Scope

This intentionally validates Markdown headings, not TypeScript symbols. Fenced examples and retired documents with explicit file-level retirement markers are excluded; inline `File.md § Section` code spans are treated as exact live references.


danielmiessler:
Passing on this shape. +382/-47 across 27 files, missing_anchor becomes blocking against a tree that's hundreds of commits past its base, it rewrites the system prompt, and it touches a hook that no longer exists (ModelRungGuard was renamed). If anchor validation comes back it'll be a small advisory lane inside ReferenceCheck, no doc repoints. Thanks for the thoroughness.

Issue #1874, opened 2026-08-16
Setup.md step 9(d) substituteTree(configRoot) rewrites InstallEngine.ts and permanently breaks checkSurvivingPlaceholders


GRyndStone:
Following `Workflows/Setup.md` step 9(d) exactly rewrites the installer's own source and breaks the verifier it tells you to run next.

### The instruction

> run the substitution pass the step-4 pass could not reach — `substituteTree(<configRoot>, vars)` over the deployed tree — then `checkSurvivingPlaceholders(<configRoot>)` and require `passed: true`

### What happens

`substituteTree`'s walk skips `node_modules`, `.git`, `MEMORY`, and — correctly, per #1828 — `skills/LifeOS/install`. It does **not** skip `skills/LifeOS/Tools/`, and `Tools/InstallEngine.ts` holds the placeholder tokens as string literals. Rooted at `<configRoot>`, the pass rewrites its own source:

```js
// before
const IDENTITY_PLACEHOLDERS = [
  "{{DA_NAME}}", "{{DA_FULL_NAME}}", "{{PRINCIPAL_NAME}}", "{{PRINCIPAL_FULL_NAME}}",
  "{{PRIMARY_VOICE_ID}}", "{{SECONDARY_VOICE_ID}}", "{{LIFEOS_VERSION}}",
] as const;

// after
const IDENTITY_PLACEHOLDERS = [
  "Kuru", "Kuru", "Cal", "Cal",
  "21m00Tcm4TlvDq8ikWAM", "pNInz6obpgDQGcFmaJgB", "7.40.4",
] as const;
```

`checkSurvivingPlaceholders` then searches the tree for the *substituted values*, so it reports tens of thousands of "placeholders" and can never pass again. `Workflows/Setup.md` itself and the doc prose in `DOCUMENTATION/Pulse/DaSubsystem.md` are rewritten the same way. Recovery is restoring `Tools/InstallEngine.ts` and `Workflows/Setup.md` from the release tarball.

### Second, quieter problem

`substituteTree` keys must include the braces. A caller passing `{ PRINCIPAL_NAME: "Cal" }` hits the `/^\{\{[A-Z0-9_]+\}\}$/` guard and every entry is skipped, returning `{ scanned: 15035, modified: 0, applied: 0 }` — success-shaped output that changed nothing. The guard is right (it prevents #1484), but the required key shape is not stated in the step-9(d) instruction that tells you to build `vars`.

### Suggested fix

[...]

Issue #1828, opened 2026-08-12
Install issue / name substitutions


Piroshki:
Diving in for the first time - just wanted to flag the error message, which Claude fixed, that I got upon install:

Worth flagging plainly: my first substitution pass ran too broadly and briefly corrupted the LifeOS skill's own source code (baking your name into ~119 installer/doc/hook files that must stay generic). I caught it via the verification gate, pulled a clean copy of v7.28.3 from danielmiessler/LifeOS on GitHub, and restored every affected file — confirmed byte-for-byte identical to upstream afterward. Nothing user-facing was left broken, but it's the kind of thing worth knowing happened.

Exicted to get it going.

(Edit: also noticed that a bunch of hooks got permission denied flags. Claude fixed them)

DRAZY:
Hit both of these independently on a 7.1.1 → 7.28.3 reconciliation today, and I can offer a root cause for each. Your "~119 files" number is almost exactly reproducible, which I think confirms the mechanism.

## 1. Name substitution reaching the installer's own source

`substituteTree()` skips very little:

```js
// skills/LifeOS/Tools/InstallEngine.ts
const SKIP_DIRS = new Set(["node_modules", ".git", "MEMORY"]);
```

`Workflows/Setup.md` step 9(d) instructs the installer to run `substituteTree(<configRoot>, vars)` over the whole config root. On a normal install the shipped payload lives at `<configRoot>/skills/LifeOS/install/`, which is *inside* that root and is not in `SKIP_DIRS`. So the identity pass walks the release payload as if it were deployed content and renders the template tokens it is supposed to leave alone.

Count on a clean v7.28.3 payload:

```console
$ grep -rl '{{DA_NAME}}\|{{PRINCIPAL_NAME}}\|{{PRINCIPAL_FULL_NAME}}\|{{DA_FULL_NAME}}\|{{LIFEOS_VERSION}}\|{{PRIMARY_VOICE_ID}}\|{{SECONDARY_VOICE_ID}}' \
    ~/.claude/skills/LifeOS/install | wc -l
121
```

121 files carry identity tokens, against the ~119 you saw. The consequence is quieter than the visible error, and worse: once the payload is substituted, it is no longer a clean template, so a later `Update` compares against personalised sources and the corruption is inherited rather than corrected.

Workaround that avoids it entirely, if useful to anyone landing here: move the payload outside the config root for the duration of the pass, then move it back.

```bash
[...]

PR #2022, opened 2026-08-29
fix: substituteTree rewrites its own engine source; unify walk with
  checkSurvivingPlaceholders (#1993)

### LINKED ISSUE #1993: Setup 9d: full-tree substituteTree(configRoot) rewrites InstallEngine's own IDENTITY_PLACEHOLDERS — identity check then scans for literal names (and empty string)
### Environment
LifeOS 7.40.4, fresh install · Windows 11 · bun 1.4.0 · Claude Code harness

### Symptom
After running the Setup step 9(d) substitution exactly as prescribed (`substituteTree(<configRoot>, vars)` over the deployed tree), `checkSurvivingPlaceholders` inside the *same process* passes:

```
substituted: { scanned: 5153, modified: 113, applied: 275 }
survivors: { "passed": true, "files": [], "total": 0 }
```

…but the **next** `Doctor.ts` run reports:

```
❌ Identity placeholders rendered — broken
   2526336 unrendered identity placeholder(s) across 258 site(s)
   E.g. Vlad in ...\hooks\AgentInvocation.hook.ts;  in ...\hooks\AgentInvocation.hook.ts
```

Note the "placeholders" it lists: the literal substituted value `Vlad`, and the **empty string** (my `{{PRIMARY_VOICE_ID}}`/`{{SECONDARY_VOICE_ID}}` were substituted to `""` because voice is not configured). Empty-string matching is what produces the 2.5M count.

### Root cause
`skills/LifeOS/Tools/InstallEngine.ts` itself lives inside configRoot and is a template-extension file, so the full-tree pass rewrites its own source, including the checker's constant:

```ts
// before
const IDENTITY_PLACEHOLDERS = ["{{DA_NAME}}", ..., "{{PRIMARY_VOICE_ID}}", ...];
// after substitution
const IDENTITY_PLACEHOLDERS = ["Rick", "Vlad", "", "7.40.4", ...];
```

The in-process check passes (the module was loaded before its file was rewritten), but every later import of the engine (Doctor) searches the tree for the literal *values*, including `""`, which matches everywhere.

The same pass also strips legitimate `{{TOKEN}}` literals from ~25 other `Tools/*.ts` sources that do their own runtime templating (ISARender, GenerateTelosSummary, RenderHermesSoul, …).

### Suggested fixes (any one suffices)
1. Build the constant so it can't be substituted in source: `const IDENTITY_PLACEHOLDERS = ["DA_NAME", "PRINCIPAL_NAME", ...].map(t => "{{" + t + "}}");`
2. Have `substituteTree` skip the engine/checker file(s), like it already skips `MEMORY/`.
3. Have `checkSurvivingPlaceholders` ignore empty-string entries defensively.

### Workaround used locally
Restore `InstallEngine.ts` from the payload after the substitution pass (and re-apply identity values to the other Tools sources). Doctor then reports identity OK correctly.

---
Found during a fresh Windows 11 install on 2026-08-26 (2 of 5 findings, see #1992). Happy to test patches.

rathko:
Fixes #1993 (same root cause family as #1874/#1852).

## What happens

Setup step 9(d) runs `substituteTree(<configRoot>, vars)` over the deployed tree. `InstallEngine.ts` itself lives inside that tree (`skills/LifeOS/Tools/`), is a template-extension file, and contained the literal identity tokens in its `IDENTITY_PLACEHOLDERS` table — so the pass rewrites the checker's own constants into the user's literal values. In-process verification still passes (the module was loaded before its file was rewritten), but every later import — `Doctor.ts`'s identity check — then scans the tree for the literal values: the user's name, and the **empty string** when a voice id was substituted to `""`. Split-on-empty-string matching is the 2.5M-survivors figure in #1993. Reproduced independently on a Linux 7.28.3→7.40.4 upgrade (370 false positives) before finding #1993.

Secondary: `checkSurvivingPlaceholders` had its own walk that skipped only `SKIP_DIRS` — not the redistribution payload — so it flagged the ~121 payload tokens `substituteTree` had (correctly, per #1828) left generic, contradicting Setup.md's "substitution and verification share one traversal".

## The fix (one file, mirrored into the payload copy)

1. **`IDENTITY_PLACEHOLDERS` is built from token names at runtime** (`"{{" + name + "}}"`), so this module's source never contains a substitutable literal. Self-immunity holds for every invocation shape — no traversal exclusion can protect a file that is the walk root.
2. **One shared `walkTemplateTree()`** used by both functions, so substitution and verification cannot drift. It keeps the #1828 payload skip, adds a skip for the engine's own source file (doc comments here legitimately show token examples; substituting them personalizes machinery and yields noisy diffs), and applies the same guard on the file-shaped-root path. Skips use path-**segment** matching, so `myskills/LifeOS` no longer false-matches a plain `endsWith`.
3. **Falsy-entry guard in the checker** as an invariant backstop against reintroducing the split-on-`""` explosion.

## Verified

Synthetic tree: user files render; `skills/LifeOS/install/` payload stays generic and is not flagged; a `myskills/LifeOS` decoy IS rendered; the engine file is byte-identical after the pass (sha256-checked); the checker passes from both `configRoot` and `skills/LifeOS` roots; a file-shaped root pointing at the engine is a no-op.

## Known tradeoffs (deliberate, flagged in cross-model review)

- A root pointed *directly at* `skills/LifeOS/install` still processes the payload — explicitly targeting the payload is treated as intent (same property as the pre-existing walk).
- The checker no longer scans `InstallEngine.ts` itself. That is by design: its comments legitimately carry token examples, and scanning it guaranteed false positives — the pre-fix checker always counted its own table as ≥7 survivors.
- The self-skip matches any `skills/LifeOS/Tools/InstallEngine.ts` path tail, including hypothetical fixture trees — same scope class as the existing #1828 payload predicate.

## Repair note for already-corrupted installs

7.40.4 setups that ran step 9(d): restore `skills/LifeOS/Tools/InstallEngine.ts` from the payload copy at `skills/LifeOS/install/skills/LifeOS/Tools/InstallEngine.ts`, then rerun Doctor.

### DIFF
--- a/LifeOS/Tools/InstallEngine.ts
+++ b/LifeOS/Tools/InstallEngine.ts
@@ -332,7 +332,7 @@ export function scanSettingsHooks(settingsPath: string): SettingsHookScan {
 // ════════════════════════════════════════════════════════════════════
 
 import { cpSync, lstatSync, mkdirSync, readdirSync, readlinkSync, renameSync, statSync, symlinkSync, unlinkSync, writeFileSync } from "node:fs";
-import { dirname } from "node:path";
+import { dirname, basename } from "node:path";
 
 // Extended 2026-07-25 (Forge finding, v7.15.0 re-audit). The set stopped at .ts,
 // so every Pulse component and the built Next bundle were invisible to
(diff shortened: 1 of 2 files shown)


danielmiessler:
#1993 was already closed by the skip list on 08-19, but your runtime-built placeholder table is cheap self-immunity, so that two-line idea is in with credit. Closing, thanks.

Issue #1848, opened 2026-08-15
## `retrieval-missing`: CortexHealth flags retrieval evidence, but nothing in the   shipped code writes it


jaylarr176:
Here is an issue I had upon upgrading to the newest version, the DA has created this writeup, hope that it is helpful!

  **Version:** LifeOS 7.40.4 / Cortex

  ### Summary

  `CortexHealth.ts` reports a `retrieval-missing` finding (`warn`) whenever
  `MEMORY/OBSERVABILITY/memory-retrievals.jsonl` doesn't exist or has no fresh
  row (`retrievalStaleMs` default: 24h, `CortexHealth.ts:16`). On a stock
  7.40.4 install this file never gets created, because no code path writes to
  it — the check exists, but its producer doesn't.

  ### Evidence

  - `CortexHealth.ts:206` reads `memory-retrievals.jsonl` and treats an absent
    file as `status: "missing"` (`CortexHealth.ts:93`), surfaced as:
    `No retrieval evidence within ${retrievalStaleMs}ms freshness window.`
  - Searching the full `LIFEOS/TOOLS/`, `skills/`, and `hooks/` trees for any
    writer of this file (`query_hash`/`returned_count`/`duration_ms` schema)
    turns up nothing. `MemoryStatus.ts` and `CortexHealth.ts` both *read* it;
    neither, nor anything else, ever *appends* to it.
  - Cortex's own search entry point (`skills/Cortex/Tools/ContextSearch.ts`)
    does not log a retrieval row when it runs a query.
  - `MEMORY/OBSERVABILITY/memory-retrievals.jsonl` is absent on disk entirely
    (not stale — never created).

  ### Impact

  `warn` severity only — can't push the overall Cortex verdict to `critical`
  on its own (`overall` = critical if any critical finding, else warn if any
  warn, `CortexHealth.ts:99`). But the health check is effectively
  unsatisfiable as shipped: there's no way for a stock install to ever clear
  `retrieval-missing`, since nothing produces the evidence it's checking for.

  ### Suggested fix

  ### Suggested fix

  Add a logging call at Cortex's retrieval call site(s) — at minimum
  `skills/Cortex/Tools/ContextSearch.ts` — that appends a row to
  `MEMORY/OBSERVABILITY/memory-retrievals.jsonl` matching the schema
  `CortexHealth.ts` already validates: `{ ts, query_hash, returned_count,
  duration_ms }`. Alternatively, if retrieval evidence isn't meant to be
[...]

PR #2023, opened 2026-08-30
fix: MemoryRetriever writes the retrieval evidence MemorySystem.md promises
  (#1910)

### LINKED ISSUE #1910: MemorySystem.md documents MemoryRetriever as the producer of memory-retrievals.jsonl, but it never writes it — CortexHealth's retrieval-missing warn can never clear

rathko:
Fixes #1910 (previously reported as #1848).

## What happens

`MemorySystem.md` documents MemoryRetriever as the producer of `MEMORY/OBSERVABILITY/memory-retrievals.jsonl` ("Per-turn retrievals | MemoryRetriever"), and CortexHealth warns `retrieval-missing` when no row is fresher than `retrievalStaleMs` — but nothing in the shipped code ever wrote that file. The warn can never clear on any install, while actual retrieval (the RELEVANT MEMORY grounding block) works fine, unlogged. The checker shipped without its producer.

## The fix

MemoryRetriever now appends one evidence row per retrieval, making the documented contract true:

- **Schema-exact:** `{ ts, query_hash, returned_count, duration_ms, top_score? }` — matches `CortexHealth.validRetrievalRow`'s exact-key validation, `top_score` present only when results exist.
- **Producer in the retriever core, so every caller logs:** the per-turn hot path (`getRelevantContext` — all exit paths including cache hits and empty results; freshness is the health signal, not hit quality) and the CLI path (including the empty-corpus exit, so a fresh install that only ever uses the CLI can still clear the warn).
- **Privacy:** query text is never stored — only a truncated sha256 of the trim+lowercase form. Short/low-entropy queries remain dictionary-guessable from the hash; same exposure class as the local transcript.
- **Fail-open:** a logging failure never breaks retrieval. First failure per process prints one stderr notice; CortexHealth's `retrieval-missing` remains the durable outer signal. On a read-only tree, behavior is unchanged.
- **Growth-bounded:** past 512KB the log trims to the newest 1000 rows (per-process tmp + atomic rename). CortexHealth reads the whole file each pass, so the bound keeps health checks cheap. Concurrent rotators can drop a few fresh rows — accepted for evidence (freshness over completeness).

Header updated: the tool stays read-only over memory CONTENT; its only writes are this observability evidence.

## Verified

End-to-end on a synthetic tree, judged by CortexHealth's own code: rows pass `validRetrievalRow` via `collectCortexEvidence` (retrieval.status `ok`, zero retrieval findings from `assessCortexEvidence`); hit, cache-hit, and empty-tokenization calls each produce a valid row; trim engages at the threshold (6000 rows → 1001); a read-only OBSERVABILITY dir neither throws nor changes retrieval results. Also running live on a 7.40.4 install: `retrieval-missing` cleared on the next health pass.

## Deliberate tradeoffs (flagged in cross-model review)

- Rotation favors freshness over completeness under concurrency (documented in-code).
- No symlink hardening on the append path — consistent with every other OBSERVABILITY JSONL writer in the tree; estate-wide hardening would be its own change.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/MemoryRetriever.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/MemoryRetriever.ts
@@ -34,7 +34,13 @@ for (const __k of ["LIFEOS_DIR", "LIFEOS_CONFIG_DIR", "PROJECTS_DIR"]) {
  * STORAGE:
  *   Reads MEMORY/KNOWLEDGE/{People,Companies,Ideas,Research}/*.md  (class: knowledge)
  *   Reads MEMORY/LEARNING/ ** /*.md recursively                    (class: learning)
- *   NEVER writes or modifies files — read-only tool.
+ *   NEVER writes or modifies memory content — read-only over the corpus.
+ *   Appends one evidence row per retrieval to
+ *   MEMORY/OBSERVABILITY/memory-retrievals.jsonl (observability, not content —
+ *   the producer contract MemorySystem.md documents and CortexHealth's
+ *   `retrieval-missing` check verifies; public issue #1910). Writes are
+ *   confined to MEMORY/OBSERVABILITY and fail-open: on a read-only tree the
+ *   tool behaves exactly as before (one stderr notice, results unchanged).
  *
  * ============================================================================
  */
@@ -44,6 +50,7 @@ import * as fs from "fs";
 import * as path from "path";
 import { spawnSync } from "child_process";
 import { homedir } from "node:os";
+import { createHash } from "node:crypto";
 
 // Normalize env path vars that Claude Code injects without shell expansion (LifeOS#1404)
 for (const k of ["LIFEOS_DIR", "LIFEOS_CONFIG_DIR", "PROJECTS_DIR"]) {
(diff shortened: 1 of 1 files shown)


danielmiessler:
The writer landed on 08-17 (#1848), so the part I took from this is the rotation: 512 KB → newest 1000 rows, per-process tmp and atomic rename, with credit. The log here was 174 KB and unbounded, so it earned its keep. Closing, thanks.

Issue #1861, opened 2026-08-15
Hermes Health.ts: false 'down' after a supervisor restart, 'degraded' for any idle gateway after 15 quiet minutes


hjbrandt:
**Version:** v7.40.4
**File:** `LifeOS/install/LIFEOS/HERMES/Health.ts`

Two related problems, both from treating `gateway_state.json` as ground truth for liveness:

**(a) False "down" after a supervisor restart.** Liveness is derived solely from the pid recorded in `gateway_state.json` (`pidAlive` → else status `"down"`). But that file is written by the gateway on its own cadence — after a `launchctl kickstart`-style restart, the file still holds the previous (dead) pid until the new process next writes state. During that window `Health.ts` reports the gateway down while the replacement process is demonstrably running (process table + supervisor both show it; the next state write clears the false alarm).

Suggested: when the recorded pid is dead but the supervisor still has the job loaded, cross-check the process table (or report `"restarting"`) instead of declaring `"down"` on the stale pid alone.

**(b) Idle gateways degrade after 15 minutes.** `STATE_STALE_MS = 15 * 60 * 1000`, and a live pid with stale state pushes a problem line — and any problem line flips status to `"degraded"`:

```ts
if (pidAlive && stateAgeMs !== null && stateAgeMs > STATE_STALE_MS) {
  problems.push(`gateway_state.json last written ${…}m ago`);
}
// …
: platforms.some((p) => p.state === "fatal") || problems.length > 0 ? "degraded" : "up";
```

The state file updates on gateway activity, not on a fixed heartbeat — so a healthy gateway that simply receives no messages for a quiet quarter-hour reports degraded. Monitoring that cries wolf on idle teaches operators to ignore it.

Suggested: make state-file age informational (report it, don't degrade on it), or key the threshold to an actual heartbeat cadence if the engine has one.

danielmiessler:
Fixed in source: a dead recorded pid with a supervised job now cross-checks the supervisor's live pid and reports `restarting` instead of `down`, and state-file age on a live pid is informational instead of a degrade (`LIFEOS/HERMES/Health.ts`). Monitoring that cries wolf on idle teaches operators to ignore it—agreed, and that line is basically in the fix comment now. Thanks.

PR #2028, opened 2026-08-30
fix(hermes): measure gateway liveness from the event-loop heartbeat, not the turn-boundary state file

kirmalyshev:
## The bug

`checkHermesHealth()` ages `gateway_state.json` and reports `degraded` past `STATE_STALE_MS` (15 min).

That file is not a heartbeat. `gateway/run.py`'s `_persist_active_agents()` is documented as running "at every turn boundary (a running-agent slot is claimed or released)". A gateway nobody has messaged legitimately never touches it. So **any idle install reports `degraded` after 15 minutes of perfect health, and stays that way** until someone sends a message.

## The fix

The event loop already writes `<HERMES_HOME>/state/gateway.heartbeat` — `shutdown_watchdog.py` describes it as the event-loop heartbeat, and it ticks without traffic. `Health.ts` referenced `gateway_state.json` four times and the heartbeat zero times.

This reads the heartbeat for liveness and keeps `gateway_state.json` for pid, platforms, and active agents, which is what the two files are actually for. The heartbeat is trusted **only when its `pid` matches the pid just probed**, so a leftover file from a dead gateway cannot vouch for a live one, and it falls back to the old field when no matching heartbeat exists.

It also guards a NaN case that predates this change: `Date.parse` returns `NaN` on an unrecognised stamp, and `NaN > STATE_STALE_MS` is **false**, so an unreadable `updated_at` silently disabled the staleness check and reported a healthy gateway. Unknown age is now its own problem — the same false-green this file's header says it exists to prevent.

## Reproduction

Real install, gateway alive, telegram connected, event loop last ticked 4 seconds before the probe:

```
before   🟡 gateway up · 1/1 channels · 1 issue
         ! gateway_state.json last written 84m ago
after    🟢 gateway up · 1/1 channels          problems: []
```

Corrupt heartbeat stamp, same install:

```
before   status: up        problems: []
after    status: degraded  problems: ["state/gateway.heartbeat has an unreadable updated_at (not-a-timestamp)"]
```

## Note

The 7.40.4 zero-token heartbeat (calendar/mail/queue pre-run ticks every ten minutes) masks this, because those ticks are turns that refresh `gateway_state.json`. Installs without them configured see it permanently.

### DIFF
--- a/LifeOS/install/LIFEOS/HERMES/Health.ts
+++ b/LifeOS/install/LIFEOS/HERMES/Health.ts
@@ -3,12 +3,13 @@
  * Health — one deterministic read of whether the Hermes sidecar is actually up.
  *
  * The sidecar is a separate process tree with its own supervisor, so "is it
- * running" has three independent answers that can disagree, and the disagreement
+ * running" has four independent answers that can disagree, and the disagreement
  * is the interesting part:
  *
  *   1. the supervisor (launchd/systemd) thinks it has a job
  *   2. `gateway_state.json` says a pid was running as of some timestamp
  *   3. that pid is, right now, alive
+ *   4. `state/gateway.heartbeat` says the event loop ticked recently
  *
  * A gateway in a crash loop satisfies (1) and (2) continuously while being down
  * essentially all the time — which is exactly how it went unnoticed on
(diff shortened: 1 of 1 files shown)


danielmiessler:
Superseded by the 08-17 change that stopped degrading on state-file age (#1861); as written this would promote that note back to degraded. The pid-matched heartbeat read is a good idea and I'll take it as the note's source when I'm next in that file, with credit. Closing, thanks.

PR #2029, opened 2026-08-30
fix(satisfaction): stop scoring numbered instructions and menu selections as ratings

kirmalyshev:
## The bug

`parseExplicitRating` treats a leading 1-10 as an explicit satisfaction rating. On one install, **8 of the 8 explicit ratings ever recorded were false** and none were genuine.

Two failure modes, neither covered by the #1670 / #1703 guards:

| input | parsed as |
|---|---|
| `1 commit and push` | rating **1/10**, comment `commit and push` |
| `one commit and push` | rating **1/10**, comment `commit and push` |
| `1` answering a menu the assistant just offered | rating **1/10** |

The last one is common: the assistant offers `1. …  2. …`, the user types `2`, and that becomes a 2/10.

## Why it matters more than a bad number

The damage compounds downstream. Every rating under 5 writes a `LEARNING_low-rating-*.md`, and every rating under 4 fires a full `FailureCapture` with an LLM-written failure title. So a successful session acquires an invented failure. A real example from the affected install:

```
commit-and-push-requested-but-conversation-shows-unresolved-symlink-conflicts
```

Nothing went wrong in that session. The "unresolved conflicts" were an installer correctly refusing to clobber symlinks it did not create. The learning corpus is now carrying a fabricated lesson about a non-event.

## The fix

Two guards:

1. A trailing comment is honoured only on the unambiguous `N/10` form. A bare number carries no evidence of intent, so anything trailing it is an instruction, not a rating comment.
2. A bare number is rejected when the previous response offered enumerated options, because that is a selection. Detected with three patterns: `1. ` / `1) ` / `(a) ` at line start, `option 2`, and inline `(a) … (b) …`.

`parseExplicitRating` takes `lastResponse` as an **optional second parameter defaulting to `''`**, so every existing caller is unaffected.

Real ratings still parse: `8`, `8/10`, `8/10 nice catch`, `eight`.

## Tests

I have a `bun test` file covering all of this — 14 cases across both failure modes plus the still-working forms. The repo ships no `*.test.ts` and no runner, so I left it out rather than bundle a testing convention into a bug fix. Say the word and I will add it here or in a follow-up.

### DIFF
--- a/LifeOS/install/hooks/SatisfactionCapture.hook.ts
+++ b/LifeOS/install/hooks/SatisfactionCapture.hook.ts
@@ -127,14 +127,39 @@ const WORD_NUMBERS: Record<string, number> = {
 
 // ── Explicit Rating Detection ──
 
-export function parseExplicitRating(prompt: string): { rating: number; comment?: string } | null {
+// Enumerated-option markers in the previous response. A bare "1" or "2" replying to a
+// response that offered choices is a SELECTION, not a rating.
+const OPTION_MENU_PATTERNS: readonly RegExp[] = [
+  /^\s*(?:\*\*)?\(?[1-9a-e][.)]\s/m,   // "1. ", "1) ", "(a) ", "**2. "
+  /\boption\s+[1-9a-e]\b/i,            // "option 2"
+  /\(\s*[a-e]\s*\)/,                   // inline "(a) ... (b) ..."
+];
+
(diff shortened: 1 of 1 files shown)


danielmiessler:
Took part (b) with credit: menu selections after a response that offered options are skipped. Part (a) I'm declining; it would drop legitimate "9 nice work" comments, and the imperative case it targets is already covered locally. Closing, next release. Thanks.

PR #2030, opened 2026-08-30
fix: create parent dir before writing state files

abseg:
## Problem

Nine tools and hooks write to paths under `LIFEOS/MEMORY/STATE` and `LIFEOS/OBSERVABILITY` whose parent directory git cannot carry, because git does not track empty directories. On a fresh clone — or any install where the directory has not happened to be created by some other code path first — the write throws `ENOENT` and the feature silently does nothing.

`SessionRename` is the clearest case: every rename lands nowhere because `MEMORY/STATE/` does not exist yet.

## Fix

Each write site calls `mkdirSync(dirname(path), { recursive: true })` immediately before the write. The call is a no-op when the directory already exists, so existing installs see no behaviour change.

## Files

`DoctrineReplay` · `GenerateKnowledgeSchemaDoc` · `MemoryRestore` · `ProposalGC` · `SessionProgress` · `SessionRename` · `UpdateModels` · `UsageAggregator` · `hooks/handlers/UpdateCounts`

## Verification

All nine parse-check clean (`bun build --target=bun`). Imports are merged into each file's existing `fs`/`path` import lines, matching that file's own quote and specifier style rather than adding a second import group.

+28 / -17 across 9 files. No behaviour change beyond the directory creation.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/DoctrineReplay.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/DoctrineReplay.ts
@@ -19,8 +19,8 @@
  *   bun DoctrineReplay.ts sample [--n 24] [--out <path>]   # emit manifest JSON
  *   bun DoctrineReplay.ts sample --json                    # manifest to stdout only
  */
-import { readdirSync, readFileSync, statSync, writeFileSync } from "node:fs";
-import { join } from "node:path";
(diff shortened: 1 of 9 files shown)


danielmiessler:
Ported eight of nine hunks with credit (GenerateKnowledgeSchemaDoc already had its mkdir). Closing, next release. Thanks.

PR #2031, opened 2026-08-30
fix: detect forked subagents from the hook payload, not env alone

abseg:
## Problem

`isSubagentContext()` tests environment variables only. A forked subagent does not reliably set any of them — including `CLAUDE_CODE_FORK_SUBAGENT` — and it shares the parent's `session_id` byte for byte. Measured live: a fork's `PostToolUse` hook process had every marker in the union unset.

So the test returns `false` inside a fork, and every consumer treats the fork as the main session:

- **PostToolObserver** injected main-session nudges and the SYSTEM line into a fork's own transcript.
- **ISASync** injected the ascent strip there too.
- **ConfigEvalFire** spawned the entire behavioural regression suite when a fork touched a sentinel file.

This is the residual half of #1831: the fork marker was added for exactly this, and on at least some runtimes it is never set.

## Fix

`PostToolUse` carries `agent_id` / `agent_type` for delegated calls and omits them in the main session, which makes the payload the reliable signal.

`isSubagentContext()` now takes the parsed hook input as an optional first argument and returns `true` when either field is present. The environment union is unchanged and still covers the non-fork delegate families and standalone runs, where no payload exists. Existing no-argument callers keep working — the parameter is optional, so this is not a breaking change.

The three consumers above pass their already-parsed `input` through. `ConfigEvalFire.readInput()` is widened to carry the two fields.

## Scope note

In `ISASync`, only the ascent strip is gated. `syncToWorkJson` is deliberately left ungated and never was gated: a fork editing an ISA is real work on the run, and its progress belongs in `work.json`. Only the strip is per-context, so only the strip is gated.

## Verification

All four files parse-check clean (`bun build --target=bun`). `input` is confirmed in scope at each insertion site, and no duplicate imports are introduced.

### DIFF
--- a/LifeOS/install/hooks/ConfigEvalFire.hook.ts
+++ b/LifeOS/install/hooks/ConfigEvalFire.hook.ts
@@ -1,6 +1,6 @@
 #!/usr/bin/env bun
 /**
- * @version 1.0.1
+ * @version 1.1.0
  * ConfigEvalFire — PostToolUse(Write|Edit) hook that fires the {{DA_NAME}} behavioural
  * regression suite when a behaviour-defining file changes.
  *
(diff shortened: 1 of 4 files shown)


danielmiessler:
Thanks for this. The flag half landed 08-17 under #1864, and the stdin signal (agent_id / agent_type / SessionStart source=fork) went in via #1958 with credit to you as well. I didn't take the composer-wide exit: LoopDetector and AlgorithmNudge deliberately process delegate events, and that gate would have silenced them. Closing.

Issue #1994, opened 2026-08-27
Windows: Inference.ts JSON envelope parse fails when MCP layer prints log lines to stdout — memory reviewer reports inference failed


spec-vova:
### Environment
LifeOS 7.40.4 fresh install · Windows 11 · bun 1.4.0 · Claude Code harness (claude.exe via WinGet)

### Symptom
MemoryHealthCheck: CRITICAL — "Latest reviewer run is parse-failed / inference failed: Process exited with code 129". Reproduced directly:

```
bun Inference.ts --level low "You answer with exactly: OK" "say OK"
# claude returns a valid envelope ({"result":"OK",...})
# then stdout also contains: Client.listTools() called but server does not advertise tools capability - returning empty list
# → Error: Failed to parse claude JSON envelope (exit 1)
```

### Root cause
Inference.ts does `JSON.parse` over the WHOLE stdout. On this setup an MCP client log line rides stdout alongside the envelope, so the parse of perfectly valid inference output fails, and the memory reviewer (and everything else built on Inference.ts) reports failure.

### Fix (verified locally — reviewer now completes 10/10 dispatches)
On parse failure, salvage per-line: split stdout by newlines, `JSON.parse` each trimmed line that looks like a JSON object, prefer the last parsable one. (Alternative: substring from first `{` to matching last `}`.)

---
Found during a fresh Windows 11 install (3 of 5 findings, see #1992, #1993). Happy to test patches.

danielmiessler:
Landed: Inference passes `--strict-mcp-config` so no MCP server loads at all, and the envelope parser now takes the last parseable JSON line (or the last `result` in an array). Reproduced with a fake claude that prints MCP log lines first. Next release. Thanks.

PR #2032, opened 2026-08-30
fix: repair four breaks in the autonomic memory loop

abseg:
## Summary

Four independent defects in the path between "memory ran" and "the health check can see that it ran". Each is small. Together they made the autonomic loop report itself broken while it was, in part, genuinely not running.

They are grouped in one PR because they were found as one investigation and two of them mask each other: the missing retrieval writer and the frozen review stamp both surface as the same "loop is stuck" reading. Happy to split if you'd rather review them separately.

## 1. `Inference.ts` — envelope shape

The CLI emits **either** a single result object **or** a JSON array of stream events whose last `type:"result"` element carries the envelope. Only the object shape was handled. An array fell through the object check (arrays are objects), `envelope.result` read `undefined`, and every programmatic caller died on `missing result field`.

Fix selects the terminal `result` element instead of assuming either shape.

## 2. `MemoryRetriever.ts` — the evidence file nothing wrote

`CortexHealth` and `MemoryStatus` have always **read** `memory-retrievals.jsonl`, and `MemoryHealthCheck` warns when no row is fresher than its window. Nothing has ever **written** it. A repo-wide grep found three references, all readers, and the path did not exist on disk. So the check was correct and the writer was simply absent: retrieval fired every turn and left no trace.

Fix adds the writer, honouring `CortexHealth.validRetrievalRow` exactly — `ts`, `query_hash`, `returned_count`, `duration_ms`, optional `top_score`. Extra keys make a row *invalid*, which is a louder failure than a missing one, so the schema is deliberately not extended. Only a hash of the query is stored, never the query text: this is freshness evidence, not a query history.

Rows are written on **every** call including cache hits, because the question the health check asks is "did retrieval run this turn?" and a cache hit is still a turn that ran it.

## 3. `MemoryReviewer.ts` — `last_review_at` only stamped on spawn

The field was written only by `MemoryReviewFire.hook.ts`, and only at the moment it decided to spawn. A review that ran by any other path — a manual run, a retry, a recovery after the hook was suppressed — left it frozen.

Two things were wrong, not one. The health report was misleading, **and** `min_minutes_between` is computed from that same field, so a completed review did not count against the rate cap and the hook could immediately spawn a duplicate.

The event the field names is "a review finished", so the reviewer is where it belongs. Skipped runs are not reviews and do not stamp.

## 4. `MemoryTurnStart.hook.ts` — emission order

The composed payload runs ~12 KB, past the harness inline budget, so the whole thing is persisted to a file and only a ~2 KB preview is seen. The two blocks whose entire contract is *render this verbatim* were emitted last, landing at byte ~9,776 of 12,226 and outside every preview.

They are short and bounded (one line each), so they are now emitted first and the unbounded hot layer follows. No content change, order only.

## Verification

All four parse-check clean (`bun build --target=bun`) against a tree with dependencies installed.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/Inference.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/Inference.ts
@@ -353,7 +353,19 @@ async function inferenceAttempt(options: InferenceOptions, modelOverride?: strin
         return;
       }
 
-      const envelope = parsedEnvelope as Record<string, unknown>;
+      // The CLI emits EITHER a single result object OR a JSON array of stream
+      // events whose last `type:"result"` element carries the envelope. Claude
+      // Code 2.1.246 returns the array form; before this, an array fell through
+      // the object check above (arrays are objects), `envelope.result` read
+      // undefined, and every programmatic caller died on "missing result field".
+      // That silently killed the memory reviewer from 2026-08-16 onward.
+      // Select the terminal result element rather than assuming either shape.
+      const envelope = (Array.isArray(parsedEnvelope)
+        ? (parsedEnvelope.filter(
+            (e): e is Record<string, unknown> =>
+              typeof e === 'object' && e !== null && (e as Record<string, unknown>).type === 'result',
+          ).at(-1) ?? parsedEnvelope.at(-1) ?? {})
+        : parsedEnvelope) as Record<string, unknown>;
       const stopReason = envelope.stop_reason;
       const resultText = typeof envelope.result === 'string' ? envelope.result : undefined;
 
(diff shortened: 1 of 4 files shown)


danielmiessler:
Split four ways. §3 (last_review_at reset on a successful run) landed with credit. §1 landed inside the same envelope-parser change as #1994. §2 is the writer that landed 08-17 under #1848. §4 I'm holding until the >12 KB truncation is shown; the format gate scans for tag presence, not order, so the reorder has no observable effect here. Closing, thanks.

Issue #1986, opened 2026-08-27
Cortex skill still lists four note types in its instructions, so adding a Book or Blog note fails validation


hjbrandt:
The knowledge schema knows six note types, but the Cortex skill's instructions still list four in most places — so asking it to add a Book or a Blog note fails its own first validation step.

**The mismatch, all in `LifeOS/install`:**

- `LIFEOS/TOOLS/KnowledgeSchema.ts` line 36 declares six types: person, company, idea, blog, research, book.
- `skills/Cortex/SKILL.md` line 14 correctly describes "six entity domains (People, Companies, Ideas, Research, Blogs, Books)".
- But the same file's add instructions say "Validate type is one of: People, Companies, Ideas, Research" (line 107), the routing table says "(People, Companies, or Ideas)" (line 37), the linking requirement says "any domain: People, Companies, Ideas, Research" (line 127), and the ingest classifier says "Determine entity type (People, Companies, Ideas, or Research)" (line 229). The frontmatter description has the same four-type list.

So `/knowledge add Books <title>` gets rejected at step 1 even though the schema, the harvester, and the folder layout all support books. Same for Blogs.

**Fix**: update the four stale lists to name all six types. Pure doc change, no code involved. I've made it locally and it's working — happy to send a PR. Found on current main (7.40.4).

garthsch:
Ahh, I was wondering about this one.Thanks for doing the investigation!

danielmiessler:
Fixed via your #2042. Next release. Thanks.

joecoxtech:
Heads up that this one is still open on main. #2042 was closed without merging (`mergedAt: null`), so the four stale lists in `skills/Cortex/SKILL.md` are still there today at lines 4, 107, 127 and 229.

There's also a code half that wasn't caught here, since the original report assumed it was doc-only. `KnowledgeGraph.ts:52`, `MemoryRetriever.ts:61` and `KnowledgeHarvester.ts:152` all hardcode `DOMAINS` to the same four types, while `KnowledgeDistill.ts:43` already has six. The practical effect is that `Blogs/` and `Books/` notes are invisible to graph traversal, stats, MOC generation and orphan detection — and via `MemoryRetriever`, to the memory block injected into every turn.

Both halves are in #2143.

PR #2042, opened 2026-09-01
Knowledge schema: enforce every declared select enum, not just type

hjbrandt:
## The bug

`validate()` in `KnowledgeSchema.ts` checks `CANONICAL_TYPES` by hand and never reads the `values` list declared on the other select fields. So `source_kind`, `status` and `quality_inferred` can hold any string and still lint clean.

I found it because a note in my archive carries `source_kind: external` — not in `SOURCE_KINDS` — and passed a 100%-conformant run.

## The fix

Walk `ENVELOPE` and check every select that declares `values`. Adding a new controlled vocabulary to the envelope is then enforced by construction, instead of needing a matching hand-written branch that nobody remembers to write.

Three regression tests added (`#7`). They fail on current `main` and pass here.

## Evidence

- Smoke suite: **32 passed, 0 failed**. Reverting only the new loop: **30 passed, 2 failed**, and only the two enforcement tests go red.
- `KnowledgeLint` over a 334-note archive: **333/334 → 334/334** envelope-conformant, catching exactly the one off-vocabulary note and nothing else. No false positives across 334 real notes.

## Two smaller things in the same area

- **`_schema.md` now says when `book` and `blog` apply**, next to the existing bookmark rule. `type` and `source_kind` both contain the value `blog` and nothing stated which axis a blog post belongs on. It is `type: idea` + `source_kind: blog`; `type: blog` is the archived post itself. Both lines are descriptive of how the types are already used, not a new policy.
- **The shipped Cortex skill listed four note types in five places.** There are six — `Books` and `Blogs` were missing.

## Scope

No migrations, no vocabulary changes, no behaviour change for conformant notes. A non-conformant note starts reporting the violation it always had.

Related: #1986.

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/GenerateKnowledgeSchemaDoc.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/GenerateKnowledgeSchemaDoc.ts
@@ -88,6 +88,10 @@ export function render(): string {
   L.push("");
   L.push("Bookmarks are NOT a type: an unprocessed saved URL is `status: inbox` + `source_kind: bookmark` on the type it will become; `ingest` promotes it.");
   L.push("");
+  L.push("A `book` note is ABOUT a book — the book itself is the entity you'd look up later; a single insight you took from a book is `type: idea` (or whatever it is), with the book credited in `source_author`/`source_name`.");
+  L.push("");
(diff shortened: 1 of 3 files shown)


danielmiessler:
Ported with credit: every declared enum is enforced, the Cortex SKILL.md lists the six types, and the two schema doc lines went in. Running the lint over my own archive found exactly one true positive, which I fixed rather than loosening. Closing, next release. Thanks.

PR #2045, opened 2026-09-01
fix(isa-utils): function replacer in appendDecisionRow so ISA dollar amounts don't backref-expand

alexfry-axeai:
## What

`appendDecisionRow` (in `LifeOS/install/hooks/lib/isa-utils.ts`) appends the auto-rewind Decisions row using a **string** replacement built from the existing Decisions body:

```ts
return content.replace(decisionsRe, `${match[1]}${match[2].trimEnd()}\n${row}\n${match[3]}`);
```

## The bug

That Decisions body routinely contains literal dollar amounts. `String.prototype.replace` interprets `$1`, `$2`, `$&`, etc. in a **replacement string** as regex backreferences — so an amount like `$14,000` is read as *capture group 1* followed by `4,000`. On any resume of a completed ISA, the rewrite expands those into the captured section text, **duplicating the `## Decisions` heading and shredding the section** (a `$14,000` becomes `<group-1>4,000`, with fragment lines and repeated headers).

## The fix

Use a **function replacer**, whose return value is inserted verbatim with no `$` interpretation:

```ts
return content.replace(decisionsRe, () => `${match[1]}${match[2].trimEnd()}\n${row}\n${match[3]}`);
```

One line of behavior change, plus a short explanatory comment. `appendDecisionRow` is exported so it can be unit-tested.

## Test

Adds `isa-utils.test.ts`, a regression test using synthetic `$1`/`$2`/`$3`/`$4`-prefixed amounts. It asserts every amount survives intact, exactly one `## Decisions` heading remains, and the rewind row lands under Decisions. Fails on the old string replacement, passes on the fix.

```
bun test LifeOS/install/hooks/lib/isa-utils.test.ts   # 1 pass
```

### DIFF
--- a/LifeOS/install/hooks/lib/isa-utils.test.ts
+++ b/LifeOS/install/hooks/lib/isa-utils.test.ts
@@ -1,0 +1,34 @@
+import { test, expect } from 'bun:test';
+import { appendDecisionRow } from './isa-utils';
+
+// Regression (2026-09-01 incident): the auto-rewind appended its Decisions row
+// with a STRING replacement built from the existing Decisions body. That body is
+// full of literal dollar amounts, and String.replace treats `$1`/`$2`/`$&` in a
+// replacement string as regex backreferences — so `$14,000` expanded to
+// `<capture-group-1>4,000`, duplicating the `## Decisions` header and shredding
+// the section. The fix uses a function replacer, whose return value is inserted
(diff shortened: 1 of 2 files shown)


danielmiessler:
Ported with credit, test included. Closing, next release. Thanks.

PR #2046, opened 2026-09-01
feat(atlas): Docker collector  - containers as graph assets

asdf8675309:
## What

Atlas has no visibility into Docker. A machine running containers has real assets — databases, services, published ports — that never appear in the asset graph, so `owns`/`blast`/`stale` queries are blind to them.

## Change

One new collector, `LIFEOS/ATLAS/collectors/Docker.ts`, plus its registration line in `Atlas.ts` and a collector-table row (with freshness stamp) in `AtlasSystem.md`.

- Enumerates ALL containers (`docker ps -aq`, running or not), then reads canonical JSON from one batched `docker inspect` — no `--format` templates, whose field names drift across Docker versions.
- Assets are keyed `docker:container:<name>` — name, not ID. A recreate (`compose up`, image upgrade) rotates the ID but keeps the name, and keying by ID would churn a fresh asset per recreate. Same reason `Launchd.ts` keys on Label.
- Attrs capture every cheap connection signal: image, state, running, ports (published and exposed-unpublished), networks, mount destinations, restart policy, compose project, created/started timestamps.
- One `RUNS_ON` edge per container to the local machine asset, the way `Launchd.ts` does it.

## Why it's safe

- Docker not installed, or installed with the daemon stopped, returns the DEGRADED result (`complete: false`, empty) — the same contract `Github.ts` and `Launchd.ts` already use. `atlas sync` cannot exit non-zero because a machine has no Docker, and an absent source can never sweep prior observations.
- A container removed between `ps` and `inspect` marks the run `complete: false` — no sweep on a racy view; prior graph state stays intact.
- `inspect` exiting 0 with unparseable output throws loudly (contract regression, not an absent source) — the same split `Github.ts` draws.
- No secret values and no host paths enter the graph: mounts record destinations only, never bind-mount sources.
- Installs without Docker are byte-identical in behavior except one PARTIAL line in sync output.

## Verification

Tested on macOS, Docker Engine 29.7.2, Bun 1.3.14. The fork has no test runner, so this table is the reviewable proof.

| case | result |
|---|---|
| live daemon, one compose-managed container | `✓ docker: 2 assets, 1 edges (complete, swept)`, exit 0 |
| container asset after sync | `docker:container:<name>` active, attrs carry image/state/ports (v4+v6 bindings)/networks/mounts/restart-policy/compose-project; active `RUNS_ON` edge to the machine asset |
| snapshot export | container present in the redacted Pulse snapshot |
| docker CLI absent from PATH | `{"complete":false,"assets":[],"edges":[]}` — no throw |
| daemon reachable, zero containers | `complete: true`, machine asset only (early-return code path; not live-tested — the test machine has a container) |
| strict tsc over all of `LIFEOS/ATLAS` | 0 errors |

## Scope

No change to `Store.ts`, the CLI surface, or any existing collector. The two edited files change by 3 code lines total (one import + one registration entry) plus one doc row. No new dependency — `Bun.spawn` and the `docker` CLI only.

### DIFF
--- a/LifeOS/install/LIFEOS/ATLAS/Atlas.ts
+++ b/LifeOS/install/LIFEOS/ATLAS/Atlas.ts
@@ -29,6 +29,7 @@ import { launchd } from "./collectors/Launchd";
 import { systemd } from "./collectors/Systemd";
 import { gear } from "./collectors/Gear";
 import { secrets } from "./collectors/Secrets";
+import { docker } from "./collectors/Docker";
 
 // The two service collectors are platform-exclusive: exactly one is ever registered,
 // so a macOS install's collector set is byte-for-byte what it was before systemd
@@ -38,7 +39,7 @@ import { secrets } from "./collectors/Secrets";
 const SERVICE_COLLECTOR = process.platform === "linux" ? systemd : launchd;
 
 const COLLECTORS: Record<string, Collector> = Object.fromEntries(
-  [cloudflare, github, projects, infraInventory, SERVICE_COLLECTOR, gear, secrets].map((c) => [c.name, c]),
+  [cloudflare, github, projects, infraInventory, SERVICE_COLLECTOR, gear, secrets, docker].map((c) => [c.name, c]),
 );
 
 const FULL_SYNC_INTERVAL_MS = 60 * 60 * 1000;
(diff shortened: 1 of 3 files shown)


danielmiessler:
Ported with credit, one adaptation: the Docker collector registers only when `docker` is on PATH, otherwise every 15-minute tick on a no-docker install would write a partial sync row forever. Rebased on the current collector list and docs. Tested with a mocked `docker ps`; I don't run Docker here so your verification table is the evidence. Closing, next release. Thanks.

PR #2053, opened 2026-09-03
Doctor: ElevenLabs probe reports the diagnosed cause and its remedy

hjbrandt:
## What was wrong

When the ElevenLabs TTS probe failed with `quota_exceeded`, Doctor printed the generic fix:

```
❌ Voice notifications (ElevenLabs) — broken
   TTS failed (401): {"detail":{"type":"invalid_request","code":"quota_exceeded","message":"This request exceeds your quota of 34292. You hav
   fix: set ELEVENLABS_VOICE_ID to an API-permitted (premade/cloned) voice in <configRoot>/.env
```

The key and voice id were fine. The monthly character quota was used up. The hint sent me to the wrong place, and the truncated JSON hid the number that explained it.

## What this changes

- A probe result may now carry its own `fixCmd`; `probeAll` prefers it over the capability's static one. Small, general, no behaviour change for probes that don't set it.
- The ElevenLabs probe gets a `quota_exceeded` branch that shows ElevenLabs' own message and points at the plan or the monthly reset.
- The two existing branches (`famous_voice_not_permitted`, `paid_plan_required`) set their own fix too, so the static voice-id hint never appears for a cause it doesn't fix.
- The error-body parser lives in `lib/ElevenLabsError.ts` so it is testable without running the CLI. It copes with the 200-char truncation the probe applies (cut-off JSON) and returns `null` for non-JSON bodies. Four contract tests beside it (`bun test LifeOS/install/LIFEOS/TOOLS/lib/ElevenLabsError.test.ts`). The install tree ships no tests today, so drop that file if you'd rather keep it that way.

## After

Real run on a quota-exhausted account:

```
❌ Voice notifications (ElevenLabs) — broken
   powers: spoken notifications via the Pulse voice server
   ElevenLabs quota exceeded — This request exceeds your quota of 34292. You have 0 credits remaining, while 1 credits are required for this request.
   fix: wait for the ElevenLabs monthly quota reset, or add credits / upgrade the plan at elevenlabs.io (key and voice id are fine)
```

Offline run (`bun Doctor.ts` without `--network`) is unchanged. `bun build --target=bun Doctor.ts` compiles.

Related: #1461 (famous-voice 401), #1496 (`paid_plan_required`).

https://claude.ai/code/session_01VDugi9RvNY4JeCfDXKUkMA

### DIFF
--- a/LifeOS/install/LIFEOS/TOOLS/Doctor.ts
+++ b/LifeOS/install/LIFEOS/TOOLS/Doctor.ts
@@ -41,6 +41,7 @@ import { existsSync, readFileSync, writeFileSync, mkdirSync, chmodSync, readdirS
 import { join, basename } from 'path';
 import { createHash, randomBytes } from 'crypto';
 import { homedir } from "node:os";
+import { elevenLabsMessage } from './lib/ElevenLabsError';
 
 const HOME = process.env.HOME ?? process.env.USERPROFILE ?? homedir();
 const CONFIG_ROOT = process.env.CLAUDE_CONFIG_DIR || join(HOME, '.claude');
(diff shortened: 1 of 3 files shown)


danielmiessler:
Ported with credit, adapted: per-probe fixCmd on the result, the quota branch with ElevenLabs' own message, regex inlined instead of a new lib file. Closing, next release. Thanks.

PR #2060, opened 2026-09-04
fix(install-engine): fail checkSurvivingPlaceholders on a zero-file scan

### LINKED ISSUE #2058: checkSurvivingPlaceholders returns passed:true on a zero-file scan, so a mis-rooted Setup 9(d) gate cannot fail
## Summary

`checkSurvivingPlaceholders()` in `skills/LifeOS/Tools/InstallEngine.ts` returns
`passed: true` when it scans **zero files**. A mis-rooted or nonexistent `rootDir` therefore
reports success, which is precisely the failure mode Setup.md step 9(d) exists to catch.

## The code

```ts
export function checkSurvivingPlaceholders(rootDir: string): {
  passed: boolean; files: Array<...>; total: number;
} {
  // ...walk, which begins: if (!existsSync(dir)) return;
  return { passed: total === 0, files, total };
}
```

`passed` is derived from `total === 0` alone. Nothing records how many files were examined,
so "clean tree" and "never looked at anything" are indistinguishable to every caller.

## Why it matters

Setup.md step 9(d) leans on this as a gate:

> then `checkSurvivingPlaceholders(<configRoot>)` and require `passed: true` [...] a skipped
> or mis-rooted pass is the difference between a system that knows the user's name and one
> that addresses them as `{{PRINCIPAL_NAME}}`

The document explicitly anticipates a mis-rooted pass, but the function's return shape cannot
express one. A bad root produces a green result and the installer proceeds.

## Reproduction

```ts
const r = checkSurvivingPlaceholders("/path/that/does/not/exist");
// => { passed: true, files: [], total: 0 }
```

Encountered for real: a mis-rooted call reported `passed: true` while `Doctor.ts`
simultaneously reported 68 unrendered placeholders across 26 sites in the same tree.

## Suggested fix

Count scanned files and require a non-empty scan.

```ts
let scanned = 0;
const processFile = (filePath: string): void => {
  if (!TEMPLATE_EXTENSIONS.has(fileExtension(filePath))) return;
  scanned++;
  // ...
};
// ...
return { passed: total === 0 && scanned > 0, files, total, scanned };
```

Returning `scanned` also gives callers something to log, which makes a mis-rooted pass
obvious in transcripts rather than invisible.

## Related

The #1874 / #1852 / #1993 cluster covers `substituteTree` corrupting the engine's own
`IDENTITY_PLACEHOLDERS` table. This is a distinct defect in the *verification* half: even
with a correct table, the gate cannot fail when it never ran.

jbmml:
Fixes #2058. Version: main @ 5e2f2e8 (also present in 7.40.4). File: `LifeOS/install/skills/LifeOS/Tools/InstallEngine.ts:518`, and its byte-identical sibling `LifeOS/Tools/InstallEngine.ts` (Setup.md step 9(d) runs the outer copy; `Doctor.ts` imports the deployed one; both are patched here).

**Repro against a clean tree** (script imports the shipped function and points it at four temp roots):

```
nonexistent                              -> passed=true  total=0 files=0   <- bug
empty dir                                -> passed=true  total=0 files=0   <- bug
positive control (placeholder present)   -> passed=false total=1 files=1
clean rendered dir                       -> passed=true  total=0 files=0
```

**After this change:**

```
nonexistent                              -> passed=false total=0 files=0
empty dir                                -> passed=false total=0 files=0
positive control (placeholder present)   -> passed=false total=1 files=1
clean rendered dir                       -> passed=true  total=0 files=0
```

**Fix:** count files actually read (`scanned`) and require `scanned > 0 && total === 0`. The return type gains `scanned` for diagnostics. `Doctor.ts` sums `total` and never reads `passed`, so its behavior is unchanged. Setup 9(d) passes the full config root, which always holds `CLAUDE.md` and `settings.json` (both template extensions), so a real install still passes.

https://claude.ai/code/session_01CvY8QkJKBWyXcZeL2Q2UXc

### DIFF
--- a/LifeOS/Tools/InstallEngine.ts
+++ b/LifeOS/Tools/InstallEngine.ts
@@ -491,14 +491,16 @@ const IDENTITY_PLACEHOLDERS = [
  * or abort.
  */
 export function checkSurvivingPlaceholders(rootDir: string): {
-  passed: boolean; files: Array<{ file: string; placeholder: string; count: number }>; total: number;
+  passed: boolean; files: Array<{ file: string; placeholder: string; count: number }>; total: number; scanned: number;
 } {
   const files: Array<{ file: string; placeholder: string; count: number }> = [];
   let total = 0;
+  let scanned = 0;
   const processFile = (filePath: string): void => {
     if (!TEMPLATE_EXTENSIONS.has(fileExtension(filePath))) return;
     let src: string;
     try { src = readFileSync(filePath, "utf-8"); } catch { return; }
+    scanned++;
     for (const placeholder of IDENTITY_PLACEHOLDERS) {
       const count = src.split(placeholder).length - 1;
       if (count > 0) { files.push({ file: filePath, placeholder, count }); total += count; }
@@ -515,7 +517,9 @@ export function checkSurvivingPlaceholders(rootDir: string): {
   };
   if (existsSync(rootDir) && lstatSync(rootDir).isFile()) processFile(rootDir);
   else walk(rootDir);
-  return { passed: total === 0, files, total };
+  // A scan that reached zero template files cannot vouch for anything: a mis-rooted or
+  // nonexistent rootDir must FAIL the Setup 9(d) gate, not pass it.
+  return { passed: scanned > 0 && total === 0, files, total, scanned };
 }
 
 /**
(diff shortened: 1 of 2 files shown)


danielmiessler:
Ported: `checkSurvivingPlaceholders` requires `scanned > 0`, credited to you and #2058, with tests. Next release carries it. Thanks, @jbmml.

PR #2084, opened 2026-09-10
fix(system-prompt): make every Constitutional No.2 self-check answer yes-compliant (#2071)

### LINKED ISSUE #2071: CONSTITUTIONAL No.2 self-check inverts on question 3: a fully verified claim answers 'no' and fails the gate
### Version

LifeOS 7.40.4 / System Prompt 3.7.3

### What is broken

The CONSTITUTIONAL No.2 self-check has a polarity bug on its third question. Questions 1 and 2 are phrased so that a compliant answer is "yes". Question 3 is phrased so that a compliant answer is "no". All three then run through one gate that reads `Any no → not done`. Evaluated literally, work that is fully verified answers no to question 3 and is therefore classified as not done, while work that still contains a "should work" answers yes to all three and passes. The gate is inverted for exactly the question that names the thing the section exists to forbid.

Question 2 has a milder version of the same problem. `Web-facing → Interceptor screenshot taken?` has no compliant answer for work that is not web-facing: the honest answer is N/A, and a strict reading of "any no" fails it.

Practical impact is small, because the two sentences immediately below the check are unambiguous ("Never claim completion without tool-based evidence", "'Should work' is forbidden") and a model resolves the list by intent rather than by boolean evaluation. It is worth fixing anyway on two grounds. The constitutional tier is the one place the prompt is asking for literal reading. And a self-check whose only literally-passing state is the failing one teaches the reader to skim self-checks.

### Where (file:line)

`LIFEOS/LIFEOS_SYSTEM_PROMPT.md:94`, and the same line in the vendored install payload at `skills/LifeOS/install/LIFEOS/LIFEOS_SYSTEM_PROMPT.md:92`, so a fresh install and an updated install both carry it.

### Repro on a clean tree

```shell
# Against a clean 7.40.4 tree
grep -n 'Any "should work"' LIFEOS/LIFEOS_SYSTEM_PROMPT.md
# 94:Self-check before any done-claim: 1. Tool evidence in hand for every claim? 2. Web-facing → Interceptor screenshot taken? 3. Any "should work" left anywhere? Any no → not done.

grep -n 'Any "should work"' skills/LifeOS/install/LIFEOS/LIFEOS_SYSTEM_PROMPT.md
# 92: (identical line)
```

Now evaluate the three questions against a piece of work that is genuinely, fully verified: tool evidence in hand, screenshot taken, no "should work" anywhere.

### Negative control

Stated without reference to any fix.

On unpatched 7.40.4, a fully verified done-claim answers the three questions **yes, yes, no**. The gate reads `Any no → not done`, so it returns **not done**.

The same gate on a claim that still has an unverified "should work" in it answers **yes, yes, yes**, and returns **done**.

The check therefore passes the failing case and fails the passing case on question 3. No fix is referenced above; the inversion is visible from the shipped line alone.

Third-question answers, unpatched:

| Work state | Q3 answer | Gate verdict | Correct verdict |
|---|---|---|---|
| Fully verified, no "should work" | no | not done | done |
| Unverified, "should work" remains | yes | done | not done |

### Suggested fix

Flip the two questions so all three pass on "yes", leaving the gate alone. One line:

```diff
-Self-check before any done-claim: 1. Tool evidence in hand for every claim? 2. Web-facing → Interceptor screenshot taken? 3. Any "should work" left anywhere? Any no → not done.
+Self-check before any done-claim: 1. Tool evidence in hand for every claim? 2. If web-facing, Interceptor screenshot taken? 3. Every "should work" eliminated? Any no → not done.
```

Applied and running locally on 7.40.4, both copies. It is a prose change with no code path, so "tested" means the line now reads consistently, not that a suite went green.

Bowlerjim:
Fixes #2071.

## The problem

Constitutional №2's self-check runs three questions through one gate, "Any no → not done." Question 3 ("Any 'should work' left anywhere?") is phrased so that a compliant answer is no, so fully verified work fails the literal reading and work still carrying a "should work" passes it. Question 2 has no compliant answer for work that is not web-facing.

## The change

One line in `LIFEOS/LIFEOS_SYSTEM_PROMPT.md`:

> Self-check before any done-claim: 1. Tool evidence in hand for every claim? 2. Web-facing output seen in an Interceptor screenshot, or nothing web-facing? 3. Every "should work" gone? Any no → not done.

Every compliant answer is now yes, question 2 admits not-web-facing, and the gate line and the two sentences below it are unchanged. The version field is left for the maintainers to roll with the release.


danielmiessler:
Ported into the system prompt (3.13.1), wording adjusted so all three questions read yes-compliant. Next release carries it. Thanks, @Bowlerjim.

PR #2097, opened 2026-09-11
Memory reviewer no longer overwrites edits made while it runs

hjbrandt:
The memory reviewer reads the two memory files, thinks for a few minutes, then writes back a full replacement. If anything else changed a file during those minutes, the reviewer's write erased that change without a trace. I hit this on my own install: I merged some entries by hand, and a review that had started earlier put them all back seconds later.

This makes the reviewer check before it writes. When it reads a file, it keeps a fingerprint of the entries. Just before writing, it compares that with the file as it is now. If they differ, it skips the write and leaves the newer version alone. The next review starts from the current file.

The fingerprint comes from the reviewer's own read, never from the model's output, so the model can't get around the check.

Smaller changes that come with it:

- A skipped write is logged, and the memory health check shows a warning for 24 hours. The reviewer runs in the background, so otherwise nobody would know.
- A skipped write counts as the check doing its job, not as a failed review.
- If another write holds the file lock, the reviewer waits up to 2 seconds instead of giving up at once.
- The single-entry add gets the same check.
- MemorySystem.md explains it.

Other code that replaces the memory files works exactly as before unless it asks for the check.

One thing I left out: when a write is skipped, that review's new facts are dropped and a later review has to find them again. Merging them into the newer file is a bigger change.


danielmiessler:
Ported: fingerprint on read, expected fingerprint on write (`ESTALE_BASE`), chained through the reviewer's dispatch, plus a refused-writes health check. Real data-loss bug; thank you for finding it, @hjbrandt. Next release carries it.

PR #2098, opened 2026-09-11
Tests for the memory reviewer overwrite fix

hjbrandt:
Tests for #2097, and built on top of it.

There are 13 cases. Each one runs in its own process with a temporary home folder, so the real memory files are never touched. One of them replays the original problem, and it fails on the code without the fix.

There are no other test files in the repo yet, so I kept these separate. Happy to move them or drop them.


danielmiessler:
Ported with #2097 as one unit: your stale-write tests are in `MemoryStaleWrite.test.ts`, adapted to the current writer API, plus one more for a file with an invalid entry. Next release carries it. Thanks, @hjbrandt.

Issue #1950, opened 2026-08-22
`OverlaySystem`'s system-tree list has drifted from what the installers place, so five payload entries are never updated


mj-deving:
## What LifeOS does here

Two tools own the two halves of the lifecycle, and each carries its own idea of which trees LifeOS
owns.

`skills/LifeOS/Tools/DeployCore.ts` places the runtime on a fresh install. It enumerates the
payload's `install/LIFEOS/` directory and copies **every** top-level entry except a small skip set:

```ts
const RUNTIME_SKIP = new Set(["USER", "MEMORY", "node_modules", ".git"]);
```

`skills/LifeOS/Tools/OverlaySystem.ts` refreshes system-owned files on an update. It does not
enumerate; it walks a hardcoded list:

```ts
const SYSTEM_TREES: Array<[string, string]> = [
  ["hooks", "hooks"], ["skills", "skills"], ["agents", "agents"],
  ["LIFEOS/TOOLS", "LIFEOS/TOOLS"], ["LIFEOS/DOCUMENTATION", "LIFEOS/DOCUMENTATION"],
  ["LIFEOS/ALGORITHM", "LIFEOS/ALGORITHM"], ["LIFEOS/RULES", "LIFEOS/RULES"],
  ["LIFEOS/PULSE", "LIFEOS/PULSE"],
];
```

Its comment states the intent plainly: "Anything not listed here is left alone — absence from this
list is the safety property."

## Current behavior

The v7.40.4 payload ships eight directories and four files at the top level of `install/LIFEOS/`.
Neither `USER` nor `MEMORY` is among them — `USER` ships as `install/USER` and is handled by
`ScaffoldUser.ts`, and `MEMORY` is not shipped at all — so `RUNTIME_SKIP` matches nothing here and
`DeployCore` copies every entry.

| Payload entry | Installed by `DeployCore` | Refreshed by `OverlaySystem` |
|---|---|---|
| `ALGORITHM`, `DOCUMENTATION`, `PULSE`, `RULES`, `TOOLS` | yes | yes |
| `LIFEOS_SYSTEM_PROMPT.md`, `VERSION` | yes | yes (explicit handling) |
| **`ATLAS`** (14 files) | **yes** | **no** |
| **`HERMES`** (12 files) | **yes** | **no** |
| **`USER_TEMPLATES`** (6 files) | **yes** | **no** |
| **`CORTEX_INDEX_POLICY.json`** | **yes** | **no** |
[...]

PR #2102, opened 2026-09-11
docs(Update): correct which update legs work offline

hifiguy:
## Problem

`Workflows/Update.md`'s **Network unreachable** branch tells the operator to
"stop after re-running the additive legs that work from the on-disk skill
(steps 4–6)", and names step 3's overlay as the one leg that needs a fetched
payload.

Steps 4 and 5 don't work offline either — and the same sentence explains why,
one clause later: on an installed system `install/` is **bootstrap-only**.

- Step 4, `InstallHooks.ts`, reads `install/hooks/hooks.json`
- Step 5, `ScaffoldUser.ts`, reads `install/USER/`

Neither path exists in a bootstrap-only payload, so both exit `ok: false`.
Step 6, `ActivateImports.ts`, is the only leg that actually runs — it edits
`CLAUDE.md` and needs no payload.

## Reproduction

Clean-room, against the shipped skill only — no install required. Extract the
v7.40.4 release, point `--skill-root` at the **installed** copy it ships
(`LifeOS/install/skills/LifeOS`, whose `install/` holds only the five bootstrap
files), and `--config-root` at a scratch dir containing just a `CLAUDE.md`
copied from `install/CLAUDE.template.md`:

```
InstallHooks.ts    → {"ok":false,"error":"payload hooks.json not found at
                      <skillRoot>/install/hooks/hooks.json"}
ScaffoldUser.ts    → {"ok":false,"error":"template USER not found at
                      <skillRoot>/install/USER"}
ActivateImports.ts → {"ok":true,"dryRun":true,"wouldActivate":[],"wouldSkip":[...]}
OverlaySystem.ts   → {"ok":false,"blockers":["payload at <skillRoot>/install is
                      bootstrap-only (no system trees: hooks, skills, agents,
                      LIFEOS/TOOLS, ...)"]}
```

Three of the four steps refuse; the doc predicts only one. Step 6 is the sole
survivor, and only because it reads `CLAUDE.md` rather than the payload.

## Why it's worth fixing

The offline branch currently sends the operator into two hard failures with no
warning that they're expected. `OverlaySystem`'s refusal is explicitly
pre-labelled "expected, not a failure" — the other two aren't, so they read as
a broken install. An operator debugging that is likely to go looking for damage
that isn't there, or to start hand-copying payload files to make the steps pass,
which is how stale trees get created.

## Fix

Documentation only, one line, no code change. The wording now:

- names step 6 as the only leg that works offline
- states that steps 3–5 all require the fetch
- quotes the two additional refusal messages so they read as expected behaviour

Applied to both copies of the file, which are byte-identical on `main`:

- `LifeOS/Workflows/Update.md`
- `LifeOS/install/skills/LifeOS/Workflows/Update.md`

## Notes

No existing issue or PR covers this. The nearest neighbours — #1950
(`OverlaySystem`'s system-tree list drift) and #1770 (`copyMissing` can't update
system-owned files) — are both closed and describe different mechanics.

---

GitHub: @HiFiGuy · Discord: SecDude2469


danielmiessler:
Ported verbatim into `Update.md` (only one copy of the file exists now). Next release carries it. Thanks, @hifiguy.

Issue #1975, opened 2026-08-25
Desktop-app sessions misdetected as headless, silencing all voice notifications


emarcelcom:
### Summary

On LifeOS 7.40.4, voice notifications are silently suppressed for every session started from the **Claude Code desktop app**. `getNotificationChannel()` classifies them as `headless`, so all voice-firing hooks skip `/notify`.

### Root cause

`hooks/lib/notification-channel.ts` uses terminal identity as a proxy for "has a speaker":

```ts
if (!env.TERM && !env.TERM_PROGRAM && !env.KITTY_WINDOW_ID && !env.SSH_TTY) {
  return 'headless';
}
```

The desktop app is a real desktop session with real speakers, but it sets none of those four variables. Confirmed in a desktop-app session: all four unset, `CLAUDECODE=1`.

The result is a `skipped` event with `reason: "remote_channel:headless"` for every notification, which reads like a deliberate suppression rather than a misdetection.

### Workaround

Setting `env.LIFEOS_NOTIFICATION_CHANNEL = "desktop"` in `settings.json` restores voice, since rule 1 (explicit value wins) short-circuits the sniff.

### Secondary issue the workaround exposes

Four spawners stamp the child channel with a fallback that only fires when the variable is unset:

```ts
env.LIFEOS_NOTIFICATION_CHANNEL = env.LIFEOS_NOTIFICATION_CHANNEL || 'headless'
```

- `LIFEOS/TOOLS/CarrierProbe.ts`
- `LIFEOS/TOOLS/Inference.ts`
- `LIFEOS/PULSE/lib.ts`
- `LIFEOS/PULSE/checks/github-work.ts`

Once the parent session carries `desktop`, these children **inherit** it instead of being silenced, which re-opens the 2026-08-14 scheduled-task voice leak for anyone using the workaround. The hard-set spawners (`imessage.ts`, `telegram.ts`, `siri.ts`) are unaffected.

### Suggested fix

1. Teach the detector about desktop-app sessions so the override is unnecessary.
2. Independently, make the four spawners never inherit `desktop`, while still letting explicit remote channels pass through:

```ts
env.LIFEOS_NOTIFICATION_CHANNEL =
  env.LIFEOS_NOTIFICATION_CHANNEL && env.LIFEOS_NOTIFICATION_CHANNEL !== 'desktop'
    ? env.LIFEOS_NOTIFICATION_CHANNEL
    : 'headless'
```

[...]

PR #2105, opened 2026-09-12
fix(hooks): classify Claude desktop-app sessions as the desktop voice channel

pai-scaffolde:
Reproduced on a fresh LifeOS 7.40.4 install (macOS, Claude Code in the Claude desktop app); the fix was applied to that install and run there before filing. Addresses the detector half left open on #1975 ("paste the CLAUDE* env from a desktop session and I'll wire it") — this PR is that answer.

## Observed

Every session started from the Claude desktop app (Code tab) produced no voice; `MEMORY/VOICE/voice-events.jsonl` filled with `event_type: "skipped"`, `reason: "remote_channel:headless"` for every notification.

Environment of a desktop-app session (measured in a tool-spawned child process; hook processes inherit the same env): `CLAUDECODE=1`, `CLAUDE_CODE_CHILD_SESSION=1`, `CLAUDE_CODE_ENTRYPOINT=claude-desktop`; `TERM`, `TERM_PROGRAM`, `KITTY_WINDOW_ID`, `SSH_TTY` all unset.

## Root cause

`hooks/lib/notification-channel.ts` `getNotificationChannel()` infers "has a speaker" from terminal identity (`TERM`, `TERM_PROGRAM`, `KITTY_WINDOW_ID`, `SSH_TTY`). The desktop app is an interactive session with a speaker that sets none of them, so it fell into the launchd/cron `headless` branch.

## Fix

Count `CLAUDE_CODE_ENTRYPOINT === 'claude-desktop'` as terminal identity. Rule 1 (explicit `LIFEOS_NOTIFICATION_CHANNEL` wins) is untouched and evaluated first, so the spawners that stamp children `headless` (Inference, CarrierProbe, PULSE lib, github-work; remote channels imessage/telegram/siri) are unaffected. `CLAUDE_CODE_CHILD_SESSION` is not used: it is set in the desktop main session too and is not a channel signal.

## How tested

`bun test hooks/lib/notification-channel.test.ts` — 4 pass; on the shipped file the first case fails (entrypoint set, no terminal vars → `headless`, expected `desktop`). Cases: desktop marker alone → `desktop`; marker + explicit `headless`/`imessage` → the explicit value; nothing set or `CLAUDE_CODE_ENTRYPOINT=cli` → `headless`; `TERM` alone → `desktop` (unchanged).

Repro on a clean tree: from a desktop-app session, `env -u TERM -u TERM_PROGRAM bun -e 'import {getNotificationChannel} from "./hooks/lib/notification-channel"; console.log(getNotificationChannel())'` in `~/.claude` → `headless` before, `desktop` after; the next Stop hook plays. Voice is back on the live desktop-app install with this change.

Note on convention: this adds a `bun:test` file next to the lib (the repo has no runner today); happy to fold it into a probe table if that is preferred.


danielmiessler:
Ported next to the existing bundle-id check; the explicit channel override still wins first. Next release carries it. Thanks, @pai-scaffolde.

PR #2106, opened 2026-09-12
fix(interceptor): Capture.sh no longer passes --out to `interceptor screenshot`

### LINKED ISSUE #2065: Capture.sh passes --out to `interceptor screenshot`, a flag only the maintainer's patched binary accepts — every capture fails with exit 9 on a stock install
### Version
LifeOS 7.40.4 / Interceptor skill (interceptor CLI 0.24.2, stock build from upstream tag v0.24.2)

### What is broken
`Tools/Capture.sh` is documented as the only sanctioned way to take an Interceptor screenshot, and every verification path in the skill routes through it. It builds its screenshot invocation as `SS_FLAGS=(--context "$CTX" --save --out "$OUT")`. Upstream `interceptor screenshot` does not accept `--out`; the flag belongs to `save` and `net`. On any binary built from the public source the CLI rejects it outright, so the wrapper's DOM path fails, its `--pixel` fallback fails for the same reason, and the bounded recovery loop exhausts and exits 9. The result is that browser verification, which the skill declares mandatory before showing web output to a user, cannot produce a single screenshot on a clean install.

SKILL.md already names the cause in its "Pinned binary" note: the maintainer's binary carries a cherry-picked `screenshot --save` patch that honours `--out`, and on a stock install `screenshot --save` writes to cwd. The wrapper was written against the patched contract, so the note documents the divergence without the code accounting for it.

### Where (file:line)
`skills/Interceptor/Tools/Capture.sh:189` — `SS_FLAGS=(--context "$CTX" --save --out "$OUT")`

Reader that disagrees: `skills/Interceptor/Tools/Capture.sh:253` `resolve_saved()`, which already lifts the daemon-chosen `filePath` into `$OUT` and whose comment states that the `--pixel` path "ignores `--out` and saves to a daemon-chosen temp path".

### Repro on a clean tree
```shell
# Prerequisites: interceptor CLI built from public source (>= 0.23.16),
# extension loaded in a dedicated Chrome test profile, context pinned.
bash skills/Interceptor/Tools/PreflightIsolation.sh
# => [PreflightIsolation] OK — interceptor 0.24.2, pinned context connected and not denied.

bash skills/Interceptor/Tools/Capture.sh "https://example.com"
# => Capture.sh: FAIL — capture did not succeed after recovery.
# =>   last error: error: unknown flag '--out' for 'screenshot'
# =>   (--out belongs to 'save' and 'net'; 'screenshot --save' writes the image to disk).
# => exit 9

# Confirm the CLI contract independently:
interceptor help screenshot | grep -c -- '--out'
# => 0
```

### Negative control
On unpatched 7.40.4 with a healthy daemon and a preflight that exits 0, `Capture.sh` returns exit 9 and writes no image, for every URL and with or without `--full`. The failure is not environmental: preflight passes immediately before, `interceptor contexts` lists the pinned context, and the same daemon answers `interceptor help screenshot`, whose output contains zero occurrences of `--out`. Removing the flag and letting the existing `resolve_saved()` reconcile `filePath` produced a 2580x1670 PNG with standard deviation 23.8 on the first attempt, so the wrapper's own recovery machinery was never the problem — only the flag was.

### Suggested fix
Drop the flag; the reconciliation path already exists and was written for exactly this shape.

```diff
-SS_FLAGS=(--context "$CTX" --save --out "$OUT")
+SS_FLAGS=(--context "$CTX" --save)
```

`--save` writes to a daemon-chosen path and reports it as `filePath`; `resolve_saved()` moves that file to `$OUT`. Both the DOM and `--pixel` paths then behave identically, which is what the comment at line 253 already assumes. Tested on a stock 0.24.2 build: exit 0, non-degenerate image, `$OUT` populated.

pai-scaffolde:
Reproduced on a fresh LifeOS 7.40.4 install (macOS, Claude Code in the Claude desktop app, stock interceptor 0.24.13); the fix was applied to that install and `Capture.sh` run end to end there before filing. Fixes #2065.

## Observed

```
$ bash skills/Interceptor/Tools/Capture.sh https://example.com
Capture.sh: FAIL — capture did not succeed after recovery.
  last error: error: unknown flag '--out' for 'screenshot' (--out belongs to 'save' and 'net';
  'screenshot --save' writes the image to disk). ...
exit 9
$ interceptor help screenshot | grep -c -- '--out'
0
```

Every capture — DOM path, `--pixel` fallback, and the recovery loop — failed the same way, so the skill's mandatory browser verification could not produce one screenshot.

## Root cause

`Capture.sh:189` builds `SS_FLAGS=(--context "$CTX" --save --out "$OUT")`. Upstream `interceptor screenshot` has no `--out`; the flag belongs to `save` and `net`. The wrapper was written against the maintainer's cherry-picked binary (SKILL.md "Pinned binary" note), which SKILL.md itself says does not describe public installs. On a stock binary `--save` writes into the CLI's cwd and reports the file as `filePath` — which `resolve_saved()` already lifts into `$OUT` for the `--pixel` path.

## Fix

Drop `--out` from `SS_FLAGS`; run both `dom_capture` and `pixel_capture` in a subshell `cd`'d to `$OUT`'s directory so `--save` lands there; the existing `resolve_saved()` moves the reported `filePath` onto `$OUT`. `--full` is still accepted by the CLI (it gets past flag parsing to the context lookup). SKILL.md's stock-install sentence now says this instead of "pass `--out` explicitly".

## How tested

- `bash -n Capture.sh` clean.
- Negative control on the stock CLI (no browser needed): `interceptor screenshot --context zz-bogus --save --out /tmp/x.png` → `error: unknown flag '--out' for 'screenshot'` (rejected before any daemon contact); `... --save --full` and `... --save` → reach the context lookup.
- End to end on the live install in the pinned test profile: `Capture.sh https://example.com` prints the saved path and exits 0; the `--save` result carries `filePath` as `resolve_saved()` expects.


danielmiessler:
Already fixed in source the same way (`--save` into a staging dir). Closing with thanks, @pai-scaffolde.

Issue #2052, opened 2026-09-02
VoiceServer reads voices from settings.json only, so the DA speaks with two voices and every notify logs "fallback defaults"


lgertel:
**Version:** LifeOS 7.40.4 (`ce046f2`)

## What

`LIFEOS_CONFIG.toml` is where the DA's voice is declared (`[da.voices.main]`, required by `LifeosConfig.ts`), and `hooks/lib/identity.ts` already reads identity through `loadLifeosConfig()`. The VoiceServer never does: it loads voices from `~/.claude/settings.json` `daidentity.voices`, a section the setup flow no longer writes. Result on a configured install:

- hook-driven lines (the "DA says" notifications) use the configured voice;
- every skill's `curl -X POST /notify` (hundreds per day) resolves no voice, logs `Voice settings: fallback defaults (no config found for main)` and speaks with the hard-coded default voice;
- the user hears two different voices for one assistant.

## Where

`LIFEOS/PULSE/VoiceServer/voice.ts`

- header comment lines 9–10: "look up in settings.json daidentity.voices … use settings.json daidentity.voices.main as default";
- `loadVoiceConfigFromSettings` (line 209): `join(homedir(), ".claude", "settings.json")` (line 210), `settings.daidentity || {}` (line 220), `daidentity.voices || {}` (line 221); no read of `LIFEOS_CONFIG.toml`;
- line 566: `log("warn", \`Voice settings: fallback defaults (no config found for ${voice})\`)`.

`LIFEOS/TOOLS/LifeosConfig.ts` lines 166–168 define and require `[da.voices.main].voice_id`; `hooks/lib/identity.ts` line 37 loads through it.

## Reproduce

1. Set `[da.voices.main] voice_id = "<any ElevenLabs id>"` in `LIFEOS_CONFIG.toml`; leave `settings.json` without a `daidentity` block (the current setup output).
2. Start Pulse; `curl -s -X POST localhost:31337/notify -H 'Content-Type: application/json' -d '{"message":"test"}'`.
3. Read the VoiceServer log: `fallback defaults (no config found for main)`; the audio plays in the default voice, not the configured one. Trigger a hook notification: the configured voice plays.

## Expected

One source of truth for voices: the VoiceServer resolves `main` and named voices from `LIFEOS_CONFIG.toml` through `LifeosConfig.ts`, the same way `hooks/lib/identity.ts` does, with `settings.json` at most a legacy fallback; no "fallback defaults" warning on a configured install.

[...]

PR #2107, opened 2026-09-12
fix(pulse): VoiceServer reads the DA voice from LIFEOS_CONFIG.toml

pai-scaffolde:
Reproduced on a fresh LifeOS 7.40.4 install (macOS, Claude Code in the Claude desktop app); the fix was applied to that install and run there before filing. Precedent: PR #1781 made `hooks/lib/identity.ts` read the config first; `voice.ts` was left on the mirror.

## Observed

After `/interview` wrote a custom `voice_id` into `USER/CONFIG/LIFEOS_CONFIG.toml [da.voices.main]`, Pulse kept speaking with the template default voice (Rachel, `21m00Tcm4TlvDq8ikWAM`). `settings.json` had no `daidentity.voices.main.voiceId` of its own; nothing in the shipped tree writes that block.

## Root cause

`LIFEOS/PULSE/VoiceServer/voice.ts` resolves the main voice from `settings.json daidentity.voices` only (`loadVoiceConfigFromSettings()` at startup and a live read on `/notify/personality`). Per `SystemUserBoundary.md` and `settings.enhancements.json` ("LIFEOS_CONFIG.toml is the canonical config source; settings.json carries a runtime mirror of daidentity for hooks") the toml is canonical and the Interview writes only the toml. No shipped tool projects the mirror, so the VoiceServer never sees the chosen voice.

## Fix

Read `[da.voices.main]` through `LIFEOS/TOOLS/LifeosConfig.ts` (Bun-native TOML, no new dependency) — the same loader identity.ts uses:

- `configMainVoice()` maps `LifeosVoiceSettings` to the module's `VoiceEntry` (same defaults the settings.json mapper applies) and returns `null` when no config loads (LifeosConfig throws on a missing file, empty `[da].name`, or missing `voice_id` — that catch is the fallback path).
- `loadVoiceConfig()` = settings.json mirror, then the config's main voice overriding `voices.main` / `voicesByVoiceId` / `defaultVoiceId` when present. `startVoice()` calls it.
- The `/notify/personality` live read tries the config first, then the mirror as before.

Why not have the installer write the mirror instead: that keeps two writers for one value and leaves this reader inconsistent with the ones that already went to the config (#1781); the doctrine is that system code reads identity through `loadLifeosConfig()`.

## How tested

Probe (`startVoice({enabled:true})` then `voiceHealth().default_voice_id`), temp HOME whose `settings.json` mirror carries Rachel, `LIFEOS_CONFIG_PATH` pointing at a temp toml with a distinct id:

| | toml present | toml absent |
|---|---|---|
| unpatched | `21m00Tcm4TlvDq8ikWAM` (mirror) | `21m00Tcm4TlvDq8ikWAM` |
| patched | `AAAAAAAAAAAAAAAAAAAA` (config) | `21m00Tcm4TlvDq8ikWAM` (mirror fallback) |

Transpiles (`bun build --target=bun --no-bundle`). On the live install, `curl localhost:31337/voice/health | jq .default_voice_id` returns the toml's voice after a Pulse restart.


danielmiessler:
Already fixed in source via #2052 by @lgertel (TOML canonical, settings mirror as fallback). Closing with thanks, @pai-scaffolde.

Issue #1946, opened 2026-08-22
healthsync points every unconfigured user at LIFEOS/USER/HEALTH/SHORTCUT_SETUP.md, which does not exist in the repo


DRAZY:
`HealthSync.ts status` prints one line per source, and each one ends with the same pointer:

```
[HealthSync] oura: unconfigured; fresh=n/a (not set up — see LIFEOS/USER/HEALTH/SHORTCUT_SETUP.md)
[HealthSync] eightsleep: unconfigured; fresh=n/a (not set up — see LIFEOS/USER/HEALTH/SHORTCUT_SETUP.md)
[HealthSync] apple: awaiting-first-export; fresh=n/a (not set up — see LIFEOS/USER/HEALTH/SHORTCUT_SETUP.md)
[HealthSync] function: unconfigured; fresh=n/a (not set up — see LIFEOS/USER/HEALTH/SHORTCUT_SETUP.md)
```

That file is not in the repository. A code search for `filename:SHORTCUT_SETUP.md` returns 0 results, while a search for the string returns 4 references in 2 files:

```
LifeOS/install/LIFEOS/TOOLS/HealthSync.ts:398
LifeOS/install/LIFEOS/TOOLS/healthsync/apple.ts:4, 229, 270
```

So the only setup pointer a new user is ever given is a dead path, and it is the pointer shown on the default state, since every source starts unconfigured.

Two of the references make it more than a missing convenience doc. `apple.ts:4` says "Path contract is duplicated in LIFEOS/USER/HEALTH/SHORTCUT_SETUP.md — keep in sync", which names it as a second source of truth for a contract that now exists in only one place. And `apple.ts:229` cites a specific section, "§ E", so the intended document had known structure.

The Apple path is the one that suffers most, because unlike oura, eightsleep and function it needs no credentials. Its working state is reachable purely by following instructions, and there are none. `healthsync/apple.ts` watches `<dataDir>/apple/_spool` for Health Auto Export v2 envelopes, but nothing in the repo tells a user that.

For the credentialed sources the state file does carry the actual requirement, which is more than the status line gives:

```
oura        OURA_CLIENT_ID / OURA_CLIENT_SECRET not set in ~/.claude/.env
eightsleep  EIGHTSLEEP_EMAIL / EIGHTSLEEP_PASSWORD not set in ~/.claude/.env
function    FUNCTION_HEALTH_EMAIL / FUNCTION_HEALTH_PASSWORD not set in ~/.claude/.env
[...]

PR #2108, opened 2026-09-12
fix(healthsync): one quiet skip line for unconfigured providers on pull

pai-scaffolde:
Reproduced on a fresh LifeOS 7.40.4 install (macOS, Claude Code in the Claude desktop app); the fix was applied to that install and run there before filing. Related: #1946 covered the `SHORTCUT_SETUP.md` pointer printed by `status`; this change does not touch `status` or that pointer.

## Observed

`MEMORY/OBSERVABILITY/health-sync.log`, every hour (`com.lifeos.healthsync`, BackgroundServices.md:56):

```
[HealthSync] oura: unconfigured; records=0; ms=2; error=OURA_CLIENT_ID / OURA_CLIENT_SECRET not set in ~/.claude/.env
[HealthSync] eightsleep: unconfigured; records=0; ms=2; error=EIGHTSLEEP_EMAIL / EIGHTSLEEP_PASSWORD not set in ~/.claude/.env
[HealthSync] function: unconfigured; records=0; ms=2; error=FUNCTION_HEALTH_EMAIL / FUNCTION_HEALTH_PASSWORD not set in ~/.claude/.env
```

Three error-shaped lines an hour for a state that is not an error: no provider was ever configured.

## Root cause

`runPull()` in `LIFEOS/TOOLS/HealthSync.ts` prints the same `error=` line for every source regardless of status. The provider modules return `status: "unconfigured"` with the missing-credential text in `lastError`, and the summary loop renders that as an error.

## Fix (in the tool)

Group `unconfigured` and `awaiting-first-export` (the same pair `runStatus()` treats as "not set up") into one `skipped (not configured): oura, eightsleep, apple, function` line. `ok` and `failed` sources keep the detailed line; the per-source reason still lands in `current.json` and the `healthsync.jsonl` row (unchanged). A configured provider that fails still logs its error.

## How tested

`HOME=<tmp> bun LIFEOS/TOOLS/HealthSync.ts pull` (store.ts derives `.env`, state, data and observability paths from HOME, so the run is hermetic):

```
unpatched: 4 lines — oura/eightsleep/function "unconfigured; ... error=<CREDS> not set", apple "awaiting-first-export; ... error=none"
patched:   1 line  — [HealthSync] skipped (not configured): oura, eightsleep, apple, function
```

On the live install the hourly job now logs the single skip line.


danielmiessler:
Ported: unconfigured providers collapse to one skipped line on pull. Next release carries it. Thanks, @pai-scaffolde.

PR #2109, opened 2026-09-12
fix(tools): MemoryReviewer --dry-run builds the prompts and stops

### LINKED ISSUE #2073: MemoryReviewer `--dry-run` calls the model and writes to the health ledger, while its own header promises "no inference"
`MemoryReviewer.ts` documents `--dry-run` as **"extract + prompt, no inference"** (line 28 of the file's own header). It calls the model anyway, and it also writes into the two artifacts CortexHealth reads to judge reviewer health. Measured on v7.40.4.

## The mechanism

`review()` threads `dryRun` into exactly one place — `dispatchItems()`:

```ts
const { summary, results } = dispatchItems(parsed.output.items, { dryRun: opts.dryRun });
```

Everything upstream of that runs unconditionally: `inference()` (twice, if the parse retry fires), `writeRunDebug()` (creates `MEMORY/OBSERVABILITY/reviewer-runs/<runId>/`), and `logRunSummary()` (appends a row to `MEMORY/OBSERVABILITY/reviewer-runs.jsonl`). The flag suppresses the memory writes and nothing else.

## Measurement

Counting run directories and ledger rows around one `--dry-run`, with a synthetic transcript and a counting double at the inference boundary:

| | before | after |
|---|---|---|
| `reviewer-runs/` directories | 254 | **255** |
| `reviewer-runs.jsonl` rows | 257 | **258** |
| calls to `inference()` | — | **1** |

## Why the ledger half is worse than the cost

`CortexHealth.reviewerEvidence()` treats a run directory newer than the latest ledger row as an orphaned run, and past `reviewerRunGraceMs` reports it as `timed-out` with a fabricated `consecutiveFailures` streak:

```ts
if (latestDir && Number.isFinite(latestDirMs) && (!Number.isFinite(latestRowMs) || latestDirMs > latestRowMs)) {
  ...
  if (nowMs - latestDirMs > thresholds.reviewerRunGraceMs) return { status: "timed-out", ... };
}
```

So the "safe" command can make the health check report a timeout that never happened, and every dry run inflates the denominator that `MemoryFixesCheck`, `MemoryInsights` and `MemoryStatus` read. A flag that promises to touch nothing, and is therefore reached for precisely when someone is being careful, is the worst place to put an unlogged side effect.

## Suggested shape

Split the two things the flag was doing:

- `--dry-run` returns right after the prompts are built — no inference, no run directory, no ledger row — and prints what it would have sent (transcript, exchange count, prompt sizes).
- a separate `--preview` keeps today's useful behaviour (call the model, show the parsed items, write no memories), with its artifacts in their own directory so they can never be mistaken for a run, and no ledger row.

## Sibling in the same release

`KnowledgeDistill.ts run --headless --dry-run` spawns `Inference.ts --level high --timeout 240000` **before** it checks the flag. There the spend is arguably inherent — the digest preview *is* the synthesis — so the fix is the usage text saying so, not a return. Worth a line either way, since today the cost is discovered by paying it.

Happy to send a PR for either or both.

pai-scaffolde:
Reproduced on a fresh LifeOS 7.40.4 install (macOS, Claude Code in the Claude desktop app); the fix was applied to that install and run there before filing. Fixes #2073 (the shape suggested there is the one taken).

## Observed

`bun LIFEOS/TOOLS/MemoryReviewer.ts review --dry-run` called the model, created `MEMORY/OBSERVABILITY/reviewer-runs/<runId>/` and appended a row to `reviewer-runs.jsonl`, while the file's own header (line 28) promises "extract + prompt, no inference".

## Root cause

`review()` threads `dryRun` into exactly one place, `dispatchItems()`. Everything upstream of that — `writeRunDebug()`, `inference()` (twice on a parse retry), `logRunSummary()` — runs unconditionally. Those two artifacts are what `CortexHealth.reviewerEvidence()` grades: a run directory newer than the latest ledger row is treated as an orphaned run and, past the grace window, reported `timed-out`; every dry run also inflates the reviewer denominators `MemoryFixesCheck`/`MemoryInsights`/`MemoryStatus` read.

## Fix (behavior, so the header becomes true)

`review()` returns right after the prompts are built when `dryRun` is set: no inference, no run directory, no ledger row. The result carries `dry_run: true` and `prompt_chars` (system/user sizes) alongside `transcript` and `exchanges`, so the operator sees what would have been sent. `dispatchItems()` keeps its own `dryRun` option (the smoke test uses it); `review()` no longer passes it since a dry run never reaches dispatch. The header line is unchanged because it now describes the code.

## How tested

`bun MemoryReviewer.ts test` (the tool's existing self-test), run under a temp HOME with the USER/MEMORY symlink layout, gains five checks: dry run returns `ok` with `dry_run`, `inference_duration_ms === 0` and no `dispatch_summary`, run-dir count unchanged, ledger row count unchanged, prompt sizes reported.

```
upstream file + the new checks (negative control):  dry-run: no run directory written — 1 → 2 ✗ ; no ledger row written — 1 → 2 ✗
patched:                                              all five dry-run checks ✓
```

The remaining self-test checks are unchanged before and after. Transpiles. On the live install: `review --dry-run` leaves the run-dir and ledger counts unchanged, and Cortex health reports ok.


danielmiessler:
Ported: `--dry-run` returns after prompt assembly with no inference, run dir or ledger row (credited with #2073). Next release carries it. Thanks, @pai-scaffolde.

Issue #1720, opened 2026-08-01
`LifeOsSchema.md` and the USER/ scaffold the installer ships describe two incompatible layouts


catchingknives:
`LifeOS/install/LIFEOS/DOCUMENTATION/LifeOs/LifeOsSchema.md` calls itself "The canonical shape of the USER directory in LifeOS" and says it "Applies to: `LIFEOS/USER/` in every LifeOS installation". The scaffold the installer actually copies contradicts it on naming, on directory placement, and on the frontmatter contract, and the runtime code sides with the installer while naming the schema as its spec.

Everything below is checkable from a clean checkout with `ls` and `grep`. Paths and line numbers are repo-relative, pinned to tag v7.1.1; section 5 also references v6.0.5, the last tag that shipped the page manifests. Paths starting `LIFEOS/USER/...` are quoted runtime values, as installed.

## 1. The two layouts

`LifeOS/Tools/ScaffoldUser.ts` line 48 is the source of truth for what a new user gets: `const templateUser = join(skillRoot, "install", "USER");`.

| | `LifeOsSchema.md` (Status: Draft v1.0, 2026-04-16, frontmatter `version: 1.0.4`) | `LifeOS/install/USER/` |
|---|---|---|
| Naming (§2) | "PascalCase, always. No underscores, no ALL_CAPS, no kebab-case", example `PrincipalIdentity.md`, "not `PRINCIPAL_IDENTITY.md`" | 23 of 24 root entries are ALL_CAPS; `PRINCIPAL/PRINCIPAL_IDENTITY.md` ships verbatim |
| Health (§4, §8) | `Health/` at USER root, a `domain` sibling of `Telos/` | `TELOS/HEALTH/` |
| Finances (§4) | `Finances/` at USER root | `TELOS/FINANCES/` |
| Business (§4) | `Business/` at USER root | `WORK/YOUR_COMPANIES/` |

Frontmatter contract (§3 requires `category`, `kind`, `publish`, `review_cadence`, `last_updated` on every `.md`):

```sh
cd LifeOS/install/USER
find . -name '*.md' | wc -l                             # 87
grep -rl '^category:'        --include='*.md' . | wc -l # 0
grep -rl '^publish:'         --include='*.md' . | wc -l # 0
grep -rl '^interview_phase:' --include='*.md' . | wc -l # 0
grep -rl '^kind:'            --include='*.md' . | wc -l # 7
[...]

PR #2110, opened 2026-09-12
fix(daemon): exclude USER/WORK/YOUR_COMPANIES from the aggregator

pai-scaffolde:
Reproduced on a fresh LifeOS 7.40.4 install (macOS, Claude Code in the Claude desktop app); the fix was applied to that install before filing. Related decision: #1720 (installer layout is canonical).

## Observed

`skills/Daemon/Tools/DaemonAggregator.ts` `EXCLUDED_PATHS` names `USER/BUSINESS` as the business-confidential zone; the installed tree has no `USER/BUSINESS` and business data lives at `USER/WORK/YOUR_COMPANIES` (present on the install), which the deny list does not cover.

## Root cause

The structural deny list was written for the pre-#1720 layout. `isExcluded()` is a prefix match on `EXCLUDED_PATHS`, so the real business directory is not excluded. Stated plainly: no current read path in the aggregator walks `USER/WORK/` (it reads TELOS, KNOWLEDGE/Ideas, PROJECTS.md, PRINCIPAL_IDENTITY.md and `MEMORY/WORK`), so this is not an observed leak; it is the guard every future read goes through being wrong about where the confidential data is.

## Fix

Add `join(USER_DIR, "WORK", "YOUR_COMPANIES")` to `EXCLUDED_PATHS`. `USER/BUSINESS` stays for installs that created one (CLAUDE.md still lists `BUSINESS/` as a conventional location).

## How tested

No test surface: `isExcluded` is module-private and nothing reads the directory today, so there is no observable before/after. Transpiles (`bun build --target=bun --no-bundle`).


danielmiessler:
Ported: `USER/WORK/YOUR_COMPANIES` is in the daemon's exclusion list. Next release carries it. Thanks, @pai-scaffolde.

Issue #1605, opened 2026-07-24
MemoryGraph.ts cannot run as shipped — graphology deps missing from install/package.json


cristbc:
## Summary

`LIFEOS/TOOLS/MemoryGraph.ts` imports three packages that `install/package.json` does not declare, so it fails at import on a stock install. The Pulse memory-graph endpoint explicitly instructs the user to run it, which means the graph layer is unreachable out of the box and fails with no signal anywhere in the UI.

## Evidence

`MemoryGraph.ts` imports:

```
graphology
graphology-communities-louvain
graphology-metrics/centrality/pagerank
```

`install/package.json` declares exactly one dependency: `yaml`.

Result on a stock install:

```
$ bun LIFEOS/TOOLS/MemoryGraph.ts build --all
error: Cannot find package 'graphology' from '.../LIFEOS/TOOLS/MemoryGraph.ts'
```

## Why it stays invisible

`/api/memory/graph` returns an empty graph plus a remediation string:

```json
{"nodes":[],"edges":[],"communities":[],"built":null,
 "note":"run: bun LIFEOS/TOOLS/MemoryGraph.ts build --all"}
```

The UI renders that as "not built yet," which is indistinguishable from "you haven't run it." Following the instruction produces a module-resolution error in a terminal, not in Pulse — so unless the user runs it by hand and reads stderr, the graph simply never exists. `KnowledgeGraph.ts` (the superseded tool) has the same import.

## Impact

This is the whole declared-edge graph layer: Louvain communities, PageRank god-nodes, betweenness bridges, orphan detection, `MEMORY/GRAPH/graph.json` and `PATTERNS.md`. On an install with a few hundred knowledge notes it is a substantial feature, and it is 100% local — no network, no model calls, sub-second.

For reference, after adding the three packages the first build on a real corpus took 0.1s and produced 497 nodes, 1306 edges, 23 communities, 2 orphans. The output is genuinely useful; it just cannot run as shipped.

## Proposed fix

Add to `install/package.json`:

```json
"graphology": "^0.26.0",
"graphology-communities-louvain": "^2.0.2",
"graphology-metrics": "^2.4.0"
```

Optionally, have the `/api/memory/graph` note distinguish "never built" from "build failed" by attempting a resolve and surfacing the error — the current wording sends the user to a command that cannot succeed.

## Related

Same class as #1600 (documented remediation that cannot work as shipped): the instruction is present, the machinery behind it is not.

[...]

PR #2111, opened 2026-09-12
fix(tools): MemoryGraph no longer crashes on an empty corpus

pai-scaffolde:
Reproduced on a fresh LifeOS 7.40.4 install (macOS, Claude Code in the Claude desktop app); the fix was applied to that install and run there before filing. (#1605 / PRs #1407, #1408 covered the graphology deps and duplicate slugs, not this.)

## Observed

`bun LIFEOS/TOOLS/MemoryGraph.ts patterns` on a fresh install (empty `MEMORY/KNOWLEDGE`, empty `MEMORY/WORK`):

```
error: graphology-metrics/centrality/pagerank: failed to converge.
      at build (LIFEOS/TOOLS/MemoryGraph.ts:552:13)
exit 1
```

## Root cause

`computePatterns()` calls `pagerank.assign(graph, ...)` unconditionally. graphology-metrics 2.4.2's power iteration checks `error < N * tolerance`; with `N = 0` that is `0 < 0`, never true, so it exits the loop unconverged and throws. Measured boundary (probe against the shipped node_modules): 0 nodes throws; 1 node, 2 nodes with or without an edge, 3-node path all converge. `louvain.assign` on the empty graph is fine. So the crash is exactly "no notes yet", i.e. every fresh install until the first knowledge note lands.

## Fix

Skip `pagerank.assign` when `graph.order === 0`. Every downstream reader of the `pagerank` attribute iterates the node list, which is empty, so `PATTERNS.md` and `graph.json` are emitted with 0 nodes instead of a crash. Graphs with nodes take exactly the old path.

## How tested

`LIFEOS_DIR=<tmp with empty MEMORY/KNOWLEDGE + MEMORY/WORK> bun MemoryGraph.ts patterns`: exit 1 with the trace above before; exit 0 after, `MEMORY/GRAPH/PATTERNS.md` + `graph.json` written. Also exit 0 for `patterns --all` on the empty corpus and for a one-note corpus. Verified on the live install as well.


danielmiessler:
Ported: pagerank is skipped on an empty graph. Next release carries it. Thanks, @pai-scaffolde.

PR #2112, opened 2026-09-12
fix(pulse): Arbol scanners are configured by their worker keys, not by config.yaml existing

pai-scaffolde:
Reproduced on a LifeOS 7.40.4 install (macOS, Claude Code in the Claude desktop app) that has Arbol configured for other uses; the fix was applied to that install and run there before filing.

## Observed

With `~/.config/arbol/config.yaml` carrying only `subdomain` and `auth_token`:

```
GET /api/bunker/arbol      → 502 {"error":"arbol worker unreachable: scanner returned HTTP 404"}
GET /api/bunker/sitehealth → 502 {"error":"site-health worker HTTP 404"}
GET /api/bunker/critical   → {"configured":true,"reachable":false,"reason":"scanner returned HTTP 404"}
```

and `/api/bunker` folds `security_source: { state: "unreachable" }` into every app card — a red "scanner unreachable" face on an install that has no scanner.

## Root cause

`LIFEOS/PULSE/modules/bunker.ts` `arbolConfig()` treats the mere existence of Arbol's general config file as "a security scanner and a site-health worker exist": when `security_worker` / `site_health_worker` are absent it substitutes default names (`infra-security`, `site-health`) and `fetchArbol()` / `siteHealthReport()` go fetch `https://<default>.<subdomain>.workers.dev/...`. The 404 that comes back is then classified `unreachable`. The module's own `ArbolFetch` type already has the state for this — `{ state: "not-configured" }` — but only a missing file could reach it.

## Fix

A worker exists only when its key is set. `arbolConfig()` returns `securityWorker` / `siteHealthWorker` as optional with no defaults; `fetchArbol()` returns `not-configured` when `security_worker` is unset (so `/critical` reports `configured:false`, `/arbol` 503 "security scanner not configured", and the app cards get `security_source: not-configured`, which renders silent by design); `siteHealthReport()` returns 503 "site-health worker not configured" when `site_health_worker` is unset. Installs with the keys set take exactly the old path. `monitorWorkersRe()` already filtered undefined names, so the cost card is unaffected.

Behavior change to disclose: an install whose workers really are named `infra-security` / `site-health` and that never set the two keys (the original comment implies such installs exist) now reads `not-configured` until it adds `security_worker:` and `site_health_worker:` to config.yaml. That is deliberate — the dropped defaults were the mechanism of the bug: a default name is a claim that a worker exists, and on any Arbol install without one it is false.

## How tested

Probe through `handleRequest` with `HOME` pointed at a temp dir whose `.config/arbol/config.yaml` holds `subdomain` + `auth_token` only (the probe's fetches went to a non-existent `*.workers.dev` host and got Cloudflare's 404, which is the real-world failure):

| | unpatched | patched |
|---|---|---|
| `/arbol` | 502 unreachable (HTTP 404) | 503 not configured |
| `/sitehealth` | 502 HTTP 404 | 503 not configured |
| `/critical` | configured:true, reachable:false | configured:false |

With both keys added to the same file the patched module behaves exactly as before (attempts the fetch → unreachable on 404). Transpiles. On the live install the bunker tab's security source now reads not-configured instead of unreachable.


danielmiessler:
Ported: the Arbol workers are optional with no default names and a `not-configured` state. Next release carries it. Thanks, @pai-scaffolde.

PR #2113, opened 2026-09-12
fix(pulse): read Bunker monitor state from the USER tree, not ~/.bunker

pai-scaffolde:
Reproduced on a LifeOS 7.40.4 install (macOS, Claude Code in the Claude desktop app) running a local Bunker reference implementation that writes its state under the USER tree; the fix was applied to that install and run there before filing.

## Observed

The Pulse overview (`/api/telos/overview`: live projects, status metrics, team) never showed any Bunker-monitored app, while the monitor state file existed at `<USER data home>/PULSE/Bunker/monitor-state.json` next to the Bunker shots the bunker module reads.

## Root cause

`LIFEOS/PULSE/Observability/observability.ts:4018` hardcodes `BUNKER_STATE_FILE = join(HOME, ".bunker", "monitor-state.json")` — a home-relative dot-directory outside both the config root and the USER data home. `BunkerSystem.md` (release note) says per-app Bunker state lives in the user tree; `SystemUserBoundary.md` places all per-install state under the USER data home; `PULSE/modules/bunker.ts` already keeps its Bunker data in `USER/PULSE/Bunker/`. The path predates the boundary and nothing else in the tree references `~/.bunker`.

## Fix

Resolve the file as `join(USER_DIR, "PULSE", "Bunker", "monitor-state.json")` — `USER_DIR` is the constant the rest of observability.ts already uses for USER-tree reads (it resolves through the `LIFEOS/USER` symlink to the data home), and `PULSE/Bunker/` is the directory bunker.ts uses. `readBunkerState()` and its three consumers are unchanged. An existing `~/.bunker/monitor-state.json` moves to `USER/PULSE/Bunker/`.

## How tested

Probe: `startObservability({enabled:true})`, `GET /api/telos/overview` through `handleObservabilityRequest`, `HOME` pointed at a temp dir holding `.claude/LIFEOS/USER/PULSE/Bunker/monitor-state.json` with one green app `demo-app`:

```
unpatched: 200, response never mentions demo-app  (reads ~/.bunker/monitor-state.json, absent)
patched:   200, response includes demo-app
```

Transpiles. On the live install the overview now lists the monitored apps after a Pulse restart.


danielmiessler:
Ported: monitor state is read from `USER/PULSE/Bunker/`. Next release carries it. Thanks, @pai-scaffolde.

PR #2119, opened 2026-09-14
fix(Doctor): accept native binaries in the hook-interpreter probe

Rubix-BNL:
## Problem

`checkHookInterpreters` in `LIFEOS/TOOLS/Doctor.ts` requires a `#!` shebang from every hook command given as a path. A compiled executable registered as a hook has no shebang, so `Doctor --hooks` flags it as broken:

```
cc-status: no #! shebang
```

Concrete case: iTerm2's `~/.config/iterm2/cc-status` (a Mach-O binary) wired as a status-line/hook command. `exec()` runs it without issue; only scripts need an interpreter line.

## Fix

Read the first four bytes of the resolved file and skip the shebang check when they are an ELF, Mach-O (32/64, either byte order) or fat/universal header. Everything else keeps the existing path (mode check, shebang check, interpreter lookup).

## Verification

- `bun LIFEOS/TOOLS/Doctor.ts --hooks` on an install with the iTerm2 binary hook: `hook interpreters: every registered hook resolves ✅` (was `needs attention` before).
- Script hooks unchanged: a hook without shebang still reports `no #! shebang`.

https://claude.ai/code/session_017w3q9FbuWxcW9fmh8G39NK


danielmiessler:
Ported: the hook-interpreter probe accepts ELF, Mach-O and fat binaries before the shebang test. Next release carries it. Thanks, @Rubix-BNL.

PR #2120, opened 2026-09-14
fix(tools): use --safe-mode in headless claude spawns to keep auth

pai-scaffolde:

## Observed

Every headless Claude call — `KnowledgeDistill`, the `MemoryReviewer` inference fallbacks, `CarrierProbe` — returned the same thing and exited 1 under launchd, so the scheduled jobs produced nothing and the only visible trace was a one-line error in the service log:

```
Not logged in · Please run /login
```

The interactive session was logged in the whole time. Reproduced by hand with `launchctl kickstart` on the distill job.

## Root cause

`LIFEOS/TOOLS/Inference.ts:264` and `LIFEOS/TOOLS/CarrierProbe.ts:117` pass `--setting-sources ''` to the nested `claude` process to keep hooks, skills and plugins out of the child. On claude-code 2.1.270 an empty source list no longer means only "load no settings files" — it also drops the user-level login, so the child starts unauthenticated. Bisected A–G over the flag set with `env -i` and the real binary; the empty `--setting-sources` is the single flag that reproduces it.

`--safe-mode` is the flag that does what the call sites actually want. Its help text on 2.1.270:

```
--safe-mode   Start with all customizations (CLAUDE.md, skills, plugins, hooks, MCP
              servers, custom commands and agents, output styles, workflows, custom
              themes, keybindings, and more) disabled — useful for troubleshooting a
              broken configuration. Admin-managed (policy) settings still apply. Auth,
              model selection, built-in tools, and permissions work normally. Sets
              CLAUDE_CODE_SAFE_MODE=1.
```

Context: closed PR #724 touched these same two flag values (proposing `'""'` for `--tools` and `--setting-sources`) for a different symptom — argument parsing and subprocess hangs. It was not merged, and the empty `--setting-sources` predates it.

## Fix

Replace `--setting-sources ''` with `--safe-mode` in both headless spawners, with a comment recording that an empty source list is not an equivalent. Nothing else changes: the same customizations stay off in the child, and the billing scrub (`ANTHROPIC_API_KEY` / `ANTHROPIC_AUTH_TOKEN` / `ANTHROPIC_BASE_URL`) above each call site is untouched.

The second commit applies the same one-flag change to the three other headless spawners in the tree — `LIFEOS/PULSE/lib.ts` (`spawnClaude()`, every Pulse job), `LIFEOS/PULSE/checks/github-work.ts` (the issue worker) and `skills/LocalIntelligence/Tools/ClaudeFill.ts` — and updates the documented flag pattern in `DOCUMENTATION/Pulse/PulseSystem.md`. On the install after the change: `spawnClaude("Reply with the single word OK.", {model:"haiku"})` → `OK` in 1.5 s, Pulse restarted on the new code, and ClaudeFill's exact argv (`--safe-mode --tools WebSearch,WebFetch --allowedTools WebSearch,WebFetch`) → `OK`. All four transpile.

## How tested

Direct probe against the installed binary (2.1.270), identical args apart from the one flag:

```
$ printf 'Reply with the single word OK.' | claude --print --model haiku --tools '' \
    --output-format json --setting-sources '' --system-prompt 'Follow the instruction literally.'
{"is_error":true,"subtype":"success","terminal_reason":"api_error","result":"Not logged in · Please run /login"}
exit 1

$ printf 'Reply with the single word OK.' | claude --print --model haiku --tools '' \
    --output-format json --safe-mode --system-prompt 'Follow the instruction literally.'
{"is_error":false,"subtype":"success","terminal_reason":"completed","result":"OK"}
exit 0
```

On the live install after the patch: `CarrierProbe` HOLDS (the fable dispatch executed `claude-fable-5-1`), and the distill job exits 0 both by hand and under launchd, with the digest written. Both files transpile (`bun build --target=bun --no-bundle`).

Repro on a clean tree: run the two commands above with `claude --version` ≥ 2.1.270; the first returns "Not logged in" on an otherwise logged-in machine, the second returns `OK`.


danielmiessler:
Ported at all five spawn sites. The auth drop did not reproduce on the current CLI, but `--safe-mode` measurably suppresses the customizations that `--setting-sources ''` no longer does, so it is the right flag regardless. Next release carries it. Thanks, @pai-scaffolde.

PR #2124, opened 2026-09-14
fix(integrity): stop the reference gate blocking on an installed tree

pai-scaffolde:

## Observed

`/ic` on that install read **136 BLOCKING** findings. Nearly all of them were the checker describing the maintainer's tree rather than the installed one. Root-caused to six causes, five of them checker assumptions: mirror / scratch / vendor-symlink scanning (84), runtime and principal-created targets required (27), release-stripped tests counted missing (20), and the model alias rejected for its `[1m]` suffix (2). The sixth was real documentation rot, fixed separately.

## Root cause

`ReferenceCheck.ts` assumes it is running on the maintainer tree.

- It walks `skills/LifeOS/install` — the shipped release payload as installed, the twin of `LIFEOS_RELEASES` on a maintainer tree. Its references describe the maintainer's layout, so every one of them resolves against the wrong root.
- It walks background-job scratch under `jobs/`, which holds throwaway copies of docs.
- It descends into skill directories that are symlinks out of `~/.claude`. Vendor-installed skills land as root-owned symlinks into `/Library/Application Support`; their relative links point into the vendor's own repo — the same class as plugin caches, and not ours to fix.
- `RUNTIME_TARGET_SUBSTRINGS` covered three `MEMORY/` subtrees only. On a fresh install every `MEMORY/` subtree, every `state/` dir and every JSONL stream is empty until something writes it.
- Files the principal materializes at setup time (the `USER/` tree, `settings.local.json`, `.mcp.json`, the checkpoint allowlist) were required to exist. A doc pointing at one is a setup instruction, not rot.
- Files the release strips (tests and the test harness, private zones, `settings.system.json`, the ARBOL tree) were required to exist on a tree that by construction does not have them.
- A doc line that itself says the target is gone — "X is absent on this install", "never shipped", "maintainer tree" — was still counted as a broken pointer.

Separately, `checkModelRungPin()` in `IntegrityCheck.ts` compared `settings.model` to the rung alias literally. A trailing `[<n>m]` / `[<n>k]` selects the context window, not the model, so `fable[1m]` read both as a rung mismatch and as a pinned model ID.

## Fix

`ReferenceCheck.ts` adds `skills/LifeOS/install` and `jobs` to the excluded prefixes, skips skill directories that are symlinks resolving outside the real `~/.claude`, and skips a reference whose own line carries an absence marker. Targets that cannot resolve by construction become a new finding type, `absent`, with a reason: runtime-written, principal-created, or release-stripped. `absent` is reported and never blocks. Release-stripped only applies on an installed tree, detected by `skills/_LIFEOS` being missing — the same tell `IntegrityCheck`'s `master_doc_paths` already uses. On the maintainer tree those targets still count as `missing`.

`IntegrityCheck.ts` counts `absent` in the reference note rather than pushing one finding each, and strips the context-window suffix before comparing the pinned alias. `ProseInCode.ts` skips the same mirrored payload via its `install` dir name.

## How tested

Probe tree with three references: a release-stripped test path, a documented absence, and real rot.

```
$ cat $HOME/.claude/LIFEOS/DOCUMENTATION/Probe.md
# Probe

a) Release-stripped: the suite lives at [foo test](test/foo.test.ts).

b) Documented absence: [Ghost](LIFEOS/DOCUMENTATION/Ghost.md) is absent on this install.

c) Real rot: [NoSuchDoc](LIFEOS/DOCUMENTATION/NoSuchDoc.md) explains the contract.
```

`ReferenceCheck.ts` resolves its root as `$HOME/.claude`, so the probe runs under a temp `HOME`. Before, all three are blocking:

```
$ HOME=<tmp> bun <base>/ReferenceCheck.ts --json
summary: {'missing': 3, 'stale': 0, 'orphan': 0}
  missing 3 test/foo.test.ts
  missing 5 LIFEOS/DOCUMENTATION/Ghost.md
  missing 7 LIFEOS/DOCUMENTATION/NoSuchDoc.md
```

After, only the real rot blocks:

```
$ HOME=<tmp> bun LIFEOS/TOOLS/ReferenceCheck.ts --json
summary: {'missing': 1, 'stale': 0, 'orphan': 0, 'absent': 1}
  absent  3 test/foo.test.ts   (release-stripped on an installed tree)
  missing 7 LIFEOS/DOCUMENTATION/NoSuchDoc.md
```

The documented absence on line 5 is not extracted at all (`scannedRefs` 2, not 3). With `skills/_LIFEOS` created in the same tree — the maintainer-tree tell — the test path goes back to blocking, confirming the narrowing is installed-tree only:

```
summary: {'missing': 2, 'stale': 0, 'orphan': 0, 'absent': 0}
  missing 3 test/foo.test.ts
  missing 7 LIFEOS/DOCUMENTATION/NoSuchDoc.md
```

Alias stripping, against `IntegrityCheck.ts:1656`:

```
"fable[1m]"            -> "fable"
"fable"                -> "fable"
"opus[200k]"           -> "opus"
```

All three files transpile (`bun build --target=bun --no-bundle`). On the live install the `/ic` verdict is CLEAN, with the absent-by-construction items listed as non-blocking info.

Repro on a clean tree: point `HOME` at a temp dir holding `.claude/LIFEOS/DOCUMENTATION/Probe.md` with the three references above, run `bun LIFEOS/TOOLS/ReferenceCheck.ts --json` from `upstream/main` and from this branch, and compare the summaries.


danielmiessler:
Ported four of the six hunks (jobs exclusion, foreign-skill symlink skip, `absent` classification, runtime targets, the last one narrowed so a missing code import is never exempted by directory); the payload exclusion and the model-alias strip were already in source. Next release carries it. Thanks, @pai-scaffolde.

PR #2125, opened 2026-09-14
fix(distill): log the model bytes when synthesis JSON is unparseable

pai-scaffolde:

## Observed

Distill's headless synthesis failed to parse twice in a row and then passed with identical code. The entire record of both failures:

```
[distill] synthesis JSON unparseable: SyntaxError: Unexpected end of JSON input
```

That line cannot distinguish a truncated answer, a fence-wrapped answer, or an empty one, so there was nothing to act on and nothing to reproduce from.

## Root cause

Not a bug in the parse — a gap in the diagnostics. `runHeadless()` has two failure branches over the same `out` buffer. The no-JSON branch at `LIFEOS/TOOLS/KnowledgeDistill.ts:258` already prints the bytes:

```
console.error(`[distill] synthesis failed — no JSON in Inference output:\n${out.slice(0, 500)}\n${proc.stderr.toString().slice(0, 500)}`);
```

The unparseable branch four lines below it printed only the exception. A model answer long enough to be truncated mid-object still matches `/\{[\s\S]*\}/`, so it lands in the quiet branch — exactly the case where the bytes matter most.

## Fix

Print the same evidence in the unparseable branch: head 500 and tail 300 of the Inference output plus the first 500 bytes of stderr. The tail is what distinguishes a truncation from a trailing fence. Diagnostics only — no control flow, no parsing, no behaviour change on a successful run; `out` and `proc` are already in scope at that line.

## How tested

Transpiles: `bun build --target=bun --no-bundle LIFEOS/TOOLS/KnowledgeDistill.ts` exits 0.

The log shape, against a fake 949-byte fenced-and-truncated answer (middle of head and tail elided here for width):

````
[distill] synthesis JSON unparseable: SyntaxError: Unexpected end of JSON input
```json
{"items":[{"lane":"content","title":"x"},{"lane":"content","title":"x"}, … {"lane":"content"
…
ntent","title":"x"}, … {"lane":"content","title":"x"},

inference: stream closed after 240000ms
````

Both tells are visible at a glance: the opening fence in the head, the severed object at the tail.

Repro on a clean tree: run `bun LIFEOS/TOOLS/KnowledgeDistill.ts run --headless` with an Inference backend that returns a truncated JSON object. Before the patch the log is one line naming the exception; after it, the same line followed by the head, an ellipsis, the tail, and stderr.


danielmiessler:
Ported: the unparseable-JSON branch logs the output head and tail plus stderr. Next release carries it. Thanks, @pai-scaffolde.

Issue #1501, opened 2026-07-15
ArchitectureSummaryGenerator --check can never pass: drift guard matches every semver, not just Algorithm versions


HyggeHacker:
## Bug

The version-drift guard in `cmdCheck` matches **every** `vX.Y.Z` in `LifeosSystemArchitecture.md` and reports them all as "older Algorithm version(s)". That doc legitimately cites hook versions, the Memory version, Bunker's spine version, and historical Algorithm notes — so `--check` exits 1 on a correct, freshly generated tree, permanently.

`LifeOS/install/LIFEOS/TOOLS/ArchitectureSummaryGenerator.ts:287`

```ts
const cited = [...archContent.matchAll(/v(\d+\.\d+\.\d+)/g)].map(m => m[1]);
```

## Repro

Against a pristine checkout of `main` (03a969dd, v7.1.1):

```sh
git archive main LifeOS/install/LIFEOS | tar -x -C /tmp/repro
L=/tmp/repro/LifeOS/install/LIFEOS
# make the summary newest, so the mtime guard doesn't short-circuit first (see note below)
touch "$L/DOCUMENTATION/ARCHITECTURE_SUMMARY.md"
LIFEOS_DIR="$L" bun "$L/TOOLS/ArchitectureSummaryGenerator.ts" check
```

Output:

```
STALE: LifeosSystemArchitecture.md references older Algorithm version(s) 8.2.0, 6.29.0, 8.0.0, 6.11.0, 6.10.0, 6.9.0, 6.8.0, 6.7.0, 1.3.0, 0.1.0, 6.18.0 — current is v8.4.0
exit 1
```

Not one of those eleven is Algorithm drift:

- `8.2.0` — the **Memory** version (`**Version:** LifeOS 7.1.1 | Algorithm v8.4.0 | Memory v8.2.0`)
- `1.3.0` — the **AgentInvocation hook** ("model injector was removed in v1.3.0")
- `0.1.0` — the **Bunker** spine
- the `6.x` — historical notes about when things changed

The only Algorithm-qualified mention in the doc is `Algorithm v8.4.0`, which is correct and current.

## Impact

The freshness gate reports STALE regardless of actual drift, so it can't pass. A check that always fails is one people learn to ignore — and real drift would look identical to the noise.

## Fix

[...]

PR #2126, opened 2026-09-14
fix(archsummary): skip the Architecture Decisions log in the drift check

pai-scaffolde:

## Observed

`bun LIFEOS/TOOLS/ArchitectureSummaryGenerator.ts check` fails on the shipped tree, with nothing wrong in it:

```
STALE: LifeosSystemArchitecture.md references older Algorithm version(s) 8.8.0 — current is v8.20.2
exit 1
```

The only hit is inside AD-4 in the master doc's `## Architecture Decisions` log:

```
**Superseded again (2026-07-24, Algorithm v8.8.0):** doctrine now names ROLES, never models.
```

That is a dated record of when a rule was superseded. The only way to make the check pass is to edit it, which the log forbids.

## Root cause

`cmdCheck()` scans the whole master doc for Algorithm-qualified versions. The doc's `## Architecture Decisions` section is an append-only history log, and says so in its own entries:

```
- **Replaced:** … (AD-4 is left intact per the append-only convention; …)
- **Superseded:** 2026-07-11 — … Entry kept per the append-only convention.
```

The same rule is stated for versioned history generally in § Versioning: "Historical changelog entries stay as recorded — no back-filling." So every superseded-at-vX.Y.Z row in that log is permanent, and the drift scan reads each one as drift. The check has been failing on every install since AD-4 was written, and it gets worse with each decision recorded, because a fresh superseding note is stale the moment the Algorithm moves on.

The narrower version of this was already fixed once: bare-semver matching was scoped to Algorithm-qualified mentions for public issue #1501. That kept the hook/Memory/Bunker versions out but not the AD log, where the mentions genuinely are Algorithm-qualified.

## Fix

Slice the doc at `\n## Architecture Decisions` before the scan and check only the living sections above it. Six lines, in `cmdCheck()` alone; `generate()` still reads the whole file, so the summary is unchanged. If the section is absent the whole doc is scanned, as before.

Drift in the living sections — the sections a reader treats as current — is still caught. Only the history log is exempt, which is the correct boundary: history is not drift.

## How tested

Probe against a sandbox `HOME`/`LIFEOS_DIR` holding upstream `main`'s own master doc plus `ALGORITHM/LATEST` at `8.20.2`, with one extra fixture AD row saying `Superseded at Algorithm v8.9.0`:

```
$ HOME=$SB/home LIFEOS_DIR=$SB/home/.claude/LIFEOS bun upstream-main.ts check
STALE: LifeosSystemArchitecture.md references older Algorithm version(s) 8.8.0, 8.9.0 — current is v8.20.2
exit=1

$ HOME=$SB/home LIFEOS_DIR=$SB/home/.claude/LIFEOS bun branch.ts check
FRESH: Summary is up to date
exit=0
```

`8.8.0` there is upstream's own AD-4 row, not a fixture. Negative control — same sandbox, one stale mention added to a living section above the log:

```
$ HOME=$SB/home LIFEOS_DIR=$SB/home/.claude/LIFEOS bun branch.ts check
STALE: LifeosSystemArchitecture.md references older Algorithm version(s) 8.7.0 — current is v8.20.2
exit=1
```

Only `8.7.0` is reported, so the guard still has teeth and the log is what got skipped. Transpiles: `bun build --target=bun --no-bundle LIFEOS/TOOLS/ArchitectureSummaryGenerator.ts` exits 0. On the live install the check now passes and the AD log was left as recorded.

Repro on a clean tree: point `LIFEOS_DIR` at a copy of the shipped `LIFEOS/`, write the current Algorithm version to `ALGORITHM/LATEST`, and touch `DOCUMENTATION/ARCHITECTURE_SUMMARY.md` so it is newer than the doc and `CLAUDE.md`. Before the patch `check` exits 1 naming the versions cited inside `## Architecture Decisions`; after it, exit 0.


danielmiessler:
Ported: the drift check slices off the Architecture Decisions log first. Next release carries it. Thanks, @pai-scaffolde.

PR #2127, opened 2026-09-14
docs(evals): match ViewResults to the run.json the runner writes

pai-scaffolde:

## Observed

Every command in the ViewResults workflow opens a file the skill has never written. On an install with one completed suite run, the workflow's own Step 2 is a dead end:

```
$ cat ~/.claude/LIFEOS/MEMORY/STATE/Evals-Results/core-dispositions/run_1789403834855_ochjyf/results.json
cat: .../results.json: No such file or directory

$ ls ~/.claude/LIFEOS/MEMORY/STATE/Evals-Results/core-dispositions/
latest.json
run_1789403834855_ochjyf
```

## Root cause

`EvalRunner.ts` writes `<suite>/<run-id>/run.json` and `<suite>/latest.json`:

```ts
// skills/Evals/Tools/EvalRunner.ts:203-206
const dir = join(RESULTS_DIR, name);
mkdirSync(join(dir, runId), { recursive: true });
writeFileSync(join(dir, runId, 'run.json'), JSON.stringify({ ...result, detail: caseResults }, null, 2));
writeFileSync(join(dir, 'latest.json'), JSON.stringify({ ...result, ts: new Date().toISOString() }, null, 2));
```

Three separate mismatches follow from that. The file is `run.json`, not `results.json`. The directory is keyed by the suite `name` passed to `runSuite()`, not by a use-case. And the JSON shape is `SuiteResult` (`EvalRunner.ts:62-72`), which has no `.trials[]` and no `.graders` — per-trial records live at `.detail[].trials[]`, and grader output at `.asserts[]` (`EvalRunner.ts:59`, `Assertions.ts:47-53`). The other consumers already agree with the runner: `SuiteManager.checkSaturation` opens `run.json` (`SuiteManager.ts:144`) and `ScenarioRunner.ts:171` writes one.

## Fix

Docs only, one file. Correct the two paths, the `<use-case>` placeholder, and the jq queries so each one names a key that exists. Both `SuiteManager` invocations the workflow already carried (`list` in Step 1, `check-saturation` in Step 3) are valid subcommands and are kept as-is. Step 3 gains two caveats that follow from the same key mismatch: `checkSaturation` reads `run.pass_rate` (`SuiteManager.ts:150`), a field `EvalRunner.ts` does not write, so the history can come back empty and read as "not saturated"; and `loadSuite` resolves only the skill's own `Suites/Capability` and `Suites/Regression` (`SuiteManager.ts:70-75`), not the USER customization layer `EvalRunner.ts` searches first (`EvalRunner.ts:32, 84`).

## How tested

Every command and key in the new doc maps to the upstream source line that produces it:

| New doc | Upstream source |
|---|---|
| `<suite>/<run-id>/run.json` | `EvalRunner.ts:203-205` |
| `<suite>/latest.json` | `EvalRunner.ts:206` |
| `ls -1t "$SUITE" \| grep '^run_'` | run id is `` `run_${Date.now()}_…` `` — `EvalRunner.ts:155` |
| `passed, score, pass_to_k, pass_at_k, summary, run_id` | `SuiteResult` — `EvalRunner.ts:62-72`, built at `190-200` |
| `latest.json` extra `ts` | `EvalRunner.ts:206` |
| `.cases[]` = `{id, mean_score, pass_to_k, pass_at_k}` | `EvalRunner.ts:71, 199` |
| `.detail[]` | `EvalRunner.ts:205` (`{...result, detail: caseResults}`) |
| `.detail[].trials[]` = `{score, passed, asserts, output}` | `CaseResult` — `EvalRunner.ts:59`, pushed at `166` |
| `.asserts[] \| {type, reason}` | `AssertResult` — `Assertions.ts:47-53` |
| `SuiteManager.ts list` | `SuiteManager.ts:316-325` |
| `SuiteManager.ts check-saturation` | `SuiteManager.ts:356-367` |
| `check-saturation` reads `pass_rate` | `SuiteManager.ts:150` (absent from `SuiteResult`) |
| skill-only suite resolution | `SuiteManager.ts:70-75` vs `EvalRunner.ts:84` |

Then run against a real artifact. A copy of this install's `core-dispositions` results was placed in a scratch directory and every command from the new doc run verbatim against it:

```
$ SUITE=<probe>/Evals-Results/core-dispositions
$ ls -1t "$SUITE" | grep '^run_'
run_1789403834855_ochjyf

$ jq '{passed, score, pass_to_k, pass_at_k, summary, run_id}' "$SUITE/latest.json"
{"passed": true, "score": 0.75, "pass_to_k": 0.8, "pass_at_k": 0.8,
 "summary": "pass^k 80%, mean 75.0% over 2 trials", "run_id": "run_1789403834855_ochjyf"}

$ RUN="$SUITE/$(ls -1t "$SUITE" | grep '^run_' | head -1)/run.json"
$ jq -c '.cases[] | select(.pass_to_k < 1)' "$RUN"
{"id":"concise_and_plain","mean_score":0,"pass_to_k":0,"pass_at_k":0}

$ jq -c '.detail[] | select(.id=="concise_and_plain") | .trials[]
        | {score, passed, failed: [.asserts[] | select(.passed==false) | {type, reason}]}' "$RUN"
{"score":0,"passed":false,"failed":[{"type":"max-length","reason":"length 1647 > 1400"},{"type":"llm-rubric","reason":"judge error: Timeout after 45000ms"}]}
{"score":0,"passed":false,"failed":[{"type":"max-length","reason":"length 1420 > 1400"},{"type":"llm-rubric","reason":"judge error: Timeout after 45000ms"}]}

$ jq -r '.detail[] | select(.id=="concise_and_plain") | .trials[0].output' "$RUN" | head -1
════ LifeOS ═══════════════════════════
```

Every query returns data. The same queries against `results.json` and `.trials[]` return nothing, because neither exists. No `.ts` file is changed, so there is nothing to transpile.

Repro on a clean tree: `rg -n 'results\.json' LifeOS/install/skills/Evals/` lists the pointers, then `rg -n "'run.json'" LifeOS/install/skills/Evals/Tools/` shows `EvalRunner.ts` writing it and `SuiteManager.ts` reading it. The same stale `<use-case>/<run-id>/results.json` pointer also appeared in `Workflows/CompareModels.md`, `CreateJudge.md`, `CreateUseCase.md` and `RunEval.md`; the second commit below sweeps those four.

## Second commit: sibling workflows

The same pointer is stale in four sibling workflows. This is a pointer sweep — only the file name, the result-directory placeholder, and jq keys naming a field `SuiteResult` does not have. Every other sentence is untouched.

| File:line | Was | Now | Upstream source |
|---|---|---|---|
| `CompareModels.md:97` | the per-run `results.json` | the per-run `run.json` | `EvalRunner.ts:205` |
| `CompareModels.md:112` | `<use-case>-<model>/<run-id>/results.json` | `<use-case>-<model>/<run-id>/run.json` | `EvalRunner.ts:203-205`; the directory is keyed by the suite `name` given to `-s`, which this workflow sets to `<use-case>-<model>` at lines 101-109, so that placeholder was already right |
| `CompareModels.md:118` | `<use-case>/models/<run-id>/` | `<use-case>-<model>/<run-id>/` | `EvalRunner.ts:203-204` — the runner creates `<suite>/<run-id>/` and no `models/` level anywhere |
| `CreateJudge.md:103` | `<use-case>/<run-id>/results.json \| jq '.trials[0].graders'` | `<suite>/<run-id>/run.json \| jq '.detail[0].trials[0].asserts'` | file `EvalRunner.ts:205`; the directory matches the `-s <suite>` two lines above; `.detail[]` `EvalRunner.ts:205`, `.trials[]` `EvalRunner.ts:59`, `.asserts[]` `Assertions.ts:47-53`. `SuiteResult` (`EvalRunner.ts:62-72`) has neither `.trials` nor `.graders` |
| `CreateUseCase.md:234` | `<name>/$(ls -1t … \| head -1)/results.json \| jq '.summary'` | `<name>/latest.json \| jq '.summary'` | `latest.json` is the rolling summary this step wants — `EvalRunner.ts:206`; `.summary` is `EvalRunner.ts:197`. The old `ls -1t … \| head -1` also selected `latest.json` itself, because the runner writes it after the run directory |
| `RunEval.md:54` | `<use-case>/<run-id>/results.json` | `<use-case>/<run-id>/run.json` | `EvalRunner.ts:205`. `<use-case>` is this workflow's own `-s` argument, so the directory key was already correct |

Checked against the same artifact as the first commit:

```
$ jq '.summary' "$SUITE/latest.json"
"pass^k 80%, mean 75.0% over 2 trials"

$ jq -c '.detail[0].trials[0].asserts' "$RUN"
[{"type":"llm-assert","passed":true,"score":1,"reason":"2/2 assertions TRUE","weight":2},{"type":"llm-rubric","passed":true,"score":1,"reason":"…","weight":1}]

$ jq -c '.trials[0].graders' "$RUN"
null
```

Deliberately left alone: `comparison.json` (`CompareModels.md:119`) and the `summary.yaml` → `report.md` render in Step 6 (`CompareModels.md:124-129`). Nothing under `skills/Evals/Tools/` writes any of those three files, so correcting their paths would only relocate prose about artifacts that do not exist. Whether those steps should exist at all is a separate question, and I did not want to delete prose inside a pointer fix.


danielmiessler:
The Evals docs were rewritten past this in source; the stale pointer is gone. Closing with thanks, @pai-scaffolde.

PR #2134, opened 2026-09-16
hooks: flag literal <DA> placeholder in the response closer

zephyr91:
## What

The 🗣️ closer's name slot must resolve to the DA identity file's name at render time. When substitution fails, the raw `<DA>` placeholder reaches the user **undetected** — `FormatGate` had no rule for it and `DriftReminder`'s pre-write contract didn't measure it.

This adds the check to both hooks:

- **`FormatGate.hook.ts`** (Stop, observation-only) records a new `da-placeholder` violation to `format-gate.jsonl`.
- **`DriftReminder.hook.ts`** (UserPromptSubmit, pre-write) reports it in the next turn's `FORMAT CONTRACT` line, so the model is told before it writes.

## Why closer-line only, and why *last visible line*

The check is anchored to the **last visible line** (the real closer), never to "last line containing the emoji". Two reasons it can't be looser:

1. Quoting the placeholder in prose — e.g. a PR or doc that discusses the format itself — is legitimate and a whole-message scan would false-positive on it.
2. A response that quotes the **emoji** in a body bullet (describing the format) must not be misread as a leaked closer. Anchoring to the final line, matching what `FormatGate` already does, keeps the two hooks consistent.

The check also requires the last line to actually contain the emoji before it treats a `<DA>` there as a leak, so a response with no closer at all isn't misattributed.

## Verification

All four cases exercised against the new `measure()`:

| Input | Result |
|---|---|
| Real message that quoted the emoji + token in body prose | clean (no false positive) |
| Leaked closer (`🗣️ <DA>: …` as final line) | **flagged** |
| Correct closer (`🗣️ <NAME>: …`) | clean |
| Emoji + token in a body bullet, clean final closer | clean |

No `~/.claude` paths, names, or identity data in the diff; the `{{PRINCIPAL_NAME}}` scaffold is preserved.


danielmiessler:
Ported the intent (both hooks had been restructured since): FormatGate now flags a literal `<DA>` in the closer's name slot only, credited to you, with a test. Next release carries it. Thanks, @zephyr91.

PR #2143, opened 2026-09-17
fix(Cortex): DOMAINS hardcodes four note types, kb-v3 declares six

joecoxtech:
`_schema.md` (kb-v3) declares six note types, but three tools still hardcode four. `KnowledgeDistill.ts` already lists all six, which is what makes this look like a migration that stopped partway.

```
LifeOS/install/LIFEOS/TOOLS/KnowledgeDistill.ts:43     6 types  ✅
LifeOS/install/LIFEOS/TOOLS/KnowledgeGraph.ts:52       4 types  ❌
LifeOS/install/LIFEOS/TOOLS/MemoryRetriever.ts:61      4 types  ❌
LifeOS/install/LIFEOS/TOOLS/KnowledgeHarvester.ts:152  4 types  ❌
```

**Effect:** every `Blogs/` and `Books/` note is invisible to graph traversal, graph stats, MOC generation, orphan detection and seedling pruning. `MemoryRetriever` is the worst of the three, since it builds the memory block injected into every turn — book and blog notes have never surfaced there.

Reproduced on a 7.40.4 install with 1 blog note and 3 book notes. Before: `Nodes: 7`, and an ingest reported 2 of a note's 3 typed links because the `Books/` target was dropped. After: `Nodes: 11 (People: 2, Ideas: 5, Blogs: 1, Books: 3)`, all 3 links resolved, and `harvester index` now emits `Blogs/_index.md` and `Books/_index.md`.

**Also included:** the four stale four-type lists in `skills/Cortex/SKILL.md` (lines 4, 107, 127, 229). That half was reported in #1986 and closed as "Fixed via your #2042. Next release." — but #2042 was closed without merging, so those lines are still on main today. #1986 also assumed it was doc-only ("Pure doc change, no code involved"), which is why the three TypeScript constants were never caught.

`KnowledgeLint` reports 11/11 fully conformant after the change.


danielmiessler:
Ported, combined with #2149: `MemoryRetriever.ts` and the Cortex docs now cover all six note types, via the shared `ALL_DIRS` constant, credited to you. Next release carries it. Thanks, @joecoxtech.

Issue #2047, opened 2026-09-02
identity proposals rejected for a path that IS allowed: normalizeProposalTargetFile only handles absolute and ~ paths, and one bad item fails the whole reviewer run


aveneri:
**Version:** 7.40.4 · **Platform:** Windows 10 (the separator half is Windows-specific; the relative-path half is not) · **Bun:** 1.3.10

`identity` proposals are rejected for naming a file that **is** on the allowlist, because `normalizeProposalTargetFile` only understands absolute paths and a leading `~`. One rejected item then marks the whole reviewer run `failed`, which surfaces as a CRITICAL memory-health alert.

## What happens

```
EINVAL_ITEM: target_file 'LIFEOS/USER/PRINCIPAL/PRINCIPAL_IDENTITY.md'
is not an allowed 'identity' target
```

`PRINCIPAL_IDENTITY.md` is one of exactly two allowed identity targets. The path is correct; only its *format* is not.

Observed across runs: a proposal naming `PRINCIPAL_IDENTITY.md` applied successfully on one run and was rejected on another, because the model emits an absolute path sometimes and a repo-relative path other times. The failure is non-deterministic from the user's side.

## Root cause

`LIFEOS/TOOLS/MemoryTypes.ts`:

```ts
export function normalizeProposalTargetFile(targetFile: string): string {
  if (targetFile === "~" || targetFile.startsWith("~/")) {
    return pathJoin(homedir(), targetFile.slice(1));
  }
  return targetFile;                       // relative paths pass through unchanged
}
```

It handles exactly one case. Anything relative is returned as-is and then compared against absolute allowlist entries, so it can never match.

This only bites `identity` because `identity` is the **sole multi-file kind**:

```ts
identity: Object.freeze([PRINCIPAL_IDENTITY_PATH, DA_IDENTITY_PATH]),
```

```ts
if (allowed.length === 1) return allowed[0];        // single-file kinds discard the supplied path
const normalized = normalizeProposalTargetFile(suppliedTargetFile);
return allowed.includes(normalized) ? normalized : null;   // identity must match exactly
```

Every other kind ignores `target_file` entirely and substitutes the canonical path, so a malformed path is harmless there. `identity` is the one kind that must match, and the normalizer isn't strong enough to let a legitimate path through.

On Windows there is a second, independent mismatch: the allowlist is built with `pathJoin` (backslashes) while the reviewer emits forward slashes, so even a correct *absolute* path can fail on separator style alone.

[...]

PR #2144, opened 2026-09-17
fix(memory): reviewer proposals with a relative target_file fail the whole run

joecoxtech:
`normalizeProposalTargetFile` expands `~/` before the allowlist comparison, but leaves repo-relative paths alone. The allowlist in `PROPOSAL_KIND_TO_FILES` is absolute, so a reviewer item carrying `LIFEOS/USER/PRINCIPAL/PRINCIPAL_IDENTITY.md` never matches. `pinProposalTargetFile` returns `null`, `MemorySystem.ts:331` rejects with `EINVAL_ITEM`, and the run is marked failed.

Only `identity` is affected in practice, because it is the single multi-file kind — every other kind is pinned to one canonical file by `pinProposalTargetFile`'s `allowed.length === 1` branch before the path is ever compared.

**Observed** on 7.40.4:

```
"failures": [{"index": 3, "type": "proposal",
  "error": "EINVAL_ITEM: target_file 'LIFEOS/USER/PRINCIPAL/PRINCIPAL_IDENTITY.md'
            is not an allowed 'identity' target"}]
"error": "dispatch failed for 1 item(s)"
```

`MemoryHealthCheck` then reports `overall: critical` via `reviewer-latest-failed`, and the fact the reviewer was trying to record is silently lost.

This is the same bug class the function already documents for tildes ("2026-08-10 run failure: a valid DA_IDENTITY proposal was rejected for the tilde alone"), fixed then for `~/` only. Resolving any non-absolute path against `CLAUDE_ROOT` makes the comparison absolute-vs-absolute for every form the reviewer emits.

**The allowlist still does the gating.** Verified after the change:

| input | result |
|---|---|
| `LIFEOS/USER/PRINCIPAL/PRINCIPAL_IDENTITY.md` | ACCEPT |
| `~/.claude/.../DA_IDENTITY.md` | ACCEPT |
| absolute `.../PRINCIPAL_IDENTITY.md` | ACCEPT |
| `LIFEOS/USER/PRINCIPAL/../../../../etc/passwd` | REJECT |
| `/etc/passwd` | REJECT |

`MemoryReviewer review` then completes `succeeded: 5, failed: 0`, and `MemoryHealthCheck` drops from `critical` to `warn`.


danielmiessler:
Already fixed in source (via #2047): `normalizeProposalTargetFile` handles repo-relative and LIFEOS-relative paths. Closing with thanks, @joecoxtech.

PR #2145, opened 2026-09-17
fix(services): commitmentsweep plist hardcodes bun's path, job silently never runs

joecoxtech:
`com.lifeos.commitmentsweep.plist.template` hardcodes `/opt/homebrew/bin/bun` (line 24). Bun's own installer puts it in `~/.bun/bin`, so on those machines launchd cannot exec it.

**The failure is silent**, which is what makes it worth fixing. The job installs, `launchctl bootstrap` succeeds, the installer's own verification step passes, and `launchctl list` shows the label. It just never runs:

```
-	78	com.lifeos.commitmentsweep     # 78 = EX_CONFIG
```

The log file is created and stays empty. Nothing surfaces the problem.

Hit while standing up a daily job from this template on a machine with bun at `~/.bun/bin/bun`. After switching to the detected path, same job: exit status `0`, work completed, log populated.

**`com.lifeos.worksweep` already has the fix** — `{{BUN}}` / `{{BUN_DIR}}` placeholders plus a `which bun` lookup in `InstallWorkSweep.ts:87`. This just applies that existing convention to `commitmentsweep`, which is the only remaining `*.plist.template` with a hardcoded interpreter path.

The installer now also fails loudly when bun is absent rather than writing a plist that cannot work.


danielmiessler:
Already fixed in source before this PR: the plist is a template with `{{BUN}}` placeholders and the installer resolves `bun` with `which`. Closing with thanks for the report, @joecoxtech.

PR #2149, opened 2026-09-17
fix(Cortex): walk every KNOWLEDGE dir, not the four that were hardcoded

Bowlerjim:
Fixes #2143.

`KnowledgeGraph.ts` and `KnowledgeHarvester.ts` each declare

```ts
const DOMAINS = ["People", "Companies", "Ideas", "Research"];
```

while kb-v3 declares six canonical types. A note of type `book` or `blog` therefore lives in a
directory that neither tool walks: no error, no warning, just absent from the graph, the stats,
the harvest queue, and anything downstream that asks the graph a question.

`KnowledgeSchema.ts` already exports `ALL_DIRS` and documents it as "Every KNOWLEDGE/ subdir,
canonical order. Single source for tools that walk the archive." Both tools now use it, so the
list cannot drift from the schema again.

## Verification

A probe archive containing exactly one `Books` note, same note in both runs:

```
before   Nodes: 0 (People: 0, Companies: 0, Ideas: 0, Research: 0)
after    Nodes: 1 (People: 0, Companies: 0, Ideas: 0, Blogs: 0, Research: 0, Books: 1)
```

The stats line also enumerates all six directories after the change, which is the same
`DOMAINS` list rendered, so the counter and the walker stay in agreement.

## Scope

Two lines of behaviour change plus an import and a comment. No schema change, no migration, no
change to any note on disk. Existing archives with only the four original directories are
unaffected, since the two added directories are simply empty.


danielmiessler:
Ported, combined with #2143: all three KNOWLEDGE walkers now import `ALL_DIRS` from `KnowledgeSchema.ts`, credited to you at each site. The public repo is generated from a private tree, so PRs are ported rather than merged; the next release carries it. Thanks, @Bowlerjim.

